PR #12, opened 2021-08-19
examples: update pytorch model inversion to custom plugins

jtsextonMITRE:


examples/pytorch-mnist-membership-inference/demo.ipynb
```diff
@@ -4,14 +4,14 @@
...
-    "# PyTorch MNIST Membership Inference demo"
+    "# MNIST Membership Inference demo for Securing AI Lab deployment"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This notebook contains an end-to-end demostration of Dioptra that can be run on any modern laptop."
+    "This notebook contains an end-to-end demostration for the Dioptra architecture that is deployed on the DGX workstation."
```
jkglasbrenner, line 14: This notebook looks like it is based off of https://github.com/usnistgov/dioptra/blob/main/examples/tensorflow-mnist-classifier/demo-dgx-lab-deployment.ipynb instead of https://github.com/usnistgov/dioptra/blob/29ee50438a001715d1ca0f3038527fe286e74719/examples/pytorch-mnist-membership-inference/demo.ipynb. Generally, we want to try to avoid including instructions about how to connect to specific machines like the DGX in the `demo.ipynb` notebook in each example since most users won't have access to it.

Is it possible to revert the notebook back to https://github.com/usnistgov/dioptra/blob/29ee50438a001715d1ca0f3038527fe286e74719/examples/pytorch-mnist-membership-inference/demo.ipynb and still have the example work? My thought is that the most important changes would potentially be to a couple of the arguments passed to `restapi_client.submit_job()`, such as `entry_point` and `entry_point_kwargs`. If that's the case, revert the notebook and update the arguments as needed.

examples/pytorch-mnist-membership-inference/docker-compose.yml
```diff
@@ -202,12 +207,13 @@ services:
...
       OMP_PROC_BIND: "false"
       RQ_REDIS_URI: redis://redis:6379/0
+      TF_CPP_MIN_LOG_LEVEL: "2"
     command:
       - pytorch_cpu
     networks:
       - lab_deployment_backend
     volumes:
-      - nfs-datasets:/nfs/data/Mnist
+      - nfs-datasets:/nfs/data
```
jkglasbrenner, line 216: If the `demo.ipynb` notebook is reverted back (see my comment on `demo.ipynb` file), then this file should reverted to https://github.com/usnistgov/dioptra/blob/29ee50438a001715d1ca0f3038527fe286e74719/examples/pytorch-mnist-membership-inference/docker-compose.yml to ensure compatibility.

jkglasbrenner (commented):
The image file `securing_ai_lab_architecture.png` should be deleted.

examples/pytorch-mnist-membership-inference/demo.ipynb
```diff
@@ -186,6 +186,74 @@
...
+    "# Delete the 'evaluation' custom task plugin package\n",
+    "restapi_client.delete_custom_task_plugin(name=\"evaluation\")\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "After you have deleted the task plugin in the testbed, re-run the `make custom-plugins` code block to update the package archive, then upload the updated plugin by re-running the `restapi_client.upload_custom_plugin_packge` block.\n",
+    "\n",
```
jkglasbrenner, line 256: It looks like the contents of the "evaluation" custom task plugin in this example is different [from the one in the tensorflow-mnist-classifier example](https://github.com/usnistgov/dioptra/tree/29ee50438a001715d1ca0f3038527fe286e74719/examples/tensorflow-mnist-classifier/task-plugins/securingai_custom/evaluation). This will lead to a naming conflict if a user runs that notebook and this notebook on the same deployment.

How to fix:

1. Rename `task-plugins/securingai_custom/evaluation` to something else, such as `task-plugins/securingai_custom/pytorch_mi`.
2. In `demo.ipynb`, update the value of the `custom_plugin_name` argument passed to `restapi_client.upload_custom_plugin_package()` to match, as well as the explanatory text surrounding these code cells to replace references to `evaluation` accordingly.
3. Update your scripts in the `src` directory to use the new custom task plugin name, for example https://github.com/usnistgov/dioptra/blob/b38422bc5254df96ba09c8b156a1ac5edf3e3a9a/examples/pytorch-mnist-membership-inference/src/train_mi.py#L263 would potentially change to `f"{_CUSTOM_PLUGINS_IMPORT_PATH}.pytorch_mi"` (it should match the `custom_plugin_name` chosen in step 2).
4. Update the Makefile to match the folder name change. For example, https://github.com/usnistgov/dioptra/blob/b38422bc5254df96ba09c8b156a1ac5edf3e3a9a/examples/pytorch-mnist-membership-inference/Makefile#L77 would update to
   ```
   CUSTOM_PLUGIN_PYTORCH_MI_SRC_FILES := $(wildcard $(PLUGINS_CUSTOM_DIR)/pytorch_mi/*.py)
   ```
   More or less a search and replace of `evalution` → `pytorch_mi` and `EVALUATION_` → `PYTORCH_MI_`.

I believe those are the main things that would need to be updated to handle the folder name change.

jkglasbrenner (changes requested):
@jtsextonMITRE Thanks for taking care of that first round of changes. Most of the comments in this second review are to restore a couple of missing lines from the Dioptra boilerplate, you should be able to accept them directly from the pull request without needing to edit and push.

The other comment is about ensuring the namespace for the custom task plugins in each of the examples is unique.

PR #64, opened 2022-06-03
Dioptra adv detectron demo update

hhuangMITRE:
Adding Detectron2 demo (poisoning on Balloon and Road Sign Detection datasets). Cleanup ImageNet patch demo outputs.


jkglasbrenner:
Looks good to me. Refactored the commits themselves a little (squashed down to two commits, use `examples:` prefix instead of `example:`). Content-wise this looks good so I'll merge this now.

PR #171, opened 2023-03-23
bug: logstream classes throw abstract method errors

alexb1200:
The classes StdoutLogStream, StderrLogStream subclassed from an abstract class without implementing all methods, causing an error when running the example notebooks. Added implementations to the classes to fix.


lbarbMITRE (approved):
LGTM

Issue #164, opened 2023-03-14
Create Python script to automate registration of Example plugins


pauldavidrowe:
**Definition of Done**
- [x] Dioptra utils.py code is moved to examples root and called client.py
- [x] Functionality of make custom-plugins is implemented in Python
- [ ] Code is tested
- [ ] Merged into dev

jkglasbrenner:
Consolidated into #177.

Issue #163, opened 2023-03-14
Port Example Makefile functionality to Python script


pauldavidrowe:
**Definition of Done**
- [x] Equivalent of make workflows is implemented (conforms to CVE requirements)
    - [ ] Looked through other examples and ported any other needed commands
- [ ] Script is tested on several examples
- [ ] Merged into dev branch

jkglasbrenner:
Consolidated into #177.

PR #172, opened 2023-03-23
Refactor: add tar help function to replace use of make and move utils to scripts folder

alexb1200:
Towards resolving #163 and #164, still need to change example notebooks and test changes.

### DIFF
--- a/examples/scripts/client.py
+++ b/examples/scripts/client.py
@@ -14,20 +14,19 @@
 #
 # ACCESS THE FULL CC BY 4.0 LICENSE HERE:
 # https://creativecommons.org/licenses/by/4.0/legalcode
+from __future__ import annotations
+
 import os
 from pathlib import Path
 from posixpath import join as urljoin
-from typing import Any, Dict, List, Optional, Union
+from typing import Any
 from urllib.parse import urlparse, urlunparse
 
 import requests
-from IPython.display import HTML
-
-PathLike = List[Union[str, Path]]
 
 
 class DioptraClient(object):
-    def __init__(self, address: Optional[str] = None) -> None:
+    def __init__(self, address: str | None = None) -> None:
         address = (
             f"{address}/api" if address else f"{os.environ['DIOPTRA_RESTAPI_URI']}/api"
         )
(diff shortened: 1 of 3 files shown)


examples/scripts/client.py
```diff
@@ -0,0 +1,258 @@
...
+            response = requests.post(
+                self.task_plugin_endpoint,
+                data=plugin_upload_form,
+                files=custom_plugin_file,
+            )
+
+        return response.json()
+
+
+def tar_helper(
```
jkglasbrenner, line 224: Style comment:

The convention in the repo is to name functions using action verbs and use nouns for classes. This should be renamed to something like `create_tar` or `make_tar`.

examples/scripts/client.py
```diff
@@ -0,0 +1,258 @@
...
+    source_dir: list[Union[str, Path]],
+    tarball_filename: str,
+    tarball_write_mode: str = "w:gz",
+    working_dir: Optional[Union[str, Path]] = None,
+) -> None:
+    """Archives a list of directories and files and compresses the archive.
+    Args:
```
jkglasbrenner, line 231: Style comment:

Google doc-strings should have a blank line separating each section.

examples/scripts/client.py
```diff
@@ -0,0 +1,258 @@
...
+    working_dir = Path(working_dir)
+    tarball_path = working_dir / tarball_filename
+
+    with tarfile.open(tarball_path, tarball_write_mode) as f:
+        for dir in source_dir:
+            dir = Path(dir)
+            f.add(dir, arcname=dir.name)
+
+    # uses octal so 436 in octal is 664, could alternatively use 0o to represent octal numbers.
+    os.chmod(tarball_path, 436)
```
jkglasbrenner, line 256: I don't think `os.chmod` is necessary here, try removing it and confirm that nothing is broken.

examples/scripts/utils.py
jkglasbrenner, line 1: This file shouldn't be moved in this merge, please undo.

examples/scripts/client.py
jkglasbrenner, line 1: The `DioptraClient` class should be in the `client.py` file by itself, since client implies code that talks to a server of some kind. The new function for creating tar files can be extracted into `examples/scripts/utils.py`.

jkglasbrenner (changes requested):
I did a quick review of the merge request and left comments and requested some changes. Please let me know if you have any questions.

lbarbMITRE (approved):
LGTM after changes

jkglasbrenner (approved):
LGTM after final edits

PR #175, opened 2023-03-31
fix(restapi): change service.py to slugify db queries

### LINKED ISSUE #71: REST API - Cannot retrieve an experiment name if a capital letter is used
## Description of bug

The bug happens when calling the GET method on the `experiment/name/<string:experimentName>` endpoint and `<string:experimentName>` contains a capital letter.

https://github.com/usnistgov/dioptra/blob/e68a0f40d6600582e9e3538faf0ceca86192f18f/src/dioptra/restapi/experiment/controller.py#L175-L189

`experimentName` is passed to `ExperimentService.get_by_name`, which does a database lookup and fails.

https://github.com/usnistgov/dioptra/blob/e68a0f40d6600582e9e3538faf0ceca86192f18f/src/dioptra/restapi/experiment/service.py#L150-L156

The lookup can fail because experiment names are intended to be case insensitive, and upon registration they are slugified (to keep them standardized):

https://github.com/usnistgov/dioptra/blob/e68a0f40d6600582e9e3538faf0ceca86192f18f/src/dioptra/restapi/experiment/schema.py#L98-L107

## How to fix

All methods in `ExperimentService` should slugify `experiment_name` before querying the database.

alexb1200:
Add the slugifiy helper function and use it on strings passed to the database to prevent misses since all names are slugified on registration.

Closes #71

### DIFF
--- a/src/dioptra/restapi/experiment/model.py
+++ b/src/dioptra/restapi/experiment/model.py
@@ -25,6 +25,7 @@ from wtforms.fields import StringField
 from wtforms.validators import InputRequired, ValidationError
 
 from dioptra.restapi.app import db
+from dioptra.restapi.utils import slugify
 
 from .interface import ExperimentUpdateInterface
 
@@ -82,9 +83,6 @@ class ExperimentRegistrationForm(FlaskForm):
         Args:
             field: The form field for `name`.
         """
-
-        def slugify(text: str) -> str:
-            return text.lower().strip().replace(" ", "-")
 
         standardized_name: str = slugify(field.data)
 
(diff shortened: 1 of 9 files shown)


jkglasbrenner:
I'm making the following additions to this contribution:

- I've added a unit test to validate that this change produces the desired behavior.
- `slugify` is used in enough places in the `dioptra.restapi` package that it makes sense to extract it into a common location and refactor the code as part of this update. I've extracted it into `dioptra.restapi.utils.slugify` and refactored the affected code to import and use this instead.

I've rebased everything on `dev` too and I'm letting the tests re-run. If they all pass, then we'll merge this.

jkglasbrenner (approved):
Tests passed

PR #178, opened 2023-04-06
feat: change behavior of make_tar to flatten directories

alexb1200:
Make tar flattens directory structure and packages all files to be unpacked into a single directory in the backend. Simplify the work done for the extract_all vulnerability patch.

### DIFF
--- a/examples/scripts/utils.py
+++ b/examples/scripts/utils.py
@@ -16,6 +16,7 @@
 # https://creativecommons.org/licenses/by/4.0/legalcode
 from __future__ import annotations
 
+import os
 import tarfile
 from pathlib import Path
 
@@ -26,7 +27,8 @@ def make_tar(
     tarball_write_mode: str = "w:gz",
     working_dir: str | Path | None = None,
 ) -> Path:
-    """Archives a list of directories and files and compresses the archive.
+    """Archives a list of directories and files, flattening the directory struture
+    and then compresses the archive.
 
     Args:
         source_dir: The directories and files which should be archived.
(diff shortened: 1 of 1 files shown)


examples/scripts/utils.py
```diff
@@ -26,7 +27,8 @@ def make_tar(
     tarball_write_mode: str = "w:gz",
     working_dir: str | Path | None = None,
 ) -> Path:
-    """Archives a list of directories and files and compresses the archive.
+    """Archives a list of directories and files, flattening the directory struture
+    and then compresses the archive.
```
jkglasbrenner, line 31: The [Google-style of docstrings](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) suggests that the first line of a docstring should be a one-line summary with no wraparound. Additional information can be included in a paragraph that starts below the one-line summary. So I would rewrite the docstring like this:

```suggestion
    """Archive a list of directories and files.
    
    This implementation flattens the source directory structure so that all the files
    are placed in the root of the archive, and then the archive is compressed.
```

examples/scripts/utils.py
```diff
@@ -52,6 +54,12 @@ def make_tar(
     with tarfile.open(tarball_path, tarball_write_mode) as f:
         for dir in source_dir:
             dir = Path(dir)
-            f.add(dir, arcname=dir.name)
+            if os.path.isdir(dir):
```
jkglasbrenner, line 57: The `Path()` objects have a builtin method for doing this check:

```suggestion
            if dir.is_dir():
```

examples/scripts/utils.py
```diff
@@ -52,6 +54,12 @@ def make_tar(
...
+                for dirpath, dirnames, filenames in os.walk(dir):
+                    for name in filenames:
+                        name = Path(os.path.join(dirpath, name))
```
jkglasbrenner, line 60: `Path()` objects can also join paths in a cross-platform way without needing to use `os.path.join`:

```suggestion
                        name = Path(dirpath, name)
```

jkglasbrenner (changes requested):
I left a couple of suggestions, one about how to format the Google-style docstrings and the others about using [`Path()`'s builtin methods](https://docs.python.org/3/library/pathlib.html) in lieu of `os.path` (style-wise, aim to use `Path()` whenever possible, as it replicates nearly all of the functionality of `os.path`).

Issue #152, opened 2023-03-08
Examples: Update documentation and support files for examples to account for new deployment documentation


hbooth:
# Definition of Done
- [ ] Documentation landing page for examples
- [ ] Documentation updated to point to deployment documentation
- [x] Spurious files removed

READMEs answer the question of "What do I do next?"
Documentation of example will walk a user through the particulars of an example. Tutorial of the example.
Documentation page for examples will describe how to accomplish a particular task or set of tasks or describes the nature of examples/plugins/etc..

### Prerequisites Setup

1. point to deployment instructions
    - (mention cpu/gpu distinction)
2. point to dataset download instructions
3. Instructions for mounting datasets for a local/single machine installation
    - point to more general instruction.
5. Instructions for minimal python env/dependencies (point to script)
6. Usage Instructions
    - (port make file to python script)

### Table Summarizing Example
Dataset used (prereq)
GPU needed? (prereq)
What does it show? Attacks/ Defenses (info)
ML Librrary? pytorch/tensorflow (info)
architecture (info)

## Each of the examples is updated:

- [x] pytorch-detectron2-demo
- [x] pytorch-mnist-membership-inference
- [x] tensorflow-adversarial-patches
- [x] tensorflow-backdoor-poisoning
- [x] tensorflow-imagenet-pixel-threshold
- [x] tensorflow-imagenet-resnet50-fgm
- [x] tensorflow-mnist-classifier
- [x] tensorflow-mnist-feature-squeezing
- [x] tensorflow-mnist-model-inversion
- [x] tensorflow-mnist-pixel-threshold

## These examples are marked for deletion

- [x] basic-mlflow-demo
- [x] patch-defended-pixel-threshold
- [x] pytorch-robust-dpatch
- [x] tensorflow-roadsigns-yolo

### Tasks for each example
- Task plugins go into common folder under examples directory
- Script to register all task plugins
- rename to avoid collisions and indicate example
- move MLproject files to src and remove use of relative path to src
- Delete extraneous files
    - init.db
    - environment.yml
    - architecture diagram
    - utils.py
    - /data subdir
    - /jupyter subdir
- simplify notebooks to focus only on example
- remove any descriptions of dataset

jkglasbrenner:
_Originally posted by @jkglasbrenner in https://github.com/usnistgov/dioptra/issues/19#issuecomment-1462671488_

[...]

PR #191, opened 2023-04-28
tf mnist pixel threshold update to new deployment 

alexb1200:
Towards closing #152


examples/task-plugins/dioptra_custom/pixel_threshold/pixelthreshold.py
```diff
@@ -124,11 +124,11 @@ def create_pt_dataset(
     target_size: Tuple[int, int] = image_size[:2]
     adv_data_dir = Path(adv_data_dir)
 
+    # attack requires clip values to be defined
+    keras_classifier.set_params(**{"clip_values": (0, 1)})
```
jtsextonMITRE, line 128: Purely stylistic suggestion but this can be removed and replaced with setting `clip_values` as a parameter to `load_wrapped_tensorflow_keras_classifier`. Not sure if this is something we actually want to do though as including it here is potentially more fitting as it's related to the pixel threshold attack directly..

examples/tensorflow-mnist-pixel-threshold/demo.ipynb
```diff
@@ -224,44 +248,16 @@
...
+   "execution_count": null,
    "metadata": {},
+   "outputs": [],
    "source": [
-    "```python\n",
-    "# Delete the 'evaluation' custom task plugin package\n",
-    "restapi_client.delete_custom_task_plugin(name=\"evaluation\")\n",
-    "restapi_client.delete_custom_task_plugin(name=\"pixel_threshold\")\n",
-    "```"
+    "restapi_client.delete_custom_task_plugin(name = \"pixel_threshold\")"
```
jtsextonMITRE, line 260: It feels a bit weird to have this in here and not also have a call to reupload the plugins. It does get in the way of clicking through the notebook as you have to be careful to reupload the plugins manually before moving on to the next step. Maybe this could be moved to a text block if we don't want to include a reupload call?

examples/tensorflow-mnist-pixel-threshold/src/train.py
```diff
@@ -327,9 +327,26 @@ def init_train_flow() -> Flow:
...
+        logged_tensorflow_keras_estimator = pyplugs.call_task(
+            f"{_PLUGINS_IMPORT_PATH}.tracking",
+            "mlflow",
+            "log_tensorflow_keras_estimator",
+            estimator=classifier,
+            model_dir="model",
+            upstream_tasks=[history],
+        )
+
+        logged_tensorflow_keras_estimator = pyplugs.call_task(
```
jtsextonMITRE, line 341: I don't think you need to have this call twice, it just has to exist at least once along with the call below to log_metrics

examples/tensorflow-mnist-pixel-threshold/demo.ipynb
jtsextonMITRE, line 260: It feels a bit weird to have this in here and not also have a call to reupload the plugins. It does get in the way of clicking through the notebook as you have to be careful to reupload the plugins manually before moving on to the next step. Maybe this could be moved to a text block if we don't want to include a reupload call?

examples/task-plugins/dioptra_custom/pixel_threshold/pixelthreshold.py
jtsextonMITRE, line 128: Purely stylistic suggestion but this can be removed and replaced with setting clip_values as a parameter to load_wrapped_tensorflow_keras_classifier. Not sure if this is something we actually want to do though as including it here is potentially more fitting as it's related to the pixel threshold attack directly..

jtsextonMITRE (commented):
Tested and it works!

examples/task-plugins/dioptra_custom/pixel_threshold/pixelthreshold.py
alexb1200: We could move it to to the entry point and make it a parameter. I'm unsure where the best place for it is as well.

examples/tensorflow-mnist-pixel-threshold/demo.ipynb
alexb1200: Thanks for catching this, I'd left that in when editing the plugins and testing.

jkglasbrenner (approved):
I reviewed the pull request and overall it looked good. All of my suggestions were minor and related to whitespace or ensuring consistency between the way things look in this notebook and the way the tensorflow-mnist-classifier notebook looks. I went ahead made those changes and pushed to the branch. For reference, a summary of the minor changes is given below:

- Standardized the white space in the Jupyter notebook.
- Removed the statement about the new setup tool in the README since the same instructions are linked further down in the enumerated list.
- Standardized the first code block with imports and constants to match `examples/tensorflow-mnist-classifier`.
- Removed a few constants and some prose that referred to the previous setup and approach to running the notebook.
- Ensured that the `entry_point_kwargs` for the train endpoint submission in the notebook was consistent with what is provided in `examples/tensorflow-mnist-classifier` since it is the same job.

jkglasbrenner:
I am preparing to merge this to dev.

jkglasbrenner:
Merging now with the following commit message:

```
examples: refactor the tensorflow-mnist-pixel-threshold example

Refactors the tensorflow-mnist-pixel-threshold demo to use the new, common approach designed for the
examples folder. This includes the following changes:

- Moved any new custom task plugins from tensorflow-mnist-pixel-threshold folder to the root of the
  examples folder.
- Fixed a clip values error by adding a --clip-values option to the entry point script.
- Moved the MLproject file into the src folder
- Updated the documentation in examples/tensorflow-mnist-classifier/README.md with instructions on
  how the user should prepare their environment for running the examples.
- Modified the demo Jupyter notebook for tensorflow-mnist-pixel-thresold to use the new scripts.
- Removed any files from the tensorflow-mnist-pixel-threshold folder that were no longer being used
  after implementing the above changes.

PR #192, opened 2023-05-08
examples: feature squeezing update

jtsextonMITRE:
Cleaned up and updated the feature squeezing example to use the new scripts in the examples folder, and checked that the stuff still runs.


andrewhand (commented):
Core code unchanged. Expected changes where to new Paths and reworking make to py functionss/plugs. Some removes of extraneous code such as imports copied+paste from templates. Code ran clean.

examples/task-plugins/dioptra_custom/feature_squeezing/tensorflow_plugin.py
jkglasbrenner, line 1: The code in this file seems to be a duplicate/earlier version of the built-in `dioptra_builtins/data/tensorflow.py`, the entrypoint scripts can be updated to just use the builtin versions instead and this file can be removed.

examples/tensorflow-mnist-feature-squeezing/src/data.py
jkglasbrenner, line 1: This is old code that was migrated to a task plugin and can be removed.

examples/tensorflow-mnist-classifier/demo.ipynb
jkglasbrenner, line 1: This notebook doesn't need to be changed, so undo this change before merging.

examples/tensorflow-mnist-feature-squeezing/src/feature_squeeze.py
jkglasbrenner, line 1: The if/else logic in lines 147–150 can be removed, as the builtin init_rng task plugin does the same thing.

examples/tensorflow-mnist-feature-squeezing/src/train.py
jkglasbrenner, line 1: The mlflow.autolog() function on line 158 should be deleted as it's functionality isn't compatible with how we've constructed the entrypoints.

examples/tensorflow-mnist-feature-squeezing/demo.ipynb
jkglasbrenner, line 1: In import code cell, for EXPERIMENT_NAME variable remove the username prefix and append `mnist_` as the prefix instead.

jkglasbrenner, line 1: Make the model training non-optional (so remove the Markdown blocks and just keep executable code blocks), and change the model registry name to something other than `mnist_<architecture>`, since otherwise it conflicts with the models created in the tensorflow-mnist-classifier example.

jkglasbrenner, line 1: Since the input parameters for the different attacks are listed, verify that the parameter names and their default values match what is in MLproject and the custom task plugins.

jkglasbrenner, line 1: Remove empty code block at the bottom of the notebook.

jkglasbrenner (changes requested):

jkglasbrenner (approved):

jkglasbrenner (approved):

PR #200, opened 2023-06-02
Fix mypy config and type errors

chisholm:
This PR makes a mypy config change which causes many more type errors to be found, and fixes many of them.  When I run mypy locally, I still get a few remaining errors:

```
src\dioptra\restapi\task_plugin\service.py:78:42: error: Incompatible types in assignment (expression has type "Optional[List[str]]", variable has type "List[str]")  [assignment]
src\dioptra\restapi\job\service.py:95:34: error: Incompatible types in assignment (expression has type "Optional[Experiment]", variable has type "Experiment")  [assignment]
src\dioptra\restapi\job\service.py:98:24: error: Incompatible types in assignment (expression has type "Optional[Queue]", variable has type "Queue")  [assignment]
Found 3 errors in 2 files (checked 176 source files)
```

We can discuss how to fix these, and any other fixes I made.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -245,6 +245,8 @@ honor_noqa = true
 [tool.mypy]
 python_version = "3.9"
 platform = "linux"
+mypy_path = "src,task-plugins"
+explicit_package_bases = true
 namespace_packages = true
 show_column_numbers = true
 follow_imports = "normal"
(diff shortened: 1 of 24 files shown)


jkglasbrenner:
In the interest of getting this merged into dev, I resolved the last three mypy errors by introducing a `if <...> is None:` check that raises an error that will be handled by one of the registered error handlers.

jkglasbrenner (approved):
LGTM, all checks passed after the final changes to resolve the last three type errors I mentioned above.

PR #202, opened 2023-06-08
Update Example: ahand-pytorch-mnist-mi

andrewhand:
Updated the Pytorch Mnist membership inference example.


jkglasbrenner (changes requested):
What follows is in addition to the requests on the changed files (GitHub doesn't provide a way to leave pull request review comments on unchanged files).

In `train_mi.py`, extract the following lines into the constant named `PERFORMANCE_METRICS` near the top of the file (this will replace the existing values):

https://github.com/usnistgov/dioptra/blob/10ce0a40743cb22f2e4cb9d2c7f89202de184605/examples/pytorch-mnist-membership-inference/src/train_mi.py#L286-L292

The same pattern should also be applied to these lines in `train_mi.py`: https://github.com/usnistgov/dioptra/blob/10ce0a40743cb22f2e4cb9d2c7f89202de184605/examples/pytorch-mnist-membership-inference/src/train_mi.py#L303-L315

examples/task-plugins/dioptra_custom/pytorch_mi/log.py
jkglasbrenner, line 1: This file configures loggers, and is likely redundant. Is this used/needed anywhere?

examples/task-plugins/dioptra_custom/pytorch_mi/metrics.py
jkglasbrenner, line 1: This file is similar to https://github.com/usnistgov/dioptra/blob/3fff494062dbeeca85e6a8936e680f05b93f33cf/task-plugins/dioptra_builtins/metrics/performance.py. Is this file used/needed anywhere?

jkglasbrenner (changes requested):
I missed a couple of things in my first review, so I'm adding them here. Basically, be sure to check through `src/train_mi.py` and `src/mi.py` and look for task plugin calls with `f"{_CUSTOM_PLUGINS_IMPORT_PATH}.pytorch_mi"`. In those, check which task plugins are being used. What we want to find out is which plugins in `examples/task-plugins/dioptra_custom/pytorch_mi` are never called and can be removed.

examples/task-plugins/dioptra_custom/pytorch_mi/log.py
andrewhand: They are not used anywhere in the example.

examples/task-plugins/dioptra_custom/pytorch_mi/metrics.py
andrewhand: It is not used. The example using the metrics file in diaoptra_builtin

jkglasbrenner (approved):
Looks good to me, preparing to merge into dev.

Issue #207, opened 2023-06-12
Migrate init-minio.sh script in cookiecutter template to use new "mc admin policy" commands


jkglasbrenner:
## Description

Breaking changes were introduced in the MinIO mc client earlier this year (2023) that broke the `init-minio.sh` script provided in Dioptra's cookiecutter template. The breakage happens for two reasons:

- The commands for attaching access control policies to users were deprecated and replaced with a new set of commands, see here for more info: <https://min.io/docs/minio/linux/reference/minio-mc-admin/mc-admin-policy.html>.
- The new `mc admin policy attach` command is not idempotent, such that executing it more than once will raise an error: <https://github.com/minio/minio/issues/16897>.

## Definition of Done

- [x] All deprecated commands are removed from the `init-minio.sh` script and replaced with the new commands.
- [x] Idempotency is restored to the `init-minio.sh` script, either through the use of error handling or by adding a validation step before `mc admin policy attach` is run.
- [x] The image tag for `minio/mc` in [cookiecutter-templates/cookiecutter-dioptra-deployment/cookiecutter.json](https://github.com/usnistgov/dioptra/blob/67a546c6e425fcc3eb5e967e2790e08f4fe72250/cookiecutter-templates/cookiecutter-dioptra-deployment/cookiecutter.json) (see below) is unpinned and set back to `latest`. https://github.com/usnistgov/dioptra/blob/67a546c6e425fcc3eb5e967e2790e08f4fe72250/cookiecutter-templates/cookiecutter-dioptra-deployment/cookiecutter.json#L86
- [x] A clean deployment folder is regenerated using the updated cookiecutter template and the `init-deployment.sh` script runs without error two times in a row.

jkglasbrenner:
Closed by #228

PR #228, opened 2023-07-06
build(deployment): revert minio version to latest and improve minio setup script

jtsextonMITRE:
Addresses #207. 

tensorflow-mnist-classifier worked after deployment, deployment runs smoothly once, and re-running the init-deployment.sh script does not crash on the minio section


jkglasbrenner (approved):
I confirmed that the updated logic worked on a fresh deployment. From there, I cleaned things up a bit so that everything was consolidated into the init-minio.sh script and removed the need to concatenate two script files together. The init-minio.sh file is now just executed on its own from inside the minio/mc container.

I also did some refactoring of init-minio.sh to make it look more like the other script files, so it should help to make maintaining this file easier going forward.

PR #243, opened 2023-07-18
build(deployment): update Tensorflow to 2.12.1

### LINKED ISSUE #205: Update the Tensorflow dependency in the worker containers to version 2.10 or greater
## Description

The current version of Tensorflow in the worker containers is 2.9.1 while as of writing the current version is 2.12.0. [As documented on the Tensorflow install page](https://www.tensorflow.org/install/pip), Tensorflow ARM64 support is available starting with version 2.10.0:

> Note: Starting with TensorFlow 2.10, Linux CPU-builds for Aarch64/ARM64 processors are built, maintained, tested and released by a third party: [AWS](https://aws.amazon.com/). Installing the [tensorflow](https://pypi.org/project/tensorflow/) package on an ARM machine installs AWS's [tensorflow-cpu-aws](https://pypi.org/project/tensorflow-cpu-aws/) package. They are provided as-is. Tensorflow will use reasonable efforts to maintain the availability and integrity of this pip package. There may be delays if the third party fails to release the pip package. See [this blog post](https://blog.tensorflow.org/2022/09/announcing-tensorflow-official-build-collaborators.html) for more information about this collaboration.

To upgrade, we need to check two things:

1. We need to verify that migrating to a newer version of Tensorflow does not introduce dependency conflicts in the Tensorflow worker containers.
2. We need to test our examples (at a minimum, `tensorflow-mnist-classifier`) to check if there are breaking changes in the Tensorflow API.

## Depends on

- #135

## Definition of Done

- [x] Update the Tensorflow version in [requirements-dev-tensorflow.in](https://github.com/usnistgov/dioptra/blob/dfeedec68c73927c49d5f3066e8b9a2794f73adf/requirements-dev-tensorflow.in) to version 2.10 or higher.
- [x] Generate the updated `requirements.txt` lockfiles in `requirements/` and `docker/requirements/` using [pip-compile](https://pip-tools.readthedocs.io/en/latest/).
- [x] Run `tensorflow-mnist-classifier` end to end using the updated Tensorflow worker containers.

jtsextonMITRE:
Closes #205.

Includes updates for optimizer import statements as well as an update to cudnn in the base tensorflow gpu image.


jkglasbrenner:
I've squashed and rebased this, and also generated the requirements files for the Apple Silicon Mac. I noticed that some requirements files were duplicated across the two folders (`requirements` and `docker/requirements`), I removed the duplicates as part of the force-push. In addition, I removed the instructions for GitHub Actions to regenerate the requirements files on each push since those files are there now.

PR #247, opened 2023-07-20
refactor: expand Dioptra Client to cover existing RestAPI

### LINKED ISSUE #211: Expand Dioptra Client to cover existing REST API
The current Dioptra Client only covers a subset of the existing REST API and should use composition with each top-level API resource being a composed class within the client code base.

## Design Spec

The formal client object is built using composition. Here's a quick spec on how that will look:

```python
class DioptraClient(object):
    def __init__(self, experiment, job, queue, ...):
        self._experiment = experiment
        self._job = job
        self._queue = queue
        # Set other endpoint objects in the same manner

    @property
    def experiment(self):
        return self._experiment

    @property
    def job(self):
        return self._job

    @property
    def queue(self):
        return self._queue
```

Using the client to submit a job would then work like this:

```python
client.job.submit(...)
```

Each sub-client in the main client represents an endpoint namespace. The classes for these endpoint clients will look something like this:

```python
class JobClient(object):
    def get_all(self) -> list[dict[str, Any]]:
        # Body of function follows

    def submit(
        self,
        workflows_file: str | Path,
        experiment_name: str,
        entry_point: str,
        entry_point_kwargs: str | None = None,
        depends_on: str | None = None,
        queue: str = "tensorflow_cpu",
        timeout: str = "24h",
    ) -> dict[str, Any]:
    # Body of function follows

# Classes like ExperimentClient, QueueClient follow
```

The function used to perform dependency injection and instantiate `DioptraClient` would look something like this:

```python
def get_dioptra_client(address: str | None = None, ...) -> DioptraClient:
    address = (
        f"{address}/api" if address else f"{os.environ['DIOPTRA_RESTAPI_URI']}/api"
    )
    scheme, netloc, path, _, _, _ = urlparse(address)
    # More needed with the address most likely

    experiment_client = ExperimentClient(...)
    job_client = JobClient(...)
    queue_client = QueueClient(...)

    return DioptraClient(
        experiment=experiment_client,
        job=job_client,
        queue=queue_client,
        ...
    )
```

In addition, it will likely be necessary to have a requests session set up to be able to handle and manage the use of cookies and other header information while using the client, so after the initial port, investigate switching to using a session object instead of direct calls to `requests.get`, etc.

## Depends On

- #210 

## Definition of Done

- [ ] Add new classes for each resource
- [ ] Class for each Resource covers all REST API calls
- [ ] New process to instantiate the new client use dependency-injection
- [ ] Merge into Dev

lbarbMITRE:
Closes #211

### DIFF
--- a/src/dioptra/client/__init__.py
+++ b/src/dioptra/client/__init__.py
@@ -16,6 +16,6 @@
 # https://creativecommons.org/licenses/by/4.0/legalcode
 from __future__ import annotations
 
-from ._client import DioptraClient
+from ._client import DioptraClient, get_dioptra_client
 
-__all__ = ["DioptraClient"]
+__all__ = ["DioptraClient", "get_dioptra_client"]
--- a/src/dioptra/client/_client.py
+++ b/src/dioptra/client/_client.py
@@ -800,7 +800,6 @@ class BuiltinTaskPluginClient(object):
         Returns:
             A list of responses detailing all builtin plugins.
 
-
             Example::
 
                 [
--- a/src/dioptra/client/_client.py
+++ b/src/dioptra/client/_client.py
@@ -56,14 +56,15 @@ def get_dioptra_client(address: str | None = None) -> DioptraClient:
 
 class DioptraClient(object):
     """Connects to the Dioptra REST api, and provides access to endpoints.
-
+    
     Args:
-        address: Address of the Dioptra REST api or if no address is given the DIOPTRA_RESTAPI_URI environment variable is used.
-
+        address: Address of the Dioptra REST api or if no address is given the
+            DIOPTRA_RESTAPI_URI environment variable is used.
+            
     Notes:
-           See https://pages.nist.gov/dioptra/user-guide/api-reference-restapi.html for more
-           information on Dioptra's REST api.
-    """  # noqa B950
+        See https://pages.nist.gov/dioptra/user-guide/api-reference-restapi.html for
+        more information on Dioptra's REST api.
+    """
 
     def __init__(
         self,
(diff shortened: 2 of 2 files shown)


jkglasbrenner:
@lbarbMITRE and @alexb1200, I'm force-pushing a fixed version of the branch. I attempted to rebase this, but it became too complicated because multiple commits were merged from dev. It was easier to just take the `_client.py` file, rebranch from dev, commit, and force-push back to this branch.

Of course, this means that the docstrings no longer match what is currently in `dev`. To resolve this, go through the docstrings on `dev`, https://github.com/usnistgov/dioptra/blob/d532c790271fd11c3f9acc0833f40aa6ef146c6a/src/dioptra/client/_client.py, and copy-and-paste the ones that did not change. For the ones that aren't the exactly same, compare and merge them accordingly, preferring to follow the style you see in `dev` as much as you can.

Also, don't worry about changing anything in the `examples/` folder, we will handle that in a separate Pull Request that will resolve #265.

jkglasbrenner:
@alexb1200 and @lbarbMITRE, just wanted to let you know that I've started taking a look at this, I agree in the comments that the requests Session should be handled as its own wrapper/proxy class. I've prototyped a version that should allow you to avoid repeating the try/except blocks everywhere, and also make it less verbose to close and refresh the session connection. It's not fully done, but I'll share when it is and I'll leave it to both of you to integrate.

I also think we're going to want to split the endpoint clients into separate files, the file is pretty big. I'll give more details when I write my actual code review.

keithmanville:
Will revisit the python client once V1 of the REST API is complete.

PR #252, opened 2023-07-28
refactor: remove dependency on Flask-Injector

### LINKED ISSUE #238: Resolve Flask compatibility issue with Flask-Injector library
The Flask-Injector library has not been updated in nearly a year, and it is preventing us from upgrading to the latest versions of the Flask library. The main functionality that we need to preserve from this package is the request-level scoping for the injected dependencies.

To fix this, we want to extract the request scope from here: https://github.com/python-injector/flask_injector/blob/993c0bbe5063bef511da60d8b25383a498a59f09/flask_injector/__init__.py#L226-L272. We need to bring that into our code base, and remove and replace the `FlaskInjector` class with our own custom use of injector (https://github.com/python-injector/injector).

We need to abide by the license when vendoring this dependency into our code base: https://github.com/python-injector/flask_injector/blob/master/COPYING.

## Definition of Done

- [x] The Request scope from Flask-Injector is vendored within our code base, with the appropriate license comments at the top of the file
- [x] The use of FlaskInjector is removed and replaced with our own bootstrapping setup using the injector library directly
- [x] All tests for the REST API pass after refactoring
- [x] Existing functionality in the REST API endpoints continue to exhibit same behavior
- [x] Remove the Flask-Injector dependency from the `pyproject.toml` file (and from tox.ini if it is used there)

chisholm:
This PR removes the Flask-Injector dependency.  It redoes some of what that library did, but in a simpler way.  For now, it assumes we only need injection into class-based view functions.  So it is tailored for Dioptra specifically.  It seems sufficient, for now.  But we can expand on this as needed, since we control this code.

The parts copy-pasted from Flask-Injector are in src/dioptra/restapi/shared/request_scope.py.  So that file has the Flask-Injector license terms, in a comment at the top (and not the NIST license).  It implements the injector request scope.  The injection setup code is in src/dioptra/restapi/utils.py, and is written by me (but based on what Flask-Injector does), so it retains the NIST license comment.

I removed the "<2.2.0" Flask version restriction in pyproject.toml.  Restapi unit tests pass for me with Flask 2.3.2.  But in my virtual environments and in the tox venvs, it still seems to choose 2.1.3 (I forced an upgrade to 2.3.2 to test).  Not sure what is still holding it back.

Fixes #238


src/dioptra/restapi/app.py
```diff
@@ -113,9 +117,8 @@ def health():
...
-    if not inject_dependencies:
-        return app
-
-    FlaskInjector(app=app, modules=modules)
+    if inject_dependencies:
+        injector = Injector(modules)
+        setup_injection(api, injector)
 
-    return app
+    return app, api
```
jkglasbrenner, line 124: What's the reasoning for changing the return signature of `create_app()` to return the flask-restx `Api` object in addition to the `Flask` app object? The `Api` object is registered with the Flask app and the user shouldn't need to access or modify it directly.

Is there some sort of scoping issue that emerged when stripping Flask-Injector down to setup_injection?

jkglasbrenner (changes requested):
The code looks good, but I had a question on whether it is necessary to alter what is returned by `create_app()`, see the inline comment for details.

src/dioptra/restapi/app.py
chisholm: The only reason is the unit tests which need to set up injection externally to `create_app()` (in the app fixture).  That requires the restx `Api` object.  I didn't find a way to access `api` via `app`, so I had to return both.  You can get app via api (which is why the `setup_injection()` function only needs api), but not vice versa.  (So you could return api only, if you wanted, but that would have been... even weirder.)

chisholm: A possibly more palatable design might be to change the `create_app()` signature to accept an injector as an arg, rather than `inject_dependencies`.  It could default to None.  If None, it would create a default injector, else it would use the one passed in.  Then, injection setup need only occur in `create_app()`, and there would be no need to return the `api` object.  With this design, it would not be possible to create an app without dependency injection at all, as is possible now (or maybe it would be possible to create a trivial injector with no bindings/modules, which can't inject anything?).  But perhaps that is not important?

jkglasbrenner: @chisholm After thinking about this some more, I like your second proposal of being able to override the dependency injection by changing `create_app()` to accept an injector as an argument. The `create_app()` function is intended to just be a factory for creating Flask app objects, and it would make sense that any (and all) dependency injection should occur there. Go ahead and update this branch with this idea.

And yes, I think if you pass something like `injector=Injector([])` you will be able to get back an app without any dependency injection, although as you hint at in your last question, it isn't clear to me why you would ever want that. Without the injection, none of the class-based views will work, and so you'd just have a broken app.

chisholm: Ok, this is done.

jkglasbrenner (approved):
Looks good to me

PR #255, opened 2023-08-07
fix: Platform specific prefixes

### LINKED ISSUE #248: Restapi task plugin service creates platform-specific S3 key prefixes
**Describe the bug**
In restapi: the task plugin service creates a S3 key prefix using the Python `pathlib` module, which causes the key prefix syntax to be platform dependent: on Windows backslashes are used; on Linux/Mac, forward slashes are used.  This causes restapi unit tests to fail on Windows.

https://github.com/usnistgov/dioptra/blob/f6d1d94a0013ab117ea1ab1f51f0d467b262a911/src/dioptra/restapi/task_plugin/service.py#L108

**To Reproduce**

Run the restapi unit tests on Windows.

**Expected behavior**
Restapi unit tests pass.

 - OS: Windows 10

Changing the above code to:

```python
prefix: Path = collection + "/" + task_plugin_name
```

enables the tests to pass.

keithmanville:
fixes #248 

This PR addresses an issue where S3 key prefixes followed the platform's path conventions causing tests that validate S3 URIs to fail. The fix is to normalize S3 prefixes to follow POSIX path conventions using `Path.as_posix()` from `pathlib`.

This PR also adds a github action for unit tests on a `windows-2022` image. `test_mlflow.py` was updated to use `pathlib` instead of `os.path` to fix cross-platform issues with the tests.

The cookiecutter tests are disabled for the `windows-2022` runner because those tests depend on a english words file for generating random passwords that does not exist on windows systems. Consider opening a new issue to address this.

### DIFF
--- a/.github/workflows/tox-tests.yml
+++ b/.github/workflows/tox-tests.yml
@@ -68,9 +68,10 @@ jobs:
         run: python3 -m tox run -e ${{ matrix.tox-testenv }}
 
   unit-tests:
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        os: ["ubuntu-20.04", "windows-latest"]
         python-version: ["3.9", "3.10"]
         tox-testenv:
           - "clean,py39-pytest-cov,report"
@@ -91,6 +92,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: install English words dictionary
+        if: ${{ matrix.os == 'ubuntu-20.04' }}
         run: sudo apt install -y wamerican
 
       - name: setup python ${{ matrix.python-version }}
@@ -104,10 +106,17 @@ jobs:
           python3 -m pip install --upgrade pip
 
       - name: get pip cache dir
+        if: ${{ matrix.os != 'windows-latest' }}
         id: pip-cache
         run: echo "dir=$(pip cache dir)" >> $GITHUB_OUTPUT
 
+      - name: get pip cache dir (Windows)
+        if: ${{ matrix.os == 'windows-latest' }}
+        id: pip-cache-win
+        run: echo "dir=$(pip cache dir)" >> $env:GITHUB_OUTPUT
+
       - name: cache dependencies
+        if: ${{ matrix.os != 'windows-latest' }}
         uses: actions/cache@v3.3.2
         with:
           path: ${{ steps.pip-cache.outputs.dir }}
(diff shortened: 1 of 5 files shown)


chisholm:
I think this PR has too many commits in it.  E.g. the mypy config and typing fixes are included.  I think that's because it is made to the main branch, so it naturally includes all commits on your branch which are not on main, which is too many.  I suspect the PR should be to dev instead.  You may just need to change the "base" branch when you create the PR.  Not sure you can do that with an already-created PR.  You may need to delete and recreate it.

jkglasbrenner:
@chisholm I edited the merge request to point at the dev branch, so now the pull request only covers the commits that @keithmanville made.

keithmanville:
Thank you both. Pointing to dev was my intent. These commits should also be squashed if/when the PR is approved. There was a lot of testing by way of github actions going on.

chisholm (approved):
When running `TaskPluginService.delete(...)` from dev branch against a real minio service (to see how the bug manifests without the mocking done in the unit tests), the relevant bits of my stack trace are:

```
  File "...\src\dioptra\restapi\task_plugin\service.py", line 117, in delete
    self._s3_service.delete_prefix(bucket=bucket, prefix=str(prefix), log=log)
  File "...\src\dioptra\restapi\shared\s3\service.py", line 55, in delete_prefix
    response: Dict[str, Any] = self._client.delete_objects(
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\Lib\site-packages\botocore\client.py", line 535, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\Lib\site-packages\botocore\client.py", line 980, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (MalformedXML) when calling the DeleteObjects operation: The XML you provided was not well-formed or did not validate against our published schema.
```

The XML request payload generated is:

```xml
<Delete xmlns="http://s3.amazonaws.com/doc/2006-03-01/" />
```

[Docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html#API_DeleteObjects_RequestBody) say the `<Object>` tag is required (nested inside `<Delete>`), which explains the validation error.  So the way the XML is designed, it seems there isn't a valid way to represent a "delete nothing" request.

On Keith's branch, it works as expected.

keithmanville:
@jkglasbrenner ready for your review.

I rebased off of dev after merging #262 and squashed commits. The windows cookiecutter tests were failing due to the password file issue, so I decided to implement a quick fix. I left this as a separate commit.

I decided to fall back to generating a list of random 4-8 character strings if the words file can't be read. This seemed to be the most minimal way to implement a fallback since only the _populate_words function needed modifications. This way the _generate_random_password function can be used as is and the fallback version of the passwords will obey the same parameters (e.g. min words, delimiter, etc.).

jkglasbrenner (approved):
Looks good, I made a few minor changes here and there before squashing and rebasing again:

- In GitHub Actions, I updated a couple things to better match the logic in the `pip-compile.yml` workflow. For example, I switched from windows-2022 to windows-latest, and created Windows and Linux specific pip caches.
- In `post_gen_project.py`, I extracted your randomized word logic from the try/except block into an if guard that checks if the dictionary file doesn't exist. I wanted to avoid having a large try/except block that would swallow all errors, even those unrelated to a FileNotFound error.

Otherwise, once the basic tests pass I'm going to merge this.

Issue #213, opened 2023-06-13
Integrate Task engine with worker backend


hbooth:
The task engine needs to be integrated with the existing worker backend infrastructure.
The existing method for running jobs needs to continue to be available until all examples have been migrated.

# Definition of Done
- [ ] Job execution workflow requires update to make the task engine runnable
- [ ] existing bash scripts updated
- [ ] REST API updated
- [ ] Possible updates to UI
- [ ] Update documentation to describe running both modes
- [ ] Merged into dev

jkglasbrenner:
Closed by #268 and #303

PR #268, opened 2023-08-26
Task engine integration into worker containers

chisholm:
Partially addresses #213 .

This PR adds a function which can be called through redis queue, to cause an experiment to run via the task engine.  In that way, it achieves task engine integration with a worker: you just have to queue the job up a bit differently.

The REST API and UI/frontend aspects of the issue are not addressed here.

### DIFF
--- a/docker/shellscripts/entrypoint-worker.m4
+++ b/docker/shellscripts/entrypoint-worker.m4
@@ -85,7 +85,7 @@ start_rq() {
   ${job_queues}"
 
   cd ${dioptra_workdir}
-  python -m ${rq_worker_module} worker\
+  PYTHONPATH="${DIOPTRA_PLUGIN_DIR}" python -m ${rq_worker_module} worker\
     --url ${rq_redis_uri}\
     --results-ttl ${rq_results_ttl}\
     ${job_queues}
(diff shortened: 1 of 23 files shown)


jkglasbrenner:
I merged #303 into this branch, now I'm going to rebase everything on `dev` and force push this branch to bring in the recent changes and resolve some of the minor conflicts that came up due to reattaching the user endpoint in `dev`.

jkglasbrenner:
I pushed an update that updates the `JobService.submit_task_engine` method to replace this, 

```python
self._queue_service.get_unlocked_by_name(queue_name)
```

with this,

```python
self._queue_name_service.get(
    queue_name, unlocked_only=True, error_if_not_found=True, log=log
)
```

This change was needed because of the queue endpoint refactor.

jkglasbrenner (approved):
I confirmed that I was able to build the containers and run the experiments in `tensorflow-mnist-classifier` using the legacy job submission system and the new "native" task engine integration. Everything seems to be in working order, so I am going to approve this request, squash the commits, and merge to `dev`.

While going through this, I noted some changes that should eventually be made, but I don't feel are enough to block the merging of this pull request. I spun those out into the following new issues: #325, #326, #327. Of these, #327 should be handled first since it's a case of state being left over between jobs.

PR #277, opened 2023-09-15
build: added support for arm64 architecture 

### LINKED ISSUE #206: Full support for the ARM64 architecture in the container images
## Description

The current container images are tied to CPUs that can only be used on the Intel architecture, and these are incompatible with Apple Silicon Macs, even when using Rosetta (Tensorflow uses SIMD chip instructions that Rosetta cannot emulate). To support ARM64 architectures, the Dockerfiles need to be updated to support multi-platform builds.

The following container images will need to be updated to support both x86-64 and ARM64 builds:

- mlflow-tracking
- nginx
- pytorch-cpu
- restapi
- tensorflow2-cpu

Since the Apple Silicon Macs are the target platform for ARM64 (Dioptra is not intended for Raspberry Pis, for example), we do not need to create multi-platform builds for `pytorch-gpu` and `tensorflow2-gpu`. This is because Macs do not have NVIDIA GPUs.

## Depends on

- #205

## Definition of Done

- [x] The dependencies in the affected Dockerfiles are updated to support ARM64 and Intel builds, ideally using the automatic ARG variables documented here: <https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope>
- [x] The `requirements.txt` lockfiles in `docker/requirements/` are regenerated for the ARM64 platform.
- [x] Run `tensorflow-mnist-classifier` end to end using the ARM64 Tensorflow worker containers on an Apple Silicon Mac.

lbarbMITRE:
Closes #206


lbarbMITRE:
Reference for tensorflow bug solution: https://forums.developer.nvidia.com/t/sklearn-cannot-allocate-memory-in-static-tls-block/158954

jkglasbrenner (approved):
The general scope of everything looks good. I had to go in and normalize the x86-64 architecture names to amd64, since that's the value for the TARGETARCH environment variable in that architecture (without the change, the Docker builds started failing on GitHub Actions). I also removed the changes to the files under `examples/`, as I took care of that in a separate commit.

Issue #214, opened 2023-06-13
Implement REST API to support Experiments and Jobs UI


hbooth:
The wireframe mockups for Experiments and Jobs is now ready to be implemented, but REST API support is needed.

# Definition of Done
- [x]  Tag Resource implemented
- [x] EntryPoint Resource implemented (including supporting resources like versions of EntryPoints, Parameters, and Parameter Values)
- [x] Group Resource implemented
- [x] Experiments updated to now include EntryPoints and Groups
- [ ] Jobs updated to use Parameter Values
- [ ] Unit Tests implemented
- [ ] passes static code analyzers
- [ ] Merged into Dev branch

keithmanville:
closed by #521

PR #281, opened 2023-10-03
refactor: reorganize and cleanup the queue restapi endpoint

jkglasbrenner:
An initial step towards cleaning and streamlining the REST API code towards being easier to maintain and build upon. This pull request applies ideas discussed with @hbooth to the queue endpoint. This also removes all the existing unit tests for the queue endpoint and replaces them with an integration style of testing that moves away from the heavy mocking. In addition, docstrings have been added to the queue service methods.

@hbooth, this isn't complete by any means, but I wanted to start seeing just how involved it would be to get the existing REST API code cleaned up in the way we discussed in advance of porting over your changes for #214. If the types of changes I made here are very different from what you were expecting to see, then we should set up a time to chat about it.

Also, it turns out that any refactoring of the endpoint code is going to break most of the associated unit tests anyway. As far as that's concerned, I think it'll be easier to do the refactoring and cleanup in stages I think, rather than trying to overhaul it all at once. This gets the ball rolling with that, but I wanted to make sure these kinds of changes aren't going to make porting your #214 content more difficult.


chisholm:
Dunno if I can stick my nose in here, but I like the code removal and simplification.  I don't like the unit test style which introduces inter-dependencies among the tests.  I think it's important to be able to cherry-pick tests to run, and having the dependencies prevents you from doing that.  E.g. if I choose to run `test_get_new_queue_by_id` by itself as:

```
>pytest -k test_get_new_queue_by_id .\tests\unit\restapi
================================================= test session starts =================================================
platform win32 -- Python 3.11.2, pytest-7.4.0, pluggy-1.2.0
rootdir: C:\dioptra
configfile: pyproject.toml
collected 101 items / 100 deselected / 1 selected

tests\unit\restapi\test_queue_management.py F                                                                    [100%]

====================================================== FAILURES =======================================================
___________________________________ TestQueueRegistration.test_get_new_queue_by_id ____________________________________

self = <tests.unit.restapi.test_queue_management.TestQueueRegistration object at 0x0000024A19DD5E50>
client = <FlaskClient <Flask 'dioptra.restapi.app'>>, state = {}, db = <SQLAlchemy engine=sqlite://>

    def test_get_new_queue_by_id(
        self, client: FlaskClient, state: dict[str, Any], db: SQLAlchemy
    ) -> None:
>       expected = state["test_register_new_queue"]["response"]
E       KeyError: 'test_register_new_queue'

tests\unit\restapi\test_queue_management.py:50: KeyError
```

This is because the `test_register_new_queue` unit test has not run to populate the `state` dict.  I prefer the unit tests to be written to just pre-populate the db (or whatever setup is required), then invoke flask to see if it gets and returns the data properly.  For example:

```python
def test_get_new_queue_by_id(
    self, client: FlaskClient, db: SQLAlchemy, dependency_injector
) -> None:
    queue_service = dependency_injector.get(QueueService)
    new_queue = queue_service.create(name="new_queue")
    response = client.get(
        f"/api/{QUEUE_BASE_ROUTE}/{new_queue.queue_id}", follow_redirects=True
    )
    assert response.status_code == 200 and response.json["queueId"] == new_queue.queue_id
```

The db would have to be wiped clean after *every* unit test though, so each one starts from a clean slate, to make sure we get deterministic repeatable results.  A class-scoped fixture would not work.

Another thing: the `test_register_new_queue` unit test tests what the endpoint returns, but doesn't test what it did (updating the database).  How about:

```python
def test_register_new_queue(
    self, client: FlaskClient, db: SQLAlchemy, dependency_injector
) -> None:
    response = client.post(
        f"/api/{QUEUE_BASE_ROUTE}/",
        json={"name": "tensorflow_cpu"},
        follow_redirects=True,
    )
    assert response.status_code == 200

    queue_service = dependency_injector.get(QueueService)
    new_queue = queue_service.get(
        response.json["queueId"],
        unlocked_only=True, error_if_not_found=False
    )

    assert new_queue.name == "tensorflow_cpu"
```
I added the part at the bottom, which ensures the queue record actually made it into the database, with the correct name.

My 2¢ :)

jkglasbrenner:
@chisholm Thanks for adding to the discussion about the testing change. The proposed new tests should have been moved under the `tests/integration` folder to better indicate their purpose. I was aiming for tests that are similar to what you might do with Postman, where you're testing that the REST API outputs follow an expected pattern instead of looking directly at the database or some other Python object that's hidden from a user.

That said, I agree there is value in having some sort of spot for independent tests that you can run one at a time instead of as a group. Since the important functionality is mostly confined to the service.py file, also having a set of more conventional unit tests targeting the individual class methods might do the trick.

jkglasbrenner:
@hbooth and @chisholm, I've modified and pushed a new version of the tests for the queue endpoint. Each test can be run on its own and no longer passes state between tests. Accordingly, I have removed the `scope="class"` argument from the pytest fixtures I originally added it to. I also resolved the mypy type checking errors, so all unit tests are passing.

The tests still follow the spirit of behavior-driven testing by validating the outputs from a sequence of actions using only the Flask testing client instead of checking the database directly or making explicit calls to the services layer (I actually had a couple of calls to the database and service layers previously, those have been replaced with client calls). The philosophy here is that the "correctness" of our API is entirely determined by whether or not it responds to a request or a sequence of requests in the expected way. In terms of technical implementation, this has the nice side-effect of decoupling our endpoint testing code from the storage and service layers. The tests should only need to be updated when something about the endpoint itself changes, either its behavior (i.e. the responses) or the endpoint URL structure.

If this all looks good, we can merge this to dev I think.

PR #303, opened 2023-10-17
feat: Add a taskEngine endpoint for task engine job submissions

### LINKED ISSUE #291: Update Job endpoint to support using the new Task Engine
**Is your feature request related to a problem? Please describe.**

Update Job endpoint to support executing Task Engine jobs directly.
Old jobs should still execute as before without changing requests to the Jobs API.
The new Jobs should be a POST request to `/job/newTaskEngine`.

**Definition of Done**
- [ ] Users can execute new (Task Engine) or legacy (MLflow) Jobs via the REST API

**Additional context**
This is a "transitional" update to the Job API to support transition to the Task Engine.

chisholm:
Fixes #291 (along with the changes in the base branch).


chisholm:
I have not added any special unit tests for the /job/newTaskEngine endpoint.  Is it worth doing when this will eventually be thrown away anyway?

jkglasbrenner (approved):
These changes overall look good to me. One small thing is that it looks like the RQ worker is invoking the job in the current working directory,

https://github.com/usnistgov/dioptra/blob/20624813e0c4eced236ca5966ed7edb3de644c68/src/dioptra/rq/tasks/run_task_engine.py#L94-L97

instead of creating and moving into a temporary directory first,

https://github.com/usnistgov/dioptra/blob/20624813e0c4eced236ca5966ed7edb3de644c68/src/dioptra/rq/tasks/run_mlflow.py#L62-L63

My understanding is that the task engine code does not set up a temporary directory to use as scratch space.

The above is minor and easy to address, and I don't want to delay getting this merged into your other branch for a full review.

> I have not added any special unit tests for the /job/newTaskEngine endpoint. Is it worth doing when this will eventually be thrown away anyway?

I'm fine with not creating a test for `/job/newTaskEngine` right now since it's a temporary endpoint. We're also refactoring all of the REST API tests anyway, so if we're going to add a simple test just to be thorough, we'll handle it after that cleanup is done with.

PR #328, opened 2023-11-29
Refactor Experiment endpoint and remove wtforms dependency

alexb1200:
Removed wtforms forms and corresponding models. Made minimal changes to unit tests.


jkglasbrenner:
Since this update changes the POST requests to no longer use forms, there are a couple places you'll need to update the Python client(s) (one being the deprecated client under `examples/scripts/client.py`). The first place is this:

https://github.com/usnistgov/dioptra/blob/1e1b5ab70381ddc294cbf55af64bb7751302d942/src/dioptra/client/_client.py#L581-L584

The second place is this:

https://github.com/usnistgov/dioptra/blob/1e1b5ab70381ddc294cbf55af64bb7751302d942/examples/scripts/client.py#L146-L149

Both need the following change:

```diff
        response = requests.post(
            self.experiment_endpoint,
-           data=experiment_registration_form,
+           json=experiment_registration_form,
        )
```

jkglasbrenner (approved):
Looks good to me

PR #330, opened 2023-12-01
Refactor the location of Flask-SQLAlchemy db object and move src/migrations folder into the dioptra package

jkglasbrenner:
This update moves the migrations folder so that it is distributed as part of the dioptra package and moves the instantiation of the Flask-SQLAlchemy object from src/dioptra/restapi/app.py to src/dioptra/restapi/db/db.py. The custom SQLAlchemy types file has also been moved into the src/dioptra/restapi/db directory. This is the first step towards consolidating all of the database models into the src/dioptra/restapi/db folder.

A `dioptra-db` command has been added to the project, which wraps `flask db` and is now the recommended tool for interacting with Dioptra's database. The project documentation has been updated to reflect this.

The following also needed to be updated/changed in response to the above refactor:

- The REST API Dockerfile was updated so that it no longer has a copying step for the migrations directory since it is now included in the `dioptra` package.
- The entrypoint-restapi.m4 script was modified to invoke `dioptra-db autoupgrade` every time the container starts.
- The database upgrade step in the `init-deployment.sh` script in the the cookiecutter deployment template was removed since `dioptra-db autoupgrade` is now invoked during container startup.
- All import statements for the db object were changed from `from dioptra.restapi.app import db` to `from dioptra.restapi.db import db`.
- The `FlaskTestServer.upgrade_db` method used by some of the REST API tests was updated to use the new `diotpra-db` command. The arguments passed to this method in the tests have been updated to no longer explicitly pass the path to the migrations folder, since the `dioptra-db` script can autodetect it.
- Project configuration scripts were updated to properly handle the files in the `src/dioptra/restapi/db/alembic` folder.


jkglasbrenner:
@keithmanville Wanted to just raise this to your attention since it's not something on an Issue card. This is with an eye towards consolidating all of the SQLAlchemy models into the same place (the `src/dioptra/restapi/db` folder), which will make it easier to implement all of the new tables we have planned.

PR #331, opened 2023-12-02
fix: run task engine jobs in a temporary directory

### LINKED ISSUE #327: Create and run new task engine jobs in a temporary folder
The script that the RQ worker uses to natively launch a new task engine job runs in the worker's current working directory. This can lead to files and folders being left behind after a job runs, which can interfere with the execution of future jobs. The script should use temporary folders in a manner similar to `dioptra.rq.tasks.run_mlflow`:

https://github.com/usnistgov/dioptra/blob/91e9aed8a4e94ffcebe60b41e7be8901a05dde5e/src/dioptra/rq/tasks/run_mlflow.py#L62

## Definition of Done

- [ ] The jobs started by `dioptra.rq.tasks.run_task_engine` create and use a temporary folder as the working directory while executing `_run_experiment`.
- [ ] Jobs submitted to the new task engine job submission endpoint run to completion without any change in behavior.

chisholm:
Fixes #327 .

### DIFF
--- a/src/dioptra/rq/tasks/run_task_engine.py
+++ b/src/dioptra/rq/tasks/run_task_engine.py
@@ -15,6 +15,8 @@
 # ACCESS THE FULL CC BY 4.0 LICENSE HERE:
 # https://creativecommons.org/licenses/by/4.0/legalcode
 import os
+import pathlib
+import tempfile
 from typing import Any, Mapping, MutableMapping, Optional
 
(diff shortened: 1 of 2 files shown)


jkglasbrenner (approved):
Ran a couple jobs in a full deployment and confirmed the temporary folders work there. Looks good to me, going to merge to dev.

Issue #313, opened 2023-10-27
refactor: reorganize and cleanup the Job restapi endpoint


keithmanville:
Refactor the Job endpoint, following the pattern from #281 

### Definition of Done
- [ ] Job endpoint is refactored
- [ ] Passes all tests from #310

jkglasbrenner:
Closed by #355

PR #339, opened 2023-12-08
Refactor jobs restapi to remove wtforms

alexb1200:
Addresses issue #313, removes the wtforms dependency, refactors the controller to accept the job form, and changes _client.py and client.py to reflect those changes. Makes minimal changes to the jobs unit tests to reflect those updates.


jkglasbrenner:
While doing a code review, I found that the Swagger documentation wasn't coming out right, even after using `load_only` and `dump_only` on the Marshmallow Schema. The issue is that `form_schema=` argument for the `@accepts` decorator is processed differently, and basically it ignores `load_only` and `dump_only` when generating the Swagger Docs.

I'm working on a way to fix this without needing to patch the `flask_accepts` package. Somewhat related to the above, I'm also fixing things so that you can access the form submitted to POST using `request.parsed_form` and so that everything can be handled with just `JobSchema` so that we can delete `JobSubmitSchema`. When I get these worked out, I'll push the update to this branch.

Finally, I want to verify that the file upload works not just in the unit tests but also when I build and run the full deployment.

jkglasbrenner:
I've implemented the changes so that we can use `form_schema=` with a Marshmallow Schema that makes use of `dump_only` and `load_only`, and allows us to generate the correct Swagger docs. Here's what the decorators on the post function look like now:

https://github.com/usnistgov/dioptra/blob/6d12f9c6e33fcc618d2d550bda4ad7543baa857f/src/dioptra/restapi/job/controller.py#L72-L78

The new part is the `as_parameters_schema_list` function, which takes the Marshmallow schema and converts it to the format needed in the `as_api_parser` function. The `as_parameters_schema_list` respects the `dump_only` and `load_only` settings, so the documentation for the POST endpoint now comes out correct:

image

Also, just to document a "gotcha" with the difference between how the requests package works and the Flask test client works:

I had to go back and restore the `files=` argument in the `requests.post` method used in the Dioptra client, as it's the only way to specify a file for uploading:

https://github.com/usnistgov/dioptra/blob/6d12f9c6e33fcc618d2d550bda4ad7543baa857f/src/dioptra/client/_client.py#L686-L691

It _is_ different from the Flask test client, which doesn't use the `files=` argument:

https://github.com/usnistgov/dioptra/blob/6d12f9c6e33fcc618d2d550bda4ad7543baa857f/tests/unit/restapi/job/test_controller.py#L151-L157

jkglasbrenner:
As long as the GitHub Actions tests all pass, this is ready to merge.

In terms of merge order, I'm going to review and merge the fix in #342 **first** to dev, since it also touches the Job endpoint. Then I'm going to resolve any conflicts that introduces in this branch, and once that's handled, merge this in.

jkglasbrenner (approved):
Rebased this branch after merging #342 to dev, all unit tests and code quality checks passed. Looks good to me for a merge.

Issue #312, opened 2023-10-27
refactor: reorganize and cleanup the Experiment restapi endpoint


keithmanville:
Refactor the Experiment endpoint, following the pattern from #281 

### Definition of Done

- [x] Dependency on WTForms is removed.
- [x] Experiment endpoint is refactored
- [x] Passes all tests from #310

jkglasbrenner:
Closed by #341

PR #341, opened 2023-12-12
Refactor experiments endpoint to breakup functionality

alexb1200:
Continues the refactor for #312, breaking up the functionality of the Experiment service into different classes. 

Should be merged after PR #339 and #328, will fail unit tests and mypy checks until then.

### DIFF
--- a/src/dioptra/restapi/experiment/controller.py
+++ b/src/dioptra/restapi/experiment/controller.py
@@ -18,23 +18,19 @@
 from __future__ import annotations
 
 import uuid
-from typing import List, Optional
+from typing import Any, cast
 
 import structlog
-from flask import jsonify, request
-from flask.wrappers import Response
+from flask import request
 from flask_accepts import accepts, responds
 from flask_login import login_required
 from flask_restx import Namespace, Resource
 from injector import inject
 from structlog.stdlib import BoundLogger
 
-from dioptra.restapi.utils import slugify
-
-from .errors import ExperimentDoesNotExistError
 from .model import Experiment
-from .schema import ExperimentSchema
-from .service import ExperimentService
+from .schema import ExperimentSchema, IdStatusResponseSchema, NameStatusResponseSchema
+from .service import ExperimentNameService, ExperimentService
 
 LOGGER: BoundLogger = structlog.stdlib.get_logger()
 
@@ -55,7 +51,7 @@ class ExperimentResource(Resource):
 
     @login_required
     @responds(schema=ExperimentSchema(many=True), api=api)
-    def get(self) -> List[Experiment]:
+    def get(self) -> list[Experiment]:
         """Gets a list of all registered experiments."""
         log: BoundLogger = LOGGER.new(
             request_id=str(uuid.uuid4()), resource="experiment", request_type="GET"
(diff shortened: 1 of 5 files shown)


jkglasbrenner:
Just rebased this branch on dev after merging #328 and #339 and resolved the conflicts. On my local machine, the restapi unit tests and code checks are now passing. This is now ready for me to do a code review.

jkglasbrenner:
I did a review and made some additional edits, which I will be pushing shortly. I'll document the spirit of the changes below so they can be applied to the future refactoring of the job and task plugins endpoints.

## general changes

**type annotations**

There's a couple of things that I did to clean up type annotations as we continue the refactors:

- In most situations, we don't need to do inline annotations, i.e. `foo: int = my_function(bar)`, if we're also providing type annotations on all of the methods/functions in the code base. I take responsibility for originally setting the pattern of putting inline type annotations everywhere, it's led to visual clutter. mypy and IDEs like VSCode are able to infer the types just fine without them, so I want to start removing inline annotations where we can.
- I switched type annotations like `Optional[Experiment]` to `Experiment | None` and `Union[int, float, str]` to `int | float | str`. This is permitted in Python 3.9 so long as the import statement `from __future__ import annotations` is at the top of the file. You can also use Python primitive types once the future import is activated, i.e. `list[str]` and `dict[str, Any]` instead of `List[str]` and `Dict[str, Any]`. This newer style is more readable and easier to type.

## controller.py changes

**slugify usages**

I've noticed some inconsistency with where `slugify` gets called in endpoints. I want to standardize on only calling `slugify` in the service layer and not the controller layer. So, any instances where I saw `slugify` being used in controller.py, I moved them into the appropriate method in service.py.

**Raising errors**

The queue endpoint refactor established a new pattern where the controller itself does not raise errors. Instead, errors should be raised in the service layer. The way this was implemented was by adding the `error_if_not_found` argument to the `get` methods in the service layer.

**Naming convention for parsed request parameters**

I'd like to standardize on always using `parsed_obj` to refer to the parameters attached to a request. So, `parsed_obj = request.parsed_ob`, `parsed_obj = request.parsed_form`, etc.

## schema.py changes

**Docstrings**

Since the schema fields contain descriptions, we don't need to duplicate them in the class docstring. For this reason, I've removed the `Attributes:` section from the class docstrings for any of the Marshmallow Schemas.

In addition, if any of the restructured text aliases declared in the top-level docstring (words surrounded by bars, i.e. `|Experiment|`) are no longer being referenced in any of the other docstrings in the file, they can be removed.

## service.py changes

**error_if_not_found**

Similar to the `QueueService` and `QueueNameService`, in the `get` methods I added an `error_if_not_found` argument with a default value of False. If the value is True, then if the experiment is not found the method raises an error. If False, then the method simply returns a None. The code in the ExperimentService that handles this looks as follows:

```python
if experiment is None:
    if error_if_not_found:
        log.error("Experiment not found", experiment_id=experiment_id)
        raise ExperimentDoesNotExistError

    return None
```

**reply vs response terminology**

This is admittedly minor. In most of the code base we use request/response to describe client/server communications, so I updated variables named `reply` to `response` to keep things consistent.

**Method names on the service layer**

Similar to the convention established in the queue endpoint refactor, since the experiment services have been separated into the `ExperimentService` and `ExperimentNameService`, I updated the method names to remove the word "experiment" (since its implied already, i.e. `ExperimentService.delete` instead of `ExperimentService.delete_experiment`) or "by_id". So, `delete_experiment` becomes `delete`, `get_by_id` becomes `get`, and so on.

**Removed extra methods where it made sense**

The `create_mlflow_experiment` method was only used in one place and would only be used to create a new Experiment resource, so I moved the method's code back into the body of `ExperimentService.create` and deleted the `create_mlflow_experiment` method.

The `check_if_experiment_name_exists` static method was removed since you can equivalently do `if self._experiment_name_service.get(experiment_name, log=log) is not None`.

jkglasbrenner (approved):
Looks good to me after incorporating changes.

PR #342, opened 2023-12-14
fix: fix race condition in mlflow job submission

### LINKED ISSUE #326: Fix restapi/worker race condition in the legacy job submission endpoint
The fix for the race condition included in `dioptra.restapi.job.service.JobService.submit_task_engine` should be ported to `dioptra.restapi.job.service.JobService.submit`.

## Definition of Done

- [x] The fix to the race condition in `submit_task_engine` is ported to the legacy `submit` method.
- [x] Jobs submitted to the legacy job submission endpoint run to completion without any change in behavior.

chisholm:
Fixes #326 .

`RQService.submit_mlflow_job()` no longer returns a value.  It was only used to get the rq job ID, and since callers no longer need that, there is no longer a need for that method to return a value.  The `submit_task_engine_job()` method also doesn't return a value, so those two job submission methods now match up.  Some of the unit tests were deleted because they were checking aspects of the return value, which is no longer applicable.


jkglasbrenner (approved):
Looks good to me

PR #348, opened 2023-12-18
Add support for updating deployment configurations in-place

### LINKED ISSUE #126: Utility for updating deployment configurations in-place
## Overview

Users would like to be able to edit their deployment configurations without needing to re-enter answers to the questions after they've been generated using the cookiecutter template. They also would like to keep the same passwords they've already generated.

## Definition of Done

TBD

jkglasbrenner:
Closes #126


jkglasbrenner:
This is a work in progress. I've made most of the necessary changes to the Cookiecutter template so that we can leverage the Cruft tool. The main change is in how the deployment passwords are generated and stored. Here we make use of a `.env` file, which Docker Compose automatically detects and sources, to hold the passwords.

I've also begun updating the documentation to explain how to use [Cruft](https://cruft.github.io/cruft/) to manage and update the deployment configuration. Cruft is basically just a wrapper around Cookiecutter, with some nice features for syncing template updates in-place.

jkglasbrenner:
Merged in #401 and #402, and then rebased on `dev`.

jkglasbrenner:
@keithmanville I've made some additional adjustments to get this over the finish line. The main updates are:

- I noticed that generating the `docker-compose.yml` file when we run `init-deployment.sh` was no longer necessary and we can just render it as part of the template, so I switched it back and deleted the `generate_docker_templates.py` file (I moved the dataset directory validation logic into the `pre_gen_project.py` hook).
    - I think we moved in that direction when we thought that users might need to edit the file. Since we now set `datasets_directory` as a cookiecutter variable and we're advocating the use of the override file, that's no longer an issue.
- I noticed that the template logic for handling `datasets_directory` was producing a 2nd `volumes` section in each service block in the `docker-compose.yml` file. It seemed to work, but it looked awkward. I updated the template logic so that all volume mounts are colocated under each service block.
- I went through the documentation files and fixed up inconsistencies.
    - One notable change is everything now says to mount the datasets directory to `/dioptra/data` in the worker containers, which is consistent with the examples we provide. This gets rid of references to `/datasets` and `/dioptra/datasets`, which was confusing.
    - A note/comment has been added to all the code blocks that invoke a command that needs to know which GitHub branch you want to point at (`cruft` commands and `init-deployment.sh` commands). The comment states that the user can substitute `"main"` with another branch name.
    - I copied @bjpatrick's writeup into the sphinx docs so its available there too.
    - I updated the section explaining how to change the number of GPUs assigned to a worker to indicate this should be done in the override file.
- I removed some vestigial code/settings that are no longer needed. The logic for replacing the `$((BASEDIR))` symbol was removed because we're no longer using it. I also removed `version: 2.4` from the top of the `docker-compose.yml` files, since that is no longer part of the spec.

I did an end-to-end test with the template on `tensorflow-mnist-classifier`, and everything worked as expected. I also tested out changing the number of worker containers to double check the templates render correctly, and they are fine. Finally, I also set `datasets_directory = ""` in the template and then used the override file instead, which also worked. So this looks to be good to go, I'm going to get this merged to dev so we can all start using it to manage our local and shared deployments going forward.

PR #350, opened 2024-01-01
Remove AWS CLI

### LINKED ISSUE #325: Remove AWS CLI dependency from Dioptra containers
The implementation for natively running jobs using the new task engine includes code for downloading the task plugins without relying on the third-party AWS CLI tool currently being packaged in the containers. The task plugin downloading functionality should be back-ported to the legacy job submission process so that the AWS CLI can be fully removed from the containers.

## Definition of Done

- [x] The `setup_task_plugins` method is applied to `dioptra.rq.tasks.run_mlflow`
- [x] The functions originally used in the shellscript `docker/shellscripts/run-mlflow-job.m4` to download the task plugins are removed
- [x] The AWS CLI tool is fully removed from all Dockerfiles
- [x] Jobs submitted to the legacy job submission endpoint run to completion without any change in behavior.

chisholm:
Fixes #325 .

I have tested the tensorflow-cpu worker but none of the others, except to verify that they built.


jkglasbrenner (approved):
Looks good to me. @keithmanville confirmed earlier that the tensorflow GPU container worked the same as the CPU version. The changes in the pytorch containers are identical, so that gives me enough confidence that they will work the same (famous last words, I suppose).

I'm going to rebase, squash and merge this to dev.

PR #354, opened 2024-01-04
fix: created pytest fixture to solve mlruns bug

### LINKED ISSUE #337: bug: Tests to not properly cleanup mlruns folder
There is an issue that the mlruns folder is not properly cleaned up. This can cause subsequent runs of the tests to fail.

Modify the test so that when it runs, the mlruns folder gets placed in a temporary location that will be deleted.

Attempt to fix this with a new fixture that can be applied to each test without a lot of boilerplate.

lbarbMITRE:
Closes #337

### DIFF
--- a/tests/unit/restapi/conftest.py
+++ b/tests/unit/restapi/conftest.py
@@ -255,3 +255,12 @@ def wait_for_healthcheck_success(client: RequestsSession, timeout: int = 10) -> 
             healthcheck = StubResponse()
 
         time_elapsed = time.time() - time_start
+
+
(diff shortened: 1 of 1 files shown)


alexb1200 (approved):
Ran the unit tests on my local machine, and no stray mlruns folder was left. LGTM.

jkglasbrenner:
Also looks good to me. I moved the fixture into a different location in the conftest.py file (keep fixtures first in file, non-fixture functions are at the bottom) and I moved the import to the top of the file, otherwise logic is the same. Getting ready to merge into dev.

PR #368, opened 2024-01-19
Jobs endpoint integration tests

keithmanville:
Rebased branches to dev to replace PR #362


jkglasbrenner:
To capture this here, one reason the unit test is failing is because the `job_form_request` fixture is using snake case for the dictionary keys, but they need to be camel case (well, technically "Pascal case"). This change is required because of the refactor that removed WTForms that was recently merged to the dev branch.
 
So it would look like this:

```python
@pytest.fixture
def job_form_request(workflow_tar_gz: BinaryIO) -> Dict[str, Any]:
    return {
        "experimentName": "mnist",
        "queue": "tensorflow_cpu",
        "timeout": "12h",
        "entryPoint": "main",
        "entryPointKwargs": "-P var1=testing",
        "workflow": (workflow_tar_gz, "workflows.tar.gz"),
    }
```

However, fixing this reveals another error that comes later. One of the steps in the job submission code in the service layer is to upload a file to S3, as well as place something on the Redis queue. Both of these are external services that are not active when running the unit test. Hence, they will fail.

We will need to mock those calls in the service layer to avoid this issue.

jkglasbrenner:
Managed to get this working using MonkeyPatch with mocks for the Redis queue and S3 file uploading services. Everything is passing and looks good, preparing to merge to dev now.

PR #369, opened 2024-01-23
Refactor: Move slugify calls to service layer

alexb1200:
There were a few stray places in the controller layer in the queue endpoint that had calls to slugify instead of being in the service layer. In addition some methods in the service layer like queue locks were not slugifying their name reference. This PR cleans that up, and slugifyies every name reference in the service layer.

### DIFF
--- a/src/dioptra/restapi/queue/controller.py
+++ b/src/dioptra/restapi/queue/controller.py
@@ -28,8 +28,6 @@ from flask_restx import Namespace, Resource
 from injector import inject
 from structlog.stdlib import BoundLogger
 
-from dioptra.restapi.utils import slugify
-
 from .model import Queue
 from .schema import IdStatusResponseSchema, NameStatusResponseSchema, QueueSchema
 from .service import QueueNameService, QueueService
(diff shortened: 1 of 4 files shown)


jkglasbrenner (approved):
I agree that the controller wasn't a good spot to handle this operation. The service layer makes more sense, although I noticed it had to be included in a lot of places. If we add another method to the service layer, it could be easy to forget, which would lead to the inevitable bug.

Your suggestion gave me the idea to actually push this into the Marshmallow schema by using the `post_load` decorator. [It's actually the example provided for this decorator in the Marshmallow docs](https://marshmallow.readthedocs.io/en/latest/extending.html#pre-processing-and-post-processing-methods). It lets us write the slugify call just once, and it ensures that the `parsed_obj` dictionary always has the `"name"` field slugified by the time we would access it. No more need to remember it any time we add a new method to the service layer 🙂

I added a new unit test to explicitly validate that the `@post_load` decorator is doing its job. Everything looks good here, so I'm going to go ahead with the merge.

PR #371, opened 2024-02-09
Remove Flask-WTF and wtforms dependencies from the REST API app

jkglasbrenner:


jkglasbrenner:
I built the containers and ran the tensorflow-mnist-classifer example to verify that everything still works as expected after removing this dependency. Everything worked. Preparing to merge to dev.

PR #393, opened 2024-02-16
fix: changed links in paging schema to URLs

keithmanville:


src/dioptra/restapi/v1/schemas.py
```diff
@@ -94,15 +94,15 @@ class BasePageSchema(Schema):
         attribute="is_complete",
         metadata=dict(description="Boolean indicating if more data is available."),
     )
-    first = fields.String(
+    first = fields.Url(
         attribute="first",
         metadata=dict(description="URL to first page in results set."),
     )
```
jkglasbrenner, line 100: Let's include `relative=True` so that we have the option of providing partial URLs like `/api/v1/<entity>/...`:

```suggestion
    first = fields.Url(
        attribute="first",
        metadata=dict(description="URL to first page in results set."),
        relative=True,
    )
```

The same should be done for `next` and `prev` as well.

jkglasbrenner (changes requested):

PR #394, opened 2024-02-16
feat: updated v1 base schemas

keithmanville:


src/dioptra/restapi/v1/schemas.py
```diff
@@ -24,63 +24,110 @@
...
+            metadata=dict(description="Tags associated with the {name} resource."),
+            many=True,
+            dump_only=True,
+        ),
+    }
+
+    if not snapshot:
+        schema.pop("snapshotId")
+        schema.pop("user")
+        schema.pop("createdOn")
```
jkglasbrenner, line 93: We should keep the `createdOn` field for regular resources.

src/dioptra/restapi/v1/schemas.py
```diff
@@ -94,15 +141,15 @@ class BasePageSchema(Schema):
         attribute="is_complete",
         metadata=dict(description="Boolean indicating if more data is available."),
     )
-    first = fields.String(
+    first = fields.Url(
```
jkglasbrenner, line 144: This change is being handled in #393, so there will be a merge conflict to address.

src/dioptra/restapi/v1/schemas.py
```diff
@@ -24,63 +24,110 @@
...
+        schema.pop("snapshotId")
+        schema.pop("user")
+        schema.pop("createdOn")
+        schema.pop("lastModifiedOn")
+        schema.pop("latestSnapshot")
+
+    return Schema.from_dict(schema, name=f"{name}BaseSchema")
+
+
+def generate_base_resource_ref_schema(name: str, snapshot: bool) -> type[Schema]:
```
jkglasbrenner, line 100: The meaning of the `snapshot` parameter is slightly different in `generate_base_resource_schema` and this function. I noticed that in the `schema.py` file for the same entity (such as queues), we end up using `snapshot=True` for the former and `snapshot=False` for the latter, which might lead to confusion. I propose we change it to `keep_snapshot_id` here to make the intent clearer. (We can leave it as `snapshot` in the other function, it makes sense there).

```suggestion
def generate_base_resource_ref_schema(
    name: str, keep_snapshot_id: bool
) -> type[Schema]:
```

jkglasbrenner (changes requested):
I like the idea of the factory functions here, they let us build the different versions of the base resource and reference Schemas using just two functions. Let's go with this.

I did a code review and am requesting some changes before we merge to `dev`. The suggestions are minor, but let me know if you have any questions.

src/dioptra/restapi/v1/schemas.py
```diff
@@ -24,63 +24,110 @@
 from dioptra.restapi.v1.users.schema import UserRefSchema
 
 
-def generate_base_resource_schema(name: str) -> type[Schema]:
+def generate_base_resource_schema(name: str, snapshot: bool) -> type[Schema]:
```
alexb1200, line 27: Given that Tags are a somewhat special object, I'm wondering if there should be a generic generate_base_schema function to handle them as well, or if they should just be handled as a special case alone.  
Tags require   
- id: int,
- user: UserRef,
- group: GroupRef,
- created_on: Timestamp,
- last_modified_on: Timestamp,


A potential pattern for the builder could a series of functions which append certain keys:values to a base dict depending on the resource. That is pulling out the functionality of the snapshot if statement into it's own function and creating another one corresponding to resource, tag, snapshot etc. 

As it stands, I can see it overcomplicating it for no real benefit. Maybe as an alternative we can just add to the function's docstring something like `Builder function for all resource snapshot and resource base schemas. Tags are a separate entity type and are handled in its respective schema.py file.`

alexb1200: See #399 for Tags schemas.

src/dioptra/restapi/v1/schemas.py
keithmanville: fixed

src/dioptra/restapi/v1/schemas.py
keithmanville: merged #393 to `dev` and rebased this branch

src/dioptra/restapi/v1/schemas.py
keithmanville: made this change.

I also opted to make `keep_snapshot_id` default to `False` because I think the only place it will be `True` is in the `SnapshotRef` in `schemas.py`. It's use for the Ref types for each of the other endpoints should be `False`.

I kept `snapshot` without a default because we will have a mix of `Resource` and `ResourceSnapshot` types in the endpoints and I want it to be clear which it is.

src/dioptra/restapi/v1/schemas.py
keithmanville: keeping this as is for now. We can revisit if we find that there are more "special cases" that could be handled in a more general way.

keithmanville:
@jkglasbrenner @alexb1200 ready for you to take another look. I addressed the feedback and squashed/rebased, so if you are happy with the changes it should be ready to merge.

src/dioptra/restapi/v1/queues/schema.py
```diff
@@ -19,37 +19,25 @@
...
     PagingQueryParametersSchema,
     SearchQueryParametersSchema,
+    generate_base_resource_ref_schema,
     generate_base_resource_schema,
 )
 
+QueueRefBaseSchema = generate_base_resource_ref_schema("Queue")
+
 
 class QueueRefSchema(Schema):
```
alexb1200, line 34: The class definition `QueueRefSchema(Schema):` should be `QueueRefSchema(QueueRefBaseSchema):` I believe.

alexb1200:
I took a look at the code and it all looks good to me minus that one typo I mentioned. Once that's resolved I think it should be good to go. As a side note, we should capture the fact that user, group, and tag refs are special and do not use the resource ref factory.

jkglasbrenner (approved):
Looks good to me, preparing to merge.

PR #400, opened 2024-02-22
refactor: move imports to body of generate_base_resource_schema to avoid circular imports

alexb1200:
The body of `generate_base_resource_schema` now imports reference schemas in the body of the function. All schemas will typically import from the base schema, and the base schema imports some reference schemas. This can result in a circular import if done at the top of the file. Moving to the body of the relevant function resolves this issue.


jkglasbrenner (approved):
LGTM

PR #403, opened 2024-02-28
refactor: run legacy mlflow jobs directly; create dioptra-worker tool

### LINKED ISSUE #296: Create dioptra-worker command for launching a Dioptra worker without a container
Make it possible to launch a worker instance without docker to support Dioptra developer workflows and experiment prototyping workflows.

## Basic requirements

The general ask is to replicate the current behavior of the `start_rq()` function in the [entrypoint-worker.m4](https://github.com/usnistgov/dioptra/blob/a76b919310267893a392bdb7f5b5e916909aba72/docker/shellscripts/entrypoint-worker.m4) script with a `dioptra-worker` command that is provided by the dioptra package.

A few things that will need to be done to accomplish this (this may not be exhaustive, but it should hit on the most important points):

- Launch the [RQ workers](https://python-rq.org/) using the `dioptra-worker` command and pass any configurations as appropriate. Currently the RQ workers are launched using [dioptra.rq.cli.rq](https://github.com/usnistgov/dioptra/blob/d9079d87d807fee80bd3ce1d30d5a6765d04bada/src/dioptra/rq/cli/rq.py), which just calls the [rq package's Click interface](https://github.com/rq/rq/blob/master/rq/cli/cli.py).
- **For legacy jobs:** Port the remainder of the [run-mlflow-job.m4 script](https://github.com/usnistgov/dioptra/blob/26ce1c85cb8c233e31bc72ee0f30e96db0c05b39/docker/shellscripts/run-mlflow-job.m4) into [rq/tasks/run_mlflow.py](https://github.com/usnistgov/dioptra/blob/26ce1c85cb8c233e31bc72ee0f30e96db0c05b39/src/dioptra/rq/tasks/run_mlflow.py).
    - By extension, the [run-mlflow-job.m4 script](https://github.com/usnistgov/dioptra/blob/26ce1c85cb8c233e31bc72ee0f30e96db0c05b39/docker/shellscripts/run-mlflow-job.m4) script will be able to be deleted.
    - Leverage/adapt [dioptra.rq.cli.mlflow](https://github.com/usnistgov/dioptra/blob/d9079d87d807fee80bd3ce1d30d5a6765d04bada/src/dioptra/rq/cli/mlflow.py) for invoking the legacy job task.
- Figure out if there is any additional functionality in the worker containers, for example [Dockerfile.tensorflow2-cpu](https://github.com/usnistgov/dioptra/blob/26ce1c85cb8c233e31bc72ee0f30e96db0c05b39/docker/Dockerfile.tensorflow2-cpu), that needs to be encapsulated in the `dioptra-worker` command
- Update the `start_rq()` function of [entrypoint-worker.m4](https://github.com/usnistgov/dioptra/blob/a76b919310267893a392bdb7f5b5e916909aba72/docker/shellscripts/entrypoint-worker.m4) to launch the worker using the `dioptra-worker` command.

When this is all done, it should be possible to start up a Dioptra worker without a container so long as it can see the Redis port.

chisholm:
Fixes #296 .

The old run-mlflow-job shell script is removed.  MLproject jobs are run directly using the mlflow Python API.  So there is one less process fork, and the code is simpler.  All preparation, e.g. workflow download/untar, task plugin download, is done in Python.  The `run_mlflow_task()` and `run_task_engine_task()` functions are stylistically more aligned now, since some of their requirements and activities are identical.

Created a dioptra-worker CLI tool, which passes through to the rq worker command.  The worker entrypoint script is changed to start the rq worker using this tool.  The tool also does a little environment sanity check, to error out sooner than if we waited for a job submission and then crashed.


chisholm:
Some path sanitization unit tests fail on the Ubuntu VMs, but work on Windows VMs.  Path sanitization depends on how Python interprets the path: the code tries to identify unsafe things in it.  How Python interprets a path is apparently platform-dependent.  I ran the unit tests on my Windows laptop, so I never saw the differences on Linux.  It has proved more complicated than I thought. :(

jkglasbrenner:
Note, the sudden failure of the user endpoint unit tests is due to an edge case involving leap years and how a password expiration timestamp was being calculated, see #405. I'm fixing it in #406, waiting for it to pass tests and get it merged to `dev` soon.

jkglasbrenner:
I created and merged #407 into dev to address an issue where the locally run worker does not have all necessary dependencies to run `tensorflow-mnist-classifier` end-to-end if you start the worker from a virtual environment created using one of the tensorflow requirements files in `requirements/`.

Also, this is a documentation thing, but MacOS users will likely need to set the environment variable `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` in order to run the worker, otherwise you get forked process errors. It has to do with certain system calls that Macs make that are distinct from Linux, such that forking behaves differently. Setting that variable seems to solve the problem for Macs.

jkglasbrenner (approved):
I tested this branch on my M1 Macbook and things worked as expected.

Looks good to me, rebasing, squashing, and preparing to merge to `dev`.

PR #410, opened 2024-03-01
feat: implement v1 schemas for artifacts

etrapnell-nist:


alexb1200:
Looks good to me!

alexb1200 (commented):
I think this is ready for a quick once over.

PR #419, opened 2024-03-18
feat: added entrypoint and entrypoint parameter schemas

lbarbMITRE:


lbarbMITRE:
The style check fails as black version 24.2.0 caused some of the parameters to be defined as tuples. This will change the data type to something not intended.

keithmanville:
included in #420

PR #424, opened 2024-03-21
V1 groups schema

andrewhand:
Rebased off most recent dev changes before creating PR.

### DIFF
--- a/src/dioptra/restapi/v1/groups/controller.py
+++ b/src/dioptra/restapi/v1/groups/controller.py
@@ -14,7 +14,7 @@
 #
 # ACCESS THE FULL CC BY 4.0 LICENSE HERE:
 # https://creativecommons.org/licenses/by/4.0/legalcode
-"""The module defining the endpoints for Plugin resources."""
+"""The module defining the endpoints for Group resources."""
 from __future__ import annotations
 
 import uuid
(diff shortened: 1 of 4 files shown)


src/dioptra/restapi/v1/groups/schema.py
```diff
@@ -36,3 +37,121 @@ class GroupRefSchema(Schema):
...
+        metadata=dict(
+            description="A url pointing to a list of Resourcess associated with the Group."
+        ),
+        relative=True,
+        dump_only=True,
+    )
+    tags = fields.Url(
+        attribute="tags",
+        metadata=dict(
+            description="A url pinting to a list of Tags associated with the Group."
```
alexb1200, line 153: small typo in the description `"A url pinting..."` should read `"A url pointing..."`

src/dioptra/restapi/v1/groups/schema.py
```diff
@@ -19,14 +19,15 @@
...
+from dioptra.restapi.v1.users.schema import UserRefSchema
+
 
 class GroupRefSchema(Schema):
     """The reference schema for the data stored in a Group resource."""
 
     id = fields.Integer(
         attribute="id",
         metadata=dict(description="ID for the Group resource."),
-        dump_only=True,
```
alexb1200, line 29: Should `dump_only` be removed here? I think it's true for every field in every ref schema, as all of those values are coming from, created, and modified on different objects. It's marked as such in the [base resource](https://github.com/usnistgov/dioptra/blob/d1303030f8a9435fb325d7765d77476fc4291bec/src/dioptra/restapi/v1/schemas.py#L35) but not in the [base ref](https://github.com/usnistgov/dioptra/blob/d1303030f8a9435fb325d7765d77476fc4291bec/src/dioptra/restapi/v1/schemas.py#L106). Is there a misunderstanding on my part @keithmanville ?

andrewhand: > Should `dump_only` be removed here? I think it's true for every field in every ref schema, as all of those values are coming from, created, and modified on different objects. It's marked as such in the [base resource](https://github.com/usnistgov/dioptra/blob/d1303030f8a9435fb325d7765d77476fc4291bec/src/dioptra/restapi/v1/schemas.py#L35) but not in the [base ref](https://github.com/usnistgov/dioptra/blob/d1303030f8a9435fb325d7765d77476fc4291bec/src/dioptra/restapi/v1/schemas.py#L106). Is there a misunderstanding on my part @keithmanville ?

I think you're right none of the Ref Schema's should be marked as dump only.

alexb1200 (commented):
I think this all looks good to me. I left a comment to make sure I understand the purpose of an endpoint, but I expect that it is written correctly.

src/dioptra/restapi/v1/groups/controller.py
```diff
@@ -0,0 +1,173 @@
...
+    @responds(schema=IdStatusResponseSchema, api=api)
+    def delete(self, id: int):
+        """Deletes a Group resource."""
+        log = LOGGER.new(
+            request_id=str(uuid.uuid4()), resource="Group", request_type="DELETE", id=id
+        )
+        log.debug("Request received")
+
+    @login_required
+    @accepts(schema=GroupMutableFieldsSchema, api=api)
```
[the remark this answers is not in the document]
keithmanville: Good catch. I was originally intending to modify group members via `/groups/{id}` but then we added the `/groups/{id}/members` endpoint for managing that. I changed `members` to dump only but it should be in `GroupSchema` not `GroupMutableFieldsSchema`.

PR #428, opened 2024-03-29
feat: added PluginTaskParameterSchema

keithmanville:


src/dioptra/restapi/v1/plugins/controller.py
```diff
@@ -112,7 +112,7 @@ class PluginIdFilesEndpoint(Resource):
     @accepts(PluginFileGetQueryParameters, api=api)
     @responds(schema=PluginFilePageSchema, api=api)
     def get(self, id: int):
-        """Gets the PluginFile resources for a Plugin resource."""
+        """Gets a list of al PluginFile resources for a Plugin resource."""
```
alexb1200, line 115: Small typo, should read `"""Gets a list of all PluginFile resources for a Plugin resource."""`

jkglasbrenner (changes requested):
Need to change data type for `parameterTypeId` from string to integer.

PR #430, opened 2024-04-05
(EXAMPLE) feat: add pytest flag to run v1 REST API tests

keithmanville:
This feature adds a `--run-v1` flag to pytest in `tests/conftest.py` to enable running of v1 API tests. To mark a test as v1, annotate the test function with `@pytest.mark.v1`.

We intend to use this feature during the development of v1 of the API. We expect tests to fail while the service layer is not implemented so v1 tests should be excluded by default.

To run only v1 tests use the command below:
```
python -m tox run -e py39-pytest -- tests/unit/restapi/v1 --run-v1
```

A copy of the v0 Queue tests has been placed in the v1 folder with the API_ROOT set to v1 and all tests marked as v1 as a demonstration of this feature. These tests to not properly exercise the new API and are expected to fail.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,36 @@
+# This Software (Dioptra) is being made available as a public service by the
+# National Institute of Standards and Technology (NIST), an Agency of the United
+# States Department of Commerce. This software was developed in part by employees of
+# NIST and in part by NIST contractors. Copyright in portions of this software that
+# were developed by NIST contractors has been licensed or assigned to NIST. Pursuant
+# to Title 17 United States Code Section 105, works of NIST employees are not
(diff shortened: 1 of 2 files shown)


jkglasbrenner:
@keithmanville The `conftest.py` implementation that adds the `--run-v1` option flag and defines `@pytest.mark.v1` all looks good to me. I approve of getting the `conftest.py` file merged into `dev` and leaving this open as an example until we get some actual tests merged.

keithmanville:
The `conftest.py` implementation is now merged into `dev` under PR #431.

This PR will be left open as an example of how to use the `@pytest.mark.v1` decorator in a test.

PR #434, opened 2024-04-09
feat: add schema and controller stubs for Authentication endpoint

keithmanville:
This feature adds the request and response marshmallow schemas for the `/auth` resource. It also includes stub functions for the controller layer that defines the `/auth` endpoints as well as register these routes in the v1 API namespace.


jkglasbrenner (approved):
Looks good to me, I see that this is, for now, just a straight copy of what we had for `/auth` in v0. I think that's a good starting point. This is ready for merging.

PR #435, opened 2024-04-09
feat: remove the MLFlowRuns Resource from Dioptra

keithmanville:
This feature removes the MLFlowRuns resource from Dioptra, which includes the marshmallow schemas and controller layer in the v1 REST API.

This resource is no longer needed due to changes in the Dioptra entity relationship diagram.


jkglasbrenner (approved):
Looks good to me, this can be merged.

PR #438, opened 2024-04-10
Add v1 users schema and controller stubs

pscemama-mitre:
@jkglasbrenner @keithmanville note that right now the controller / schema allows one to list all the users.


keithmanville:
replaced with https://github.com/usnistgov/dioptra/pull/451

Issue #413, opened 2024-03-04
Backend tool for parsing Plugin files and extracting task function signatures


jkglasbrenner:
Use the Python [ast module](https://docs.python.org/3.9/library/ast.html) to build a backend tool that parses Plugin files and extracts the task function signatures, including type hints if they exist. This information will be shown on the frontend when a user is registering plugin tasks, and ultimately will be used to construct the [tasks section](https://pages.nist.gov/dioptra/user-guide/task_engine_doc.html#tasks) of the Declarative Experiment Description. The type hints can be used to figure out if new parameter types can be suggested for registration, which will ultimately be used in the [types section](https://pages.nist.gov/dioptra/user-guide/task_engine_doc.html#types) of the Declarative Experiment Description.

Note: We want to use `ast` instead of `importlib` because we do not want to execute arbitrary code just to extract the function signature information.

## Basic requirements

The general idea is that, given a plugin task function signature that looks like this,

```python
@pyplugs.register
def init_classifier(
    model_architecture: str,
    optimizer: Optimizer,
    metrics: list[Metric | FunctionType],
    input_shape: tuple[int, int, int],
    n_classes: int,
    loss: str = "categorical_crossentropy",
) -> Sequential:
    ...
```

the tool will produce an output that looks like this, 

```json
{
  "name": "init_classifier",
  "suggested_types": [
    {"suggestion": "optimizer", "type_annotation": "Optimizer"},
    {"suggestion": "list_metric_functiontype", "type_annotation": "list[Metric | FunctionType]"},
    {"suggestion": "tuple_int_int_int", "type_annotation": "tuple[int, int, int]"}
  ],
  "inputs": [
    {"name": "model_architecture", "type": "string"},
    {"name": "optimizer", "type": "optimizer"},
    {"name": "metrics", "type": "list_metric_functiontype"},
    {"name": "input_shape", "type": "tuple_int_int_int"},
    {"name": "n_classes", "type": "integer"},
    {"name": "loss", "type": "string", "required": false}
  ],
  "outputs": [
    {"name": "output1", "type": "sequential"}
  ]
}
```

It is understood that the user will need to make modifications to the type information (content under `"suggested_types"` and the `"type"` fields under `"inputs"` and `"outputs"`) before registration can be completed. The purpose for giving type suggestions is to provide initial suggestions and context that will be displayed in the frontend to help the user in this process.

[...]

PR #443, opened 2024-04-18
WIP feat: statically analyze python code to find task plugin function signatures

chisholm:
Addresses #413 

Some caveats:

- Depending on how one normalizes annotations, one can derive the same type name from distinct annotations.  That means task plugins with different typed arguments can wind up with the same suggested type names.  For example, my current normalization coalesces contiguous underscores to a single underscore and strips leading/trailing underscores, but it is easy to create examples where that is a problem.  But without that, you can wind up with suggested types like \_list\_\_str\__, list\_str\_, etc.  Funny looking type names.
- The signature structure's "outputs" property maps to a list.  That suggests support for functions returning multiple values (which does align with how task definitions are).  But whether a function returns more than one value is really only in the developer's mind.  A given function can be used as if it returned one or more than one value.  It's more of a usage difference.  If we choose some criteria to distinguish, it will be a guess, and may be wrong.  E.g. maybe a return type annotation as "tuple" or "typing.Tuple".  Should we guess?  If so, what should be the criteria?  Currently, signatures only include one output.
- When processing import statements, relative imports are ignored.  When all you have is some python source, you don't know what the imports are relative to, so you can't make sense of them.

### DIFF
--- a/src/dioptra/restapi/v1/lib/__init__.py
+++ b/src/dioptra/restapi/v1/lib/__init__.py
@@ -1,0 +1,16 @@
+# This Software (Dioptra) is being made available as a public service by the
+# National Institute of Standards and Technology (NIST), an Agency of the United
+# States Department of Commerce. This software was developed in part by employees of
+# NIST and in part by NIST contractors. Copyright in portions of this software that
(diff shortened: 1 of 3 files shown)


keithmanville:
included in #719

PR #446, opened 2024-04-19
feat: add batch permissions changes to groups

keithmanville:
This feature adds an endpoint for updating all user permissions for a group to the v1 REST API. This is implemented via a PUT to `/groups/{id}/members`. It includes the necessary changes to the groups schema and controller layer.

This commit also includes some minor changes to the group's schema that do not change functionality. This includes moving an import nested inside of a class and a nested class definition to the top level.


jkglasbrenner (approved):
LGTM

PR #447, opened 2024-04-19
feat: add tests for v1 queues

alexb1200:
Work in progress branch for V1 Queues api endpoint. Needs a few changes once the search functionality is completed. Covers all of the current user stories. Creating a new branch because of some rebasing weirdness I keep running into.


jkglasbrenner:
@keithmanville I've pushed the code for the "given" fixtures and the code for a new assert statement that explicitly validates the contents of the POST response after creating a new queue.

pscemama-mitre:
Just a question (not sure if this is the right place to do so):

Learning about fixtures and was wondering where we define the `client` and `db` in the file in order to fully define the fixtures used in the "arrange" phase of testing. Are they grabbed from some global environment variables? If so, where is the code that does this? Maybe I'm thinking about this wrong. Thanks!

chisholm:
`conftest.py` is a special file used by pytest.  It can contain fixture implementations to share across test suites.  You can search the same directory as a test suite and in parent directories for conftest.py files to see what fixtures are available (higher level directories allow broader sharing).

In this case, there is a conftest.py file in the same directory, with:

https://github.com/usnistgov/dioptra/blob/7dc25ce3ede9901c1f2f5b3eb1b12024262abba9/tests/unit/restapi/conftest.py#L202-L204

The `db` fixture is there as well.

tests/unit/restapi/lib/actions.py
```diff
@@ -0,0 +1,93 @@
...
+    Returns:
+        The response from the API.
+    """
+
+    payload = {"name": name}
+
+    if description:
+        payload["description"] = description
+    if group_id:
+        payload["group_id"] = group_id
```
jkglasbrenner, line 87: `group_id` is always required when registering a queue (and any of the other entities that are a resource/resource snapshot), so I'd remove the `if` statement and the default `None`:

```suggestion
def register_queue(
    client: FlaskClient,
    name: str,
    group_id: int,
    description: str | None = None,
) -> TestResponse:
    """Register a queue using the API.

    Args:
        client: The Flask test client.
        name: The name to assign to the new queue.
        group_id: The group to create the new queue in.
        description: The description of the new queue.

    Returns:
        The response from the API.
    """

    payload = {"name": name, "group_id": group_id}

    if description:
        payload["description"] = description
```

tests/unit/restapi/v1/conftest.py
```diff
@@ -0,0 +1,77 @@
...
+def registered_users(client: FlaskClient, db: SQLAlchemy) -> dict[str, Any]:
+    user1_response = register_user(client, "user1", "user1@example.org")
+    user2_response = register_user(client, "user2", "user2@example.org")
+    user3_response = register_user(client, "user3", "user3@example.org")
+    return {
+        "user1": user1_response,
+        "user2": user2_response,
+        "user3": user3_response,
+    }
+
```
jkglasbrenner, line 49: I'd put this before `auth_account` to follow the convention used elsewhere that fixtures that depend on other fixtures come later in the file. I.E. `auth_account` depends on `registered_users`, so we should encounter `registered_users` first in the file.

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,516 @@
...
+
+import pytest
+from flask.testing import FlaskClient
+from flask_sqlalchemy import SQLAlchemy
+from werkzeug.test import TestResponse
+
+from dioptra.restapi.routes import V1_QUEUES_ROUTE, V1_ROOT
+
+from ..lib.actions import register_queue
+from ..lib.helpers import is_iso_format
```
jkglasbrenner, line 35: The number of actions and helpers might grow large as we build out the testing infrastructure for v1. I suggest we export the `actions` and `helpers` module in `tests/unit/restapi/lib/__init__.py` so that we can just import as follows:

```suggestion
from ..lib import actions, helpers
```

Then we can make calls like `actions.register_queue` and `helpers.is_iso_format`, which make the purpose of the functions even more clear (these are "actions", these are "helpers").

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,516 @@
...
+
+# -- Actions ---------------------------------------------------------------------------
+
+
+def rename_queue(
```
jkglasbrenner, line 40: We might want to consider renaming this to `modify_queue` since the function can do more than just rename things (it can change the description).

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,516 @@
...
+        queue_id: The id of the queue to rename.
+        new_name: The new name to assign to the queue.
+        new_description: The new description to assign to the queue.
+
+    Returns:
+        The response from the API.
+    """
+    payload = {"name": new_name}
+    if new_description:
+        payload["description"] = new_description
```
jkglasbrenner, line 59: Since this is a PUT, then a value for description must always be provided in the payload.

If `new_description` is None, then you'll need to issue a GET to retrieve the current value for description so you can include it in the PUT. Otherwise, the description will disappear, which isn't the expected behavior.

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,516 @@
...
+    """
+
+    payload = {}
+
+    if Q:
+        payload.update(Q)
+    if G:
+        payload.update(G)
+    if P:
+        payload.update(P)
```
jkglasbrenner, line 193: Be careful about what's the payload and what's a query parameter, they go in different arguments in the get request. This is mixing up the JSON payload (which are embedded in the body of a HTTP Request) with the query parameters (which end up in the URL).

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,516 @@
...
+
+    response = client.get(
+        f"/{V1_ROOT}/{V1_QUEUES_ROUTE}", payload=payload, follow_redirects=True
```
jkglasbrenner, line 196: I don't think there's a `payload` argument in the `.get` method, it's usually either `json=` (when passing a JSON serializable dict) or `data=` (which I think allows you to pass anything, not just dicts).

In the `requests` library, there is a `params` argument for the query parameters: https://requests.readthedocs.io/en/latest/user/quickstart/#passing-parameters-in-urls. I suspect the Flask Test Client has a similar field.

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,516 @@
...
+        queue_id: The id of the queue to rename.
+        name: The name of an existing queue.
+    Raises:
+        AssertionError: If the response status code is not 400.
+    """
+    response = rename_queue(
+        client=client,
+        queue_id=queue_id,
+        new_name=existing_name,
+    )
```
jkglasbrenner, line 296: Needs a `group_id`

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,516 @@
...
+    - The user is able to retrieve a list of all registered queues with a description
+      that contains 'queue'.
+    - The returned list of queues matches the expected matches from the query.
+    """
+    queue1_expected = registered_queues["queue1"].get_json()
+    queue2_expected = registered_queues["queue2"].get_json()
+    queue_expected_list = [queue1_expected, queue2_expected]
+
+    # TODO the query field must be updated when the grammar for searches is updated.
+    search_parameters = {"query": "description:*queue*"}
```
jkglasbrenner, line 384: I recommend updating the assert functions so we don't have to remember to create a dict with the `"query"` key for the `Q=` argument. We should be able to just pass the string `"description:*queue*"` since there can only be one instance of a query parameter per request.

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,516 @@
...
+    queue1_expected = registered_queues["queue1"].get_json()
+    queue2_expected = registered_queues["queue2"].get_json()
+    queue3_expected = registered_queues["queue3"].get_json()
+    queue_expected_list = [queue1_expected, queue2_expected, queue3_expected]
+
+    search_parameters = {"group_id": auth_account["default_group_id"]}
+
+    assert_retrieving_all_queues_works(
+        client, expected=queue_expected_list, G=search_parameters
+    )
```
jkglasbrenner, line 416: The `G` argument is shorthand for a `group_id` query, right? If so, we shouldn't need to provide the `group_id` key in a dict, the assert function can build it for us.

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,516 @@
...
+    - The request fails with an appropriate error message and response code.
+    """
+    updated_queue_name = "tensorflow_gpu"
+    queue_to_rename = registered_queues["queue1"].json()
+    existing_queue = registered_queues["queue2"].json()
+
+    rename_queue(
+        client, queue_id=queue_to_rename["queueId"], new_name=updated_queue_name
+    )
+
```
jkglasbrenner, line 487: Minor nitpick, the whitespace between these functions can be removed. (comment applies to other test functions as well).

jkglasbrenner (changes requested):

tests/unit/restapi/v1/test_queue.py
keithmanville: looks like you can pass a dict as `query_string`: https://flask.palletsprojects.com/en/2.3.x/testing/

tests/unit/restapi/v1/test_queue.py
keithmanville: This is a rename/modify, not a register. Why do we need `group_id`?

tests/unit/restapi/lib/actions.py
```diff
@@ -0,0 +1,88 @@
...
+        "email_address": response["email"],
+        "default_group_id": response["groups"][0]["id"],
+    }
+
+
+def register_queue(
+    client: FlaskClient,
+    name: str,
+    group_id: int,
+    description: str = "",
```
jkglasbrenner, line 68: Minor nitpick: The description field should be optional in the API since it will be nullable in the database. So if you do this,

```suggestion
    description: str | None = None,
```

Then you can use an if statement to optionally add the description field to the payload.

The rationale for this is so we have one way to signify that a description is empty/blank, otherwise we'll need to do compound checks, i.e. "a description is considered empty if the description field is null or its a zero-length string or its a string that only contains whitespace."

tests/unit/restapi/v1/conftest.py
```diff
@@ -0,0 +1,79 @@
...
+
+@pytest.fixture
+def auth_account(
+    client: FlaskClient,
+    db: SQLAlchemy,
+    registered_users: dict[str, Any],  # noqa: F811
+) -> dict[str, Any]:
+    user_info = registered_users["user1"].json()
```
jkglasbrenner, line 46: I believe you can drop this `.json()`

```suggestion
    user_info = registered_users["user1"]
```

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,500 @@
...
+#
+# ACCESS THE FULL CC BY 4.0 LICENSE HERE:
+# https://creativecommons.org/licenses/by/4.0/legalcode
+"""Test suite for queue operations.
+
+This module contains a set of tests that validate the CRUD operations and additional
+functionalities for the queue entity. The tests ensure that the queues can be
+registered, renamed, deleted, and locked/unlocked as expected through the REST API.
+"""
+from __future__ import annotations
```
jkglasbrenner, line 23: Since we've migrated to Python 3.11, we don't need to include this anymore.

```suggestion
```

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,500 @@
...
+    """
+
+    query_string = {}
+    if group_id is not None:
+        query_string["groupId"] = group_id
+    if search is not None:
+        query_string["query"] = search
+    if paging_info is not None:
+        query_string["index"] = paging_info["index"]
+        query_string["pageLength"] = paging_info["page_length"]
```
jkglasbrenner, line 189: Minor whitespace nitpick, I prefer to separate `if` blocks with a blank line (I think it's more readable), take or leave:

```suggestion
    query_string = {}

    if group_id is not None:
        query_string["groupId"] = group_id

    if search is not None:
        query_string["query"] = search

    if paging_info is not None:
        query_string["index"] = paging_info["index"]
        query_string["pageLength"] = paging_info["page_length"]
```

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,500 @@
...
+    existing_description: str,
+) -> None:
+    """Assert that renaming a queue with an existing name fails.
+    Args:
+        client: The Flask test client.
+        queue_id: The id of the queue to rename.
+        name: The name of an existing queue.
+    Raises:
+        AssertionError: If the response status code is not 400.
+    """
```
jkglasbrenner, line 269: Whitespace nitpick for the docstring, although I think the Google-style docstring style requires separating the sections with a newline:

```suggestion
    """Assert that renaming a queue with an existing name fails.

    Args:
        client: The Flask test client.
        queue_id: The id of the queue to rename.
        name: The name of an existing queue.

    Raises:
        AssertionError: If the response status code is not 400.
    """
```

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,500 @@
...
+        follow_redirects=True,
+    )
+
+
+# -- Assertions ------------------------------------------------------------------------
+
+
+def assert_queue_response_contents_matches_expectations(
+    response: dict[str, Any], expected_contents: dict[str, Any]
+) -> None:
```
jkglasbrenner, line 90: Missing a docstring, only mentioning this since we've been consistent with adding docstrings to the rest of the actions and assertions.

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,500 @@
...
+    - A user registers three queues, "tensorflow_cpu", "tensorflow_gpu", "pytorch_cpu".
+    - The user is able to retrieve a list of all registered queues.
+    - The returned list of queues matches the full list of registered queues.
+    """
+    queue1_expected = registered_queues["queue1"]
+    queue2_expected = registered_queues["queue2"]
+    queue3_expected = registered_queues["queue3"]
+    queue_expected_list = [queue1_expected, queue2_expected, queue3_expected]
+
+    assert_retrieving_queues_works(client, expected=queue_expected_list)
```
jkglasbrenner, line 339: Double check that the behavior in `assert_retrieving_queues_works` is consistent with the objects in `queue_expected_list`. To me, it looks like `registered_queues` contains `TestResponse` objects (no `.get_json()` is issued in that fixture), but the assert function looks like its treating these objects like dictionaries.

tests/unit/restapi/v1/test_queue.py
```diff
@@ -0,0 +1,500 @@
...
+def test_queue_get_all(
+    client: FlaskClient,
+    db: SQLAlchemy,
+    auth_account: dict[str, Any],
+    registered_queues: dict[str, Any],
+) -> None:
+    """Test that all queues can be retrieved.
+
+    Given an authenticated user and registered queues, this test validates the following
+        sequence of actions:
```
jkglasbrenner, line 328: Minor whitespace nitpick, the prose paragraphs at the top of a Google-style docstring don't need to be indented on subsequent lines:

```suggestion
    Given an authenticated user and registered queues, this test validates the following
    sequence of actions:
```

This comment also applies to other docstrings in the file.

jkglasbrenner (changes requested):
I finished my new review of this code and have another set of requested changes. The majority of these are minor nitpicks, although one of them asks to double check that the "expected" lists of registered queue objects are being handled consistently in the assert statements (it looks like they're treated as dicts, but might still be `TestResponse` objects).

Also, are there Gerkhin-style user stories written for these tests? If so, we can add them to the `test_` functions.

tests/unit/restapi/v1/test_queue.py
keithmanville: I am making all of the actions return `TestResponse` (this is consistent with actions in `test_queue.py`) and fixtures return a dictionary. This required moving the logic for adding additional information to the reponse in `actions.register_user` to the `registered_users` fixture.

This allows the create/register tests validate the response codes when calling actions directly, but other tests that utilize the fixtures (where the creation of the resource is a given and not being tested) direct access to the underlying contents.

jkglasbrenner (approved):
LGTM, I think we're good to go for this now!

PR #448, opened 2024-04-19
feat: add a module for defining and parsing the v1 API query language

keithmanville:
This feature defines Dioptra's REST API query language and provides a parser. It is responsible for validating search query syntax and parsing searches into a list of search terms. It does not perform any further interpretation of search terms or construct queries for the Dioptra database.

### DIFF
--- a/src/dioptra/restapi/errors.py
+++ b/src/dioptra/restapi/errors.py
@@ -31,6 +31,10 @@ class SearchNotImplementedError(Exception):
     """The search functionality has not been implemented."""
 
 
+class SearchParseError(Exception):
+    """The search query could not be parsed."""
+
+
 def register_base_v1_error_handlers(api: Api) -> None:
     @api.errorhandler(BackendDatabaseError)
     def handle_backend_database_error(error):
@@ -40,8 +44,16 @@ def register_base_v1_error_handlers(api: Api) -> None:
         }, 500
 
     @api.errorhandler(SearchNotImplementedError)
-    def handle_no_current_user_error(error):
+    def handle_search_not_implemented_error(error):
         return {"message": "The search functionality has not been implemented"}, 501
+
+    @api.errorhandler(SearchParseError)
+    def handle_search_parse_error(error):
+        return {
+            "message": "The provided search query could not be parsed",
+            "query": error.args[0],
+            "reason": error.args[1],
+        }, 422
 
 
 def register_error_handlers(api: Api, restapi_version: str) -> None:
--- a/src/dioptra/restapi/v1/groups/controller.py
+++ b/src/dioptra/restapi/v1/groups/controller.py
@@ -17,6 +17,7 @@
 """The module defining the endpoints for Group resources."""
 import uuid
 from typing import cast
+from urllib.parse import unquote
 
 import structlog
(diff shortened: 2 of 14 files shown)


keithmanville:
@jkglasbrenner requesting feedback on this. You can play around with it and run `search_parser.py`. Once we are happy with how this works I can write some unit tests.

jkglasbrenner:
@keithmanville Found that we need to update these lines in the `build_paging_url` function:

https://github.com/usnistgov/dioptra/blob/6cc156f06c6752d0bad1a354022264b70cba8b9d/src/dioptra/restapi/v1/utils.py#L292-L293

Based on the Swagger docs, it should be changed to this:

```python
 if search: 
     query_params["search"] = search
```

jkglasbrenner:
@keithmanville I think I found an edge case in the search parser when you misuse double quotes. Below are 2 queues I've registered:

```json
[
  {
    "id": 2,
    "snapshot": 5,
    "group": {
      "id": 1,
      "name": "public",
      "url": "/groups/1"
    },
    "user": {
      "id": 1,
      "username": "jglasbrenner",
      "url": "/users/1"
    },
    "createdOn": "2024-05-30T16:13:05.766912+00:00",
    "lastModifiedOn": "2024-05-30T16:13:05.766912+00:00",
    "latestSnapshot": true,
    "tags": [],
    "name": "Tensorflow CPU",
    "description": "Allows users to run lighter jobs that leverage the Tensorflow library."
  },
  {
    "id": 3,
    "snapshot": 7,
    "group": {
      "id": 1,
      "name": "public",
      "url": "/groups/1"
    },
    "user": {
      "id": 1,
      "username": "jglasbrenner",
      "url": "/users/1"
    },
    "createdOn": "2024-05-30T16:14:15.194490+00:00",
    "lastModifiedOn": "2024-05-30T16:14:15.194490+00:00",
    "latestSnapshot": true,
    "tags": [],
    "name": "Tensorflow GPU",
    "description": "Run a tensorflow job that requires a GPU."
  }
]
```

I then run this search:

    /api/v1/queues/?search=description%3A%22*%22%22lighter%20jobs*%22%2Cname%3A%3Fensorflow*&groupId=1

Unquoted, that search corresponds to `description:"*""lighter jobs*",name:?ensorflow*`. My thought is that it should either return nothing, or maybe return the result for `description:"*lighter jobs*",name:?ensorflow*`. Instead it returns the full list of registered queues.

jkglasbrenner (commented):
I added a couple of comments for fixing, the first is about switching "query" to "search" in the build_paging_url function. The second is about some unexpected behavior when you misuse quotes or neglect to quote something and try and search for spaces.

keithmanville:
Those queries are invalid and should fail to parse. I needed to pass `parse_all=True` to `parse_string` to make sure pyparsing raises an exception if the entire string cannot be parsed. The partial parsing was causing the the unexpected behavior.

jkglasbrenner (approved):
I tried my best to break the search box after the latest fixes and wasn't successful, which is good :) It looks like the implementation is good enough to merge into dev and roll out into the rest of the service layer as its built out.

LGTM

PR #452, opened 2024-04-22
[WIP] feat: proof-of-concept for re-using controller layer Endpoints

keithmanville:
This is a proof-of-concept for re-using Endpoints from the controller layer to achieve common functionality across multiple root Endpoints with minimal code re-use.

The use case illustrated in this proof-of-concept is providing tagging functionality via a `/<resource>/{id}/tags` endpoint for all Dioptra Resources in the API. An Endpoint is created in `tags/controller.py` which defines the operations for tagging resource, but it is not given a route. In the other controller layers that wish to use this tagging functionality, a subclass of the Endpoint is created and the `route` decorator is applied.

Some customization can still be achieved via the constructor. For example, we can set a `resource_name` member variable can be set to customize logging messages. The correct Service object for the given resource can be provided here. We might want some kind of base ResourceService class that can handle common service layer operations like tagging.

### DIFF
--- a/src/dioptra/restapi/v1/entrypoints/controller.py
+++ b/src/dioptra/restapi/v1/entrypoints/controller.py
@@ -27,6 +27,7 @@ from flask_restx import Namespace, Resource
 from structlog.stdlib import BoundLogger
 
 from dioptra.restapi.v1.schemas import IdStatusResponseSchema
+from dioptra.restapi.v1.tags.controller import ResourceTagEndpoint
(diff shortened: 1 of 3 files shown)


keithmanville:
@jkglasbrenner putting this up to spark discussion on how we can hopefully re-use portions of the controller layer (and maybe service layer) for functionality that is common across Dioptra resources like tagging and versioning.

PR #459, opened 2024-05-09
feat: updated v1 API for changes to Experiments and Jobs

keithmanville:
This features modifies how jobs are handled in the REST API. Since Jobs can only be created as part of an experiment, the POST has been moved to /experiments/{id}/jobs. The GETs and DELETE associated with a specific Job have been replicated under /experiments/{id}/jobs because Job operations will often be performed within the context of an Experiment.

### DIFF
--- a/src/dioptra/restapi/v1/experiments/controller.py
+++ b/src/dioptra/restapi/v1/experiments/controller.py
@@ -26,7 +26,7 @@ from flask_login import login_required
 from flask_restx import Namespace, Resource
 from structlog.stdlib import BoundLogger
 
-from dioptra.restapi.v1.jobs.schema import JobSchema
+from dioptra.restapi.v1.jobs.schema import JobSchema, JobStatusSchema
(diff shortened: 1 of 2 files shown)


src/dioptra/restapi/v1/experiments/controller.py
```diff
@@ -122,16 +122,70 @@ def put(self, id: int):
...
+        log = LOGGER.new(
+            request_id=str(uuid.uuid4()), resource="Job", request_type="POST"
+        )
+        log.debug("Request received")
+        parsed_obj = request.parsed_obj  # noqa: F841
+
+
+@api.route("/<int:id>/jobs/<int:job_id>")
+@api.param("id", "ID for the Experiment resource.")
+@api.param("job_id", "ID for the Job resource.")
```
jkglasbrenner, line 153: Found some snakecase instead of camel case, I'll take care of it

jkglasbrenner (approved):
LGTM

PR #467, opened 2024-05-15
feat: add a logging endpoint, cookie HTTP logging handler

chisholm:
This PR updates the prototype branch to add a protected endpoint for receiving logging data, and a logging handler design which is able to send requests to that endpoint.  See src/proto/utils.py.

The handler has a general design which can send cookies with requests.  That could include session cookies, but is not a requirement.  With respect to the flask prototype though, it must include an authenticated session cookie since the endpoint is protected.  If "strong" session protection is in use, one must also be careful to always use the same User-Agent header value, otherwise flask-login will reject the cookie.  For that reason, the handler accepts a user_agent argument.

With respect to prototype authentication, it is up to the user of the logging handler to establish a login session.  Any cookie(s) from that must be passed to the handler when it is set up (and possibly a correct user agent value too).  The handler itself is agnostic to where the cookies come from.

The implementation uses only standard library APIs.  That makes it simple to incorporate into other codebases (which is not the plan here, but still a nice thing to think about when designing software components).

### DIFF
--- a/src/proto/app.py
+++ b/src/proto/app.py
@@ -3,6 +3,7 @@
 This module contains the application factory function and the application's
 configuration parameters.
 """
+
 from __future__ import annotations
 
 import os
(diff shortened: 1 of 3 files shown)


jkglasbrenner:
#856 has been closed and so this PR and branch are no longer needed as a reference. Closing this PR and deleting the [flask-prototype-logging-endpoint](https://github.com/usnistgov/dioptra/tree/flask-prototype-logging-endpoint) branch.

PR #469, opened 2024-05-16
V1 experiments tests

etrapnell-nist:
Experiments v1 unit tests using placeholder entrypoints. Jobs tests not included.

### DIFF
--- a/src/dioptra/restapi/errors.py
+++ b/src/dioptra/restapi/errors.py
@@ -91,6 +91,7 @@ def register_v1_error_handlers(api: Api) -> None:
     from dioptra.restapi import v1
 
     register_base_v1_error_handlers(api)
+    v1.experiments.errors.register_error_handlers(api)
     v1.groups.errors.register_error_handlers(api)
     v1.plugins.errors.register_error_handlers(api)
     v1.queues.errors.register_error_handlers(api)
--- a/src/dioptra/restapi/v1/experiments/controller.py
+++ b/src/dioptra/restapi/v1/experiments/controller.py
@@ -15,17 +15,19 @@
 # ACCESS THE FULL CC BY 4.0 LICENSE HERE:
 # https://creativecommons.org/licenses/by/4.0/legalcode
 """The module defining the endpoints for Experiment resources."""
-from __future__ import annotations
-
 import uuid
+from typing import cast
 
 import structlog
 from flask import request
 from flask_accepts import accepts, responds
 from flask_login import login_required
 from flask_restx import Namespace, Resource
+from injector import inject
 from structlog.stdlib import BoundLogger
 
+from dioptra.restapi.db import models
+from dioptra.restapi.v1 import utils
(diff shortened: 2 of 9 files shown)


tests/unit/restapi/v1/test_experiments.py
```diff
@@ -0,0 +1,859 @@
...
+        new_name: The new name to assign to the experiment.
+        new_entrypoints: The new entrypoints to associate with the experiment.
+
+    Returns:
+        The response from the API.
+    """
+
+    payload = {}
+
+    response = client.get(
```
keithmanville, line 62: The actions should only have one call to the API. Since we already have access to the contents of the registered experiments in the tests via fixtures, any logic for choosing which fields to change can be placed there.

See https://github.com/usnistgov/dioptra/blob/db41860d97a22547d6decdf10232b6a2e2007617/tests/unit/restapi/v1/test_queue.py#L38 for example modify action.

tests/unit/restapi/v1/test_experiments.py
```diff
@@ -0,0 +1,859 @@
...
+        AssertionError: If the response status code is not 200 or if the API response
+            does not match the expected response.
+    """
+    response = client.get(
+        f"/{V1_ROOT}/{V1_EXPERIMENTS_ROUTE}/{experiment_id}", follow_redirects=True
+    )
+    assert response.status_code == 200 and response.get_json() == expected
+
+
+def assert_retrieving_experiment_entrypoints_works(
```
keithmanville, line 207: same as `assert_retrieving_experiment_by_id` above?

tests/unit/restapi/v1/test_experiments.py
```diff
@@ -0,0 +1,859 @@
...
+        client: The Flask test client.
+        experiment_id: The id of the experiment to add entrypoints.
+        expected: The expected response from the API.
+
+    Raises:
+        AssertionError: If the response status code is not 200 or if the API response
+            does not match the expected response.
+    """
+    payload = {"entrypoints": entrypoints}
+    response = client.post(
```
keithmanville, line 250: adding entrypoints is a PUT and the payload needs to be the complete experiment object.

keithmanville (changes requested):
Partial review. Will add additional comments later.

tests/unit/restapi/v1/test_experiments.py
```diff
@@ -0,0 +1,859 @@
...
+        client: The Flask test client.
+        experiment_id: The id of the experiment to replace entrypoints.
+        entrypoints: The new entrypoints for the experiment.
+        expected: The expected response from the API.
+
+    Raises:
+        AssertionError: If the response status code is not 200 or if the API response
+            does not match the expected response.
+    """
+    current_experiment = client.get(
```
keithmanville, line 321: We should pass one of the experiments from the fixture from the test function instead of issuing a GET.

PR #473, opened 2024-05-21
[WIP] Proposed changes for V1 restapi helpers

andrewhand:
Move validation for expected content to helpers file

### DIFF
--- a/tests/unit/restapi/lib/helpers.py
+++ b/tests/unit/restapi/lib/helpers.py
@@ -21,6 +21,7 @@ API endpoints.
 """
 
 import datetime
+from typing import Any
 
 
 def is_iso_format(date_string: str) -> bool:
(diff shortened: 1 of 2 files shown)


tests/unit/restapi/lib/helpers.py
```diff
@@ -30,3 +31,86 @@ def is_iso_format(date_string: str) -> bool:
...
+
+def validate_group_ref(
+    group_ref: dict[str, Any], expected_contents: dict[str, Any]
+) -> bool:
+    try:
+        assert isinstance(group_ref["id"], int)
+        assert isinstance(group_ref["group"]["name"], str)
+        assert isinstance(group_ref["group"]["url"], str)
+        assert group_ref["id"] == expected_contents["group"]["id"]
+        return True
```
keithmanville, line 90: `group_ref` doesn't have a `"group"` key. Should just be `group_ref["name"]` etc.

tests/unit/restapi/lib/helpers.py
```diff
@@ -30,3 +31,86 @@ def is_iso_format(date_string: str) -> bool:
...
+    
+    try:
+        assert isinstance(response["id"], int)
+        assert isinstance(response["snapshotId"], int)
+        assert isinstance(response["createdOn"], str)
+        assert isinstance(response["lastModifiedOn"], str)
+        assert isinstance(response["latestSnapshot"], bool)
+    except ValueError("Response keys do not match expected keys types"):
+        return False
+
```
keithmanville, line 62: black caught some extra whitespace characters here and a couple other places.

keithmanville (changes requested):
Does the try/except ValueError with message pattern provide the most informative message in pytest? i.e. is it adding something over just having the sequence of asserts?

I tested this on my in progress queues branch and it works (after making the suggested changes with group_ref.

I like breaking the repeated chunks into helpers.

I suggest we get users/groups merged into dev, thenre base this and integrate the new pattern into test_user and test_group before merging this PR.

PR #477, opened 2024-05-30
feat: v1 drafts for queues endpoint

keithmanville:


src/dioptra/restapi/v1/shared/drafts/service.py
```diff
@@ -0,0 +1,433 @@
...
+        if total_num_drafts == 0:
+            return [], total_num_drafts
+
+        stmt = select(models.DraftResource).filter(  # type: ignore
+            models.DraftResource.resource_type == self._resource_type,
+            models.DraftResource.user_id == current_user.user_id,
+        )
+        drafts = db.session.scalars(stmt).all()
+
+        return drafts, total_num_drafts
```
jkglasbrenner, line 106: From what I can tell, the drafts are returned in a paging envelope, but this method isn't using the page index or page length to offset or limit the query.

Also, it doesn't appear that `error_if_not_found` argument is being used either.

jkglasbrenner:
I merged the search functionality to dev, so I switched the merge target for this branch to `dev` to keep the "stacked diff" moving.

I handled the rebase on the updated dev, had to resolve a couple of merge conflicts.

@keithmanville I'll take a look at this in more depth later on, but I just wanted to let you know about the rebase so that you can pull and reset locally if you need to make other changes.

PR #479, opened 2024-06-04
feat: entrypoints service layer

keithmanville:
This feature adds an implementation of the service layer for v1 of the REST API for the Entrypoints
endpoint. It includes the Tags, ResourceDrafts, and ResourceSnapshots (versions) subendpoints. as
well as search functionality. This commit also includes unit tests that cover common user
interactions with Entrypoints.

The primary implementations are in `restapi/entrypoints/service.py`. The services are injected into
controller layer in `restapi/entrypoints/controller.py` which calls the appropriate service layer
functionality. The controller layer uses helper functions from `restapi/utils.py` to translate ORM
objects back to the response types expected by the schema. Error handling is added to v1 routes by
registering error handlers in `restapi/errors.py`.


jkglasbrenner (approved):
It took a while, but I think this is finally in a place where we can merge to dev!

LGTM

PR #482, opened 2024-06-06
feat: implemented versions for the v1 Queues endpoint

keithmanville:
This feature implements versions for the Queues endpoint. It includes a generic versions service implementation as well as the controller layer for queues.

Tests for accessing versions of queues have been added to the queues test suite.


jkglasbrenner:
Noticed on this branch that the Drafts sub-endpoint is showing in the Swagger Docs, I don't see it in the v1-drafts branch, so maybe it's not an issue and will be fixed with a rebase?

image

PR #483, opened 2024-06-06
examples: create pytorch example using MAITE and HEART

jtsextonMITRE:
This example demonstrates compatibility with [MAITE ](https://github.com/mit-ll-ai-technology/maite/) and [HEART](https://github.com/IBM/heart-library/), adding task plugins for interacting with both. 
- Through MAITE, it demonstrates the ability to quickly load in public datasets and models, save both of these to the system, and load them into the typical MLFlow workflow for use in adversarial examples
- It also demonstrates the ability to use a single task plugin for multiple types of evasion examples by leveraging HEART
- Lastly, it demonstrates the ability to quickly test that example against multiple public models


jtsextonMITRE:
do not merge this yet, notebook for the example requires a bit more work to account for changes to the entry points

PR #493, opened 2024-06-10
Dc prerequisite checks

JustKuzya:
Working pre-check. Probably will require some addition refactoring and doc-stringing, but currently works


chisholm:
I got:

```
Traceback (most recent call last):
  File "/home/chisholm/dioptra/./cookiecutter-templates/cookiecutter-dioptra-deployment/hooks/pre_prompt.py", line 466, in <module>
    Prompts.check_docker_version(tool)
  File "/home/chisholm/dioptra/./cookiecutter-templates/cookiecutter-dioptra-deployment/hooks/pre_prompt.py", line 260, in check_docker_version
    (is_version_OK, need_ver, have_ver) = is_docker_OK()
                                          ^^^^^^^^^^^^^^
  File "/home/chisholm/dioptra/./cookiecutter-templates/cookiecutter-dioptra-deployment/hooks/pre_prompt.py", line 154, in is_docker_OK
    return (MinVersions.INFO_DOCKER_ENGINE['ver'] <= server_engine_version,
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '<=' not supported between instances of 'tuple' and 'NoneType'
```

It seems `server_engine_version` can still be `None` when it checks the version.  You may have assumed that if an `except` clause is run and it returns from a function, that the `finally` clause will not run.  I don't think it works that way.  Once inside the `try` clause, I think there is no way to skip an associated `finally` clause.  That's why it exists: it provides a place to put cleanup or other code which runs no matter what.

I see a more awkward problem with it though: on my VM, docker compose is invoked as `docker compose` (no hyphen).  There's a cookiecutter template variable I have to override for that.  So on my VM, that check will always fail.  I checked cookiecutter docs, and it looks like the pre_prompt hook is not subject to template expansion, so there may be no nice way to make it work on my VM. :(

chisholm:
Btw, the above crash is caused by an exception: the exception is caught but the finally clause executes anyway.  The original exception is of type `yaml.scanner.ScannerError`:

```
mapping values are not allowed here
  in "<unicode string>", line 2, column 9:
     Version:           26.1.4
```

my `docker version` output is:

```
Client: Docker Engine - Community
 Version:           26.1.4
 API version:       1.45
 Go version:        go1.21.11
 Git commit:        5650f9b
 Built:             Wed Jun  5 11:29:19 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          26.1.4
  API version:      1.45 (minimum version 1.24)
  Go version:       go1.21.11
  Git commit:       de5c9cf
  Built:            Wed Jun  5 11:29:19 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.33
  GitCommit:        d2d58213f83a351ca8f528a95fbd145f5654e957
 runc:
  Version:          1.1.12
  GitCommit:        v1.1.12-0-g51d5e94
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

That's not valid YAML (the Client/Server header parts mess it up).

keithmanville:
I tried running like this: `cruft create https://github.com/usnistgov/dioptra --checkout dc-prerequisite-checks --directory cookiecutter-templates/cookiecutter-dioptra-deployment` but it did not trigger the pre_prompt script. Am I doing something wrong?

PR #494, opened 2024-06-11
feat: updated schemas to indicate required and optional fields

keithmanville:
This feature updates the marshmallow schemas for the v1 REST API to indicate required fields and provide default values for optional fields.

Forced type conversions were removed from the controller layers to avoid incorrect conversions for optional fields.


jkglasbrenner (approved):
Removed the `load_default=list()` since it doesn't appear to be necessary when using `fields.Nested` (it's already a "list of dicts" type).

LGTM

PR #496, opened 2024-06-13
V1 plugins file types service layer

pscemama-mitre:


pscemama-mitre:
@keithmanville @jkglasbrenner two things.

1) See [this note](https://github.com/usnistgov/dioptra/blob/820c61a4cdcac091b55ee126fbfdb8752e9de761/src/dioptra/restapi/v1/plugins/service.py#L870-L874) about modifying a plugin file


2) There is a linting error arising from [here](https://github.com/usnistgov/dioptra/blob/820c61a4cdcac091b55ee126fbfdb8752e9de761/src/dioptra/restapi/v1/plugins/controller.py#L232-L240) saying argument 7 is an integer and not a boolean.

jkglasbrenner:
TODO: In controller.py for plugins endpoint, switch the `file_id` parameter to `fileId`.

pscemama-mitre:
Ready for review @jkglasbrenner @keithmanville

jkglasbrenner (approved):
LGTM

PR #507, opened 2024-06-26
feat: serialize draft payload with resource schema

keithmanville:
This feature serializes the contents of the draft payload using the resources schema before returning to the user. This ensures consistent field names in the payload for the user.


jkglasbrenner (approved):
LGTM

PR #510, opened 2024-06-27
[WIP] feat: test for constructing task engine yaml files

keithmanville:


tests/unit/restapi/test_db_models.py
```diff
@@ -907,3 +911,57 @@ def test_db_invalid_resource_dependency_fails(
...
+        if param.parameter_type.name not in BUILTIN_TYPES
+    }
+
+    parameters = {param.name: param.default_value for param in entrypoint.parameters}
+
+    tasks = {
+        task.plugin_task_name: {
+            "plugin": ".".join(
+                [plugin.name, Path(file.filename).stem, task.plugin_task_name]
+            ),
```
jkglasbrenner, line 941: I think this section will require a bit more logic, since it doesn't allow for subfolders. Say we have `utils/raw/sample.py` as the filename (I'm realizing now that may not be permitted in the Marshmallow schema, so I'll need to adjust it if that's not allowed). We could do something like the following to construct the plugin section:

```python
".".join(
    [
        plugin.name,
        *[Path(x).stem] for x in Path("utils/raw/sample.py").parts],
        task.plugin_task_name,
    ]
)
```

Although I don't think we should rely on `Pathlib` here, since that means this may not work as expected if you run the Flask server on Windows. This is just a rough prototype of what we can support.

tests/unit/restapi/test_db_models.py
```diff
@@ -907,3 +911,57 @@ def test_db_invalid_resource_dependency_fails(
...
+            "outputs": [
+                {param.name: param.parameter_type.name}
+                for param in task.output_parameters
+            ],
+        }
+        for file in files
+        for task in file.tasks
+    }
+
+    graph = yaml.safe_load(entrypoint.task_graph)["graph"]
```
jkglasbrenner, line 955: I know in the earlier test I included an explicit `graph:` section in the entrypoint, although I feel like we can omit that to save a level of indentation.

tests/unit/restapi/test_db_models.py
```diff
@@ -907,3 +911,57 @@ def test_db_invalid_resource_dependency_fails(
...
+
+    parameters = {param.name: param.default_value for param in entrypoint.parameters}
+
+    tasks = {
+        task.plugin_task_name: {
+            "plugin": ".".join(
+                [plugin.name, Path(file.filename).stem, task.plugin_task_name]
+            ),
+            "inputs": [
+                {"name": param.name, "type": param.parameter_type.name}
```
jkglasbrenner, line 943: Since the `required` field is now available, this can be updated to:

```python
                {"name": param.name, "type": param.parameter_type.name, "required": param.required}
```

tests/unit/restapi/test_db_models.py
jkglasbrenner: We also want to be able to handle the case where users put their plugin tasks in an `__init__.py` file.

tests/unit/restapi/test_db_models.py
keithmanville: agreed. Since the user will only have to provide the graph section, it doesn't make sense for them to nest everything. I just added that since it was in the example registered entrypoint.

tests/unit/restapi/test_db_models.py
keithmanville: Got it. Yeah, I wasn't considering how nested folders were implemented. From your comment, it sounds like this is handled by putting the full path in the filename.

PR #512, opened 2024-06-28
feat: add search, drafts, snapshots, and tags to Experiments

keithmanville:
feat: add search, tags, drafts, and snapshots to Experiments

This feature adds the following funcionality to Experiments:
- search by name, description, or tag
- ability to manage (retriev, apply, and remove) tags
- ability to create drafts
- the ability to view the history of resource


jkglasbrenner:
The functionality for the endpoint seems to be correct from what I can tell. I've noticed an issue with drafts where the `"group"` field is showing up in some unexpected places, but it's unrelated to the edits in this branch.

PR #513, opened 2024-06-28
Auto-register and protect built-in plugin task parameter types

jkglasbrenner:
This update makes it so that the 6 built-in plugin task parameter types, "any", "number", "integer", "string", "boolean", and "null", are automatically registered and protected with a readonly lock immediately after the default public group is created (currently, this happens when the first user account is created). The plugin parameter types endpoint has been updated to prevent users from creating new plugin parameter types with case-insensitive names that match the built-in types. The endpoint has also been updated to check for the presence of the readonly lock to prevent modification in PUT requests.

In addition, an alembic migration script has been added that:

1. Scans the database for existing parameter types that match the built-ins and sets a read-only lock on them.
2. Checks if the public group exists and, if so, registers any missing built-in parameter types.


jkglasbrenner:
The changes made in this branch are causing several of the integration tests related to plugin parameter types to fail. This isn't unexpected, since now registering a user automatically registers six parameter types by default. The existing tests need to be fixed to cope with this change.

PR #515, opened 2024-07-01
V1 schema fixes

keithmanville:
This PR includes the following fixes:

    fix: set custom schema names to avoid name clashes in swagger

    This commit sets custom schema model names in generated subendpoint
    Resource classes. flask accepts autogenerates these names by default,
    but name clashes can occur when variations of a schema are used across
    different endpoints (e.g. when using exclude to remove a field), which
    was causing incorrect documentation in swagger.

    fix: exclude groupId from draft schema where it should be inferred

    fix: make group required in base resource load schema


jkglasbrenner (approved):
LGTM

PR #517, opened 2024-07-03
Vue frontend

henrychoy:
This is the Vue frontend.  Below are some of the known issues that are still being looked at.

1. Drafts and tags are only implemented for queues so far
3. Some fields may not be showing a "this is required" error.  Need to add this error to the code editor as well.
4. When creating an experiment with an entrypoint, it is possible to create a job under that experiment with a different entrypoint, resulting in an error.  The entrypoint selection when creating the job should be limited to those that are available on the experiment.
5. Saving the experiment form when leaving and having it restored when coming back is not implemented yet.


keithmanville:
merged rebased branch into dev

PR #519, opened 2024-07-03
V1 fix experiment service

andrewhand:
Fixed issues with Experiment Services where Entrypoints were not being appended correctly to Experiments with a GET All.


jkglasbrenner (approved):
LGTM

PR #520, opened 2024-07-03
feat: add MlflowRun endpoint to Jobs

keithmanville:
This commit adds the MlflowRun endpoint to Jobs. It contains functionality to get and set the MlflowRun ID for a job. The endpoint is nested under `/experiments/{id}/jobs/{jobId}/mlflowRrun`. The id can be set via a POST and retrieved via a GET. Once set, the id cannot be changed and the API will return a 400 BAD REQUEST error. Setting the mlflow run id does not create a new snapshot or change the last modified on timestamp.


jkglasbrenner (approved):
LGTM

PR #528, opened 2024-07-11
V1 data access - Repository Pattern

chisholm:
Work on building repositories and transitioning services to use them

### DIFF
--- a/src/dioptra/restapi/db/repository/experiments.py
+++ b/src/dioptra/restapi/db/repository/experiments.py
@@ -1,0 +1,365 @@
+# This Software (Dioptra) is being made available as a public service by the
+# National Institute of Standards and Technology (NIST), an Agency of the United
+# States Department of Commerce. This software was developed in part by employees of
+# NIST and in part by NIST contractors. Copyright in portions of this software that
+# were developed by NIST contractors has been licensed or assigned to NIST. Pursuant
+# to Title 17 United States Code Section 105, works of NIST employees are not
+# subject to copyright protection in the United States. However, NIST may hold
+# international copyright in software created by its employees and domestic
+# copyright (or licensing rights) in portions of software that were assigned or
+# licensed to NIST. To the extent that NIST holds copyright in this software, it is
+# being made available under the Creative Commons Attribution 4.0 International
+# license (CC BY 4.0). The disclaimers of the CC BY 4.0 license apply to all parts
+# of the software developed or licensed by NIST.
+#
+# ACCESS THE FULL CC BY 4.0 LICENSE HERE:
+# https://creativecommons.org/licenses/by/4.0/legalcode
+"""
+The experiments repository: data operations related to experiments
+"""
+
+from collections.abc import Iterable, Sequence
+from typing import Any, Final, overload
+
+import dioptra.restapi.db.repository.utils as utils
+from dioptra.restapi.db.models import (
+    EntryPoint,
+    Experiment,
+    Group,
+    Tag,
+)
+
+
+class ExperimentRepository:
+    SEARCHABLE_FIELDS: Final[dict[str, Any]] = {
+        "name": lambda x: Experiment.name.like(x, escape="/"),
+        "description": lambda x: Experiment.description.like(x, escape="/"),
+        "tag": lambda x: Experiment.tags.any(Tag.name.like(x, escape="/")),
+    }
+
+    # Maps a general sort criterion name to an Experiment attribute name
+    SORTABLE_FIELDS: Final[dict[str, str]] = {
+        "name": "name",
+        "createdOn": "created_on",
+        "lastModifiedOn": "last_modified_on",
+        "description": "description",
+    }
+
+    def __init__(self, session: utils.CompatibleSession[utils.S]):
+        self.session = session
+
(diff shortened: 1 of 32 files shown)


src/dioptra/restapi/db/unit_of_work.py
```diff
@@ -0,0 +1,63 @@
...
+class UnitOfWork(contextlib.AbstractContextManager):
+    """
+    A class whose instances act as a focal point for data operations.
+    Instances can commit/rollback transactions; they are also usable as
+    context managers which manage transactions automatically.  One can access
+    repositories via attributes on the instance.
+    """
+
+    def __init__(self) -> None:
+        self.session = db.session
```
jkglasbrenner, line 38: Is there an argument against passing the session to the constructor like so:

```suggestion
    def __init__(self, session) -> None:
        self.session = session
```

We could then set up a `UnitOfWork` provider in https://github.com/usnistgov/dioptra/blob/31d2087cd41362297b777a9d5e0e14c4c7b9ae60/src/dioptra/restapi/bootstrap.py that passes in `db.session`.

jkglasbrenner (commented):
@chisholm I think this is looking good overall, and does a lot to clean up and organize the database access patterns in the service layer. I left a couple of minor comments/questions, but nothing major.

Since you're wrapping up work on the Plugin Parameter Types repository, I'll hold off on processing this for a merge until that's in place and I get take a look at it.

src/dioptra/restapi/v1/groups/service.py
```diff
@@ -305,63 +242,26 @@ def delete(self, group_id: int, **kwargs) -> dict[str, Any]:
...
-        stmt = select(models.Group).filter_by(group_id=group_id, is_deleted=False)
-        group = db.session.scalars(stmt).first()
+        group = self._uow.group_repo.get(group_id, DeletionPolicy.NOT_DELETED)
 
         if group is None:
             raise EntityDoesNotExistError(GROUP_TYPE, group_id=group_id)
 
-        name = group.name
+        with self._uow:
+            self._uow.group_repo.delete(group)
```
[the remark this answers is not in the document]
chisholm: I like that style's simplicity and cleanliness, but it means you will always get a commit (or rollback in case of exception).  Unfortunately, many of the service methods have a `commit` arg which needs to be honored.  Always-commit doesn't work correctly in that case (if commit is false, you mustn't commit).  So most of the time, the code is more complicated.

So, the criteria for that style for me is: whenever you can get away with it!

src/dioptra/restapi/db/unit_of_work.py
chisholm: No good argument.  If `db` is global (and we're willing/able to import it), and it's suitable for use as-is in the contexts we need, it seemed like there was no need to inject it (or its session).  So I just imported it and used it directly.

jkglasbrenner:
@chisholm I wanted to verify that rebasing this branch on the current state of dev wouldn't break anything, as I want to see this branch merged into dev next. When I did a rebase locally, it brought in a couple of recent additions to the Plugin Parameter Types services that happened since the last time you rebased on dev. I wanted to try and start getting a handle on how you would resolve this using the repository patterns you've established, so I tried refactoring them to use the pattern myself.

I didn't want to overwrite your branch in place without your awareness, so I pushed the rebase + refactors to a separate branch: [v1-data-access-rebased-migrated](https://github.com/usnistgov/dioptra/tree/v1-data-access-rebased-migrated). Here's a link to my commit so you can get a sense of the way I handled it: https://github.com/usnistgov/dioptra/commit/cf5bbd7f7a648807226ce7d1203e9e7d861ce45b.

I have not written any extra tests to cover the new `get_snapshots_by_names` function I added to the repo utils file (you'll see why I needed to add something like it based on the upstream changes pulled from dev). So right now that new function is only indirectly being tested via unit tests that cover the `PluginParameterTypeNameService.get` and `BuiltinPluginParameterTypeService.get` methods.

I'm happy to proceed however works best in your view. If the updates I proposed are close enough to how you'd solve them, we can force push [v1-data-access-rebased-migrated](https://github.com/usnistgov/dioptra/tree/v1-data-access-rebased-migrated) into this branch too, and then add any additional tests to get proper coverage of the new repo util function. If you think there's a better/simpler way to handle the new additions, maybe by generalizing the `get_snapshot_by_name` function to work with one or more names, then I can leave it to you to implement.

jkglasbrenner:
Based on discussion, we will want our repos to include `get` and `get_one` methods. The former returns the object or None, the latter only returns an object otherwise it will raise an EntityDoesNotExistError.

src/dioptra/restapi/db/repository/experiments.py
```diff
@@ -0,0 +1,366 @@
...
+        """
+
+        return utils.get_latest_child_snapshots(
+            self.session,
+            EntryPoint,
+            experiment,
+            deletion_policy,
+        )
+
+    def set_children(
```
jkglasbrenner: Later repositories will have more than one available resource dependency to set, and they can be manipulated independently. To support this pattern, we can be explicit about the name of each child in the repository set method, e.g.:

```suggestion
    def set_entrypoint(
```

PR #532, opened 2024-07-12
Models and artifacts frontend

henrychoy:
This PR contains the following changes:

- Add Models and Artifacts pages
- Add field required warning to code editor if it is required
- Fixes tags
- Add ability to edit/remove plugin tasks after they are saved
- Remove api version from vite config proxy and store it as a contestant variable that all endpoints use
- Various bug fixes


keithmanville:
@henrychoy I'm getting the same bug we had on the plugin-params page for artifacts, where once you navigate to it clicking any of the other tabs fails and you have to refresh the page.

keithmanville:
For the model versions table that is visible when expanding, can we add some additional information like the description of the version and maybe the artifact URI? This is available in the response from `/api/v1/models/{id}/versions`. And can we make it so selecting a version opens up a new panel with more complete information?

henrychoy:
> For the model versions table that is visible when expanding, can we add some additional information like the description of the version and maybe the artifact URI? This is available in the response from `/api/v1/models/{id}/versions`. And can we make it so selecting a version opens up a new panel with more complete information?

I just added the additional version info to the expanded model row.  I was able to fit all of the data in there, but let me know if you still want the popup panel.

Also regarding the crashing artifacts page, I made a small edit that I think may be related.  Can you pull the latest and send me the error in the console if it's still happening?  Can you also send your db file?  I can't seem to replicate it on my end.

henrychoy:
FYI for anyone testing this branch, there was an issue with the Models and Artifacts page crashing due to the nested table the expanded rows getting undefined data.  This should be fixed with the latest push.  Let me know if anyone still has the issue with the site crashing.

jkglasbrenner (approved):
LGTM, I also pushed a change to revert the workaround I had in the NGINX configuration files.

PR #535, opened 2024-07-15
fix: queue deletion now removes association with entrypoints

### LINKED ISSUE #531: [BUG] Web UI entrypoints error after queue deletion
**Describe the bug**
When using the web UI, deleting a queue that is in use by an entry point causes a failure in the entrypoints endpoint. This is a fatal error for the frontend since the deleted queue cannot be restored and entry points can no longer be retrieved while it is in this state. This same issue is persistent on the create jobs page under an experiment when trying to add an entry point from the drop down menu.

Both of these scenarios cause an error to be displayed, which is included below in "Additional context."

To recover from this error, the dioptra-dev.db had to be deleted, followed by running dioptra-db autoupgrade to start again from a clean state.

**To Reproduce**
1. Login to a user account
2. Navigate to the Queue tab and create a queue
3. Navigate to the Entry-Points tab and select that queue when creating an Entry Point
4. Navigate back to the Queues tab and delete the queue
5. Navigate to the Entry-Points tab and the error below is displayed, and the entry point is no longer listed

**Expected behavior**
The queue should no longer show up in the entry point after the queue is deleted. All created entry points should still be retrievable. 

 - OS: macOS Sonoma 14.5
 - Browser: Firefox
 - Version: 127.0.2

**Additional context**
This is the error that is displayed from the issue described above when navigating to the Entry-Points tab and when clicking the entry point drop down menu when creating a job:
Not Found - The requested queue does not exist. You have requested this URI [/api/v1/entrypoints/] but did you mean /api/v1/entrypoints/<int:id>/tags/<int:tagId> or /api/v1/entrypoints/<int:id>/tags or /api/v1/entrypoints/<int:id>/draft ?

keithmanville:
This commit fixes a bug where retrieving an entrypoint that was associated with a queue that has since been deleted resulted in a 404 error. The issue was that queue deletion did not remove the resource dependency association between the queue and any entrypoints.

This fix resolves the issue by:
- Updating the queue service to remove all parents of the queue on deletion
- Adding deletion checks when retrieving queues associated with an entrypoint
- Updating the queue deletion test assert an associated entrypoint can be retrieved and the deleted queue is no longer listed.

Resolves #531

### DIFF
--- a/src/dioptra/restapi/v1/entrypoints/service.py
+++ b/src/dioptra/restapi/v1/entrypoints/service.py
@@ -249,7 +249,7 @@ class EntrypointService(object):
             resource.resource_id
             for entrypoint in entrypoints
             for resource in entrypoint.children
-            if resource.resource_type == "queue"
+            if resource.resource_type == "queue" and not resource.is_deleted
         )
         queues = {
             queue.resource_id: queue
@@ -266,7 +266,7 @@ class EntrypointService(object):
         }
         for entrypoint in entrypoint_dicts.values():
             for resource in entrypoint["entry_point"].children:
-                if resource.resource_type == "queue":
+                if resource.resource_type == "queue" and not resource.is_deleted:
                     entrypoint["queues"].append(queues[resource.resource_id])
 
         drafts_stmt = select(
@@ -349,7 +349,7 @@ class EntrypointIdService(object):
         queue_ids = set(
             resource.resource_id
             for resource in entrypoint.children
-            if resource.resource_type == "queue"
+            if resource.resource_type == "queue" and not resource.is_deleted
         )
         queues = self._queue_ids_service.get(list(queue_ids), error_if_not_found=True)
 
@@ -633,15 +633,11 @@ class EntrypointIdPluginsService(object):
                 for plugin in new_entrypoint.entry_point_plugin_files
             }.values()
         )
-
-        queue_ids = set(
-            resource.resource_id
+        queue_resources = [
+            resource
             for resource in entrypoint.children
             if resource.resource_type == "queue"
-        )
-        queues = self._queue_ids_service.get(list(queue_ids), error_if_not_found=True)
-        queue_resources = [queue.resource for queue in queues]
-
+        ]
         new_entrypoint.children = plugin_resources + queue_resources
 
         db.session.add(new_entrypoint)
@@ -756,6 +752,7 @@ class EntrypointIdPluginsIdService(object):
         if plugin_id not in plugin_ids:
             raise EntrypointPluginDoesNotExistError
 
+        # create a new snapshot with the plugin removed
         new_entrypoint = models.EntryPoint(
             name=entrypoint.name,
             description=entrypoint.description,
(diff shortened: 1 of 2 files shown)


keithmanville:
After discussion, we decided that the correct behavior should be to keep the resource dependency association when a queue is deleted, and fix the issue by checking for a delete lock before retrieving the queue. This would allow us to maintain the relationship if we choose to offer an "un-delete" feature in the future.

I pushed a commit that makes this change.

keithmanville:
@jkglasbrenner should be ready for your review. One of the tests is failing due to a mypy error in task_engine that I don't get locally. Perhaps a mypy version bump in the runner?

Let me know if this approach looks good and I can check for this issue with the other resource dependency relationships.

chisholm:
I suspect the mypy error is due to a more recent version of the jsonschema library.  Version 4.23.0 was released just in the last week apparently.  We had been using a separate typing library (types-jsonschema); with the new release, more typing information seems to have migrated into the library itself.  So the typing situation has changed.  We may just need to revisit usage of this library, and bring it up to date.

https://github.com/python-jsonschema/jsonschema/releases/tag/v4.23.0

jkglasbrenner:
> I suspect the mypy error is due to a more recent version of the jsonschema library. Version 4.23.0 was released just in the last week apparently. We had been using a separate typing library (types-jsonschema); with the new release, more typing information seems to have migrated into the library itself. So the typing situation has changed. We may just need to revisit usage of this library, and bring it up to date.
> 
> https://github.com/python-jsonschema/jsonschema/releases/tag/v4.23.0

Looks like we can satisfy this new mypy error by removing types-jsonschema (since the typing info has been integrated into the main library) and adding an explicit `cast()` to this line,

https://github.com/usnistgov/dioptra/blob/5071049240ee3049a9602ec5668ec36a7a148d8d/src/dioptra/task_engine/error_message.py#L262

updating it to this:

```python
    return cast(bool, validator.is_valid(sub_instance))
```

For reasons unknown, mypy cannot figure out what `validator.is_valid` is supposed to return (the type inference engine for VSCode, on the other hand, seems to be able to figure it out). Since we know its always going to be a boolean, let's use `cast` to help the type checker along.

I'll push a fix to dev and then rebase this branch.

jkglasbrenner:
> Hmmm, the types-jsonschema library still claims to "provide accurate annotations for jsonschema==4.23.*". But I also did notice new type annotations in the jsonschema library itself. So I was thinking perhaps some but not all of the relevant type information had migrated into the jsonschema library...? And that perhaps the separate library was still relevant for some things? But I had not really looked into it.
> 
> https://pypi.org/project/types-jsonschema/

We can always add `types-jsonschema` back later if we find that we're missing important static type checking information without it. For now, the above change + removing types-jsonschema satisfied the type checker.

jkglasbrenner (approved):
LGTM

chisholm:
For posterity, in case funny errors in this area of the code come up again in the future:

The actual error from the github runner was:

```
src/dioptra/task_engine/error_message.py:283:47: error: Argument 1 to "_get_one_of_alternative_names" has incompatible type "Any | Unset"; expected "Iterable[Any]"  [arg-type]
src/dioptra/task_engine/error_message.py:287:48: error: Argument 2 to "zip" has incompatible type "Any | Unset"; expected "Iterable[Any]"  [arg-type]
src/dioptra/task_engine/error_message.py:328:47: error: Argument 1 to "_get_one_of_alternative_names" has incompatible type "Any | Unset"; expected "Iterable[Any]"  [arg-type]
Found 3 errors in 1 file (checked 293 source files)
```

`validator.is_valid()` is not actually mentioned here.  The is_valid() error is what I was addressing with the Validator protocol type annotation (it's explained in the code comments), and may have also needed the type stubs.  Maybe that popped up again after `types-jsonschema` had been uninstalled?

I think the real problem had to do with `ValidationError.validator_value`.  The value of that attribute is the piece of sub-schema which failed to validate.  But that could be anything (any valid piece of schema).  So it has a very abstract type (Any | Unset).  But the Python code which shows the errors activates under very specific schema validation conditions: what failed was a `oneOf`, and either too many alternatives validated, or none validated (there are two different functions for those two different scenarios).  In those cases, validator_value *must* be a list (or Iterable), since oneOf always maps to a list (of alternative schemas).  That's too much for mypy to understand, so the type check failed.  Maybe it's surprising it had worked before, but maybe the type annotations were less complete before, so the check passed.

A better fix would probably have been to just cast the validator_value values appropriately.

PR #555, opened 2024-07-23
feat: Truncate long text

henrychoy:
This MR truncates long text for names, tags, and descriptions to better fit everything on the table pages.  If truncated, you can mouse-over to view a tooltip of the full text.  The cutoff for names and tags is 18 characters.  The cutoff for descriptions is 40 characters.

### DIFF
--- a/src/frontend/src/components/CodeEditor.vue
+++ b/src/frontend/src/components/CodeEditor.vue
@@ -14,6 +14,7 @@
       :style="{ 'min-height': '250px', 'max-height': '70vh',
         'border': `${showError ? '2px solid red' : '2px solid black'}`
       }"
+      :disabled="readOnly"
     />
     <caption
       :class="{ invisible: showError?.length === 0 ? true : false }"
@@ -76,7 +77,7 @@
       oneDark,
       yamlLinter,
       lintGutter(),
-      EditorState.readOnly.of(props.readOnly)
+      EditorState.readOnly.of(props.readOnly) // not working, using disabled prop instead
     ]
   })
 </script>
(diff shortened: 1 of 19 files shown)


etrapnell-nist:
This looks good except for one thing I noticed. If a name, description, or tag has only underscores and no spaces, the full text can get truncated on mouse-over. This is mainly a problem for plugin names, which can't contain spaces.

henrychoy:
> This looks good except for one thing I noticed. If a name, description, or tag has only underscores and no spaces, the full text can get truncated on mouse-over. This is mainly a problem for plugin names, which can't contain spaces.

@etrapnell-nist Thanks, I just pushed a fix that wraps the text in the tooltip even if it has no spaces.  

@keithmanville I also noticed that the long text could still be an issue in many other places in the UI such as the page title, breadcrumbs, or buttons that have a dynamic label.  An easier way to tackle all those would be to enforce a character limit rule upon input, either on the frontend or backend.  If you want the frontend to handle it, let me know if you have any guidance on what the character limits for each type of resource should be.

etrapnell-nist:
The only thing I see now is that experiment names still have the underscore wrap problem. Maybe because they are links? Looks good otherwise.

henrychoy:
> The only thing I see now is that experiment names still have the underscore wrap problem. Maybe because they are links? Looks good otherwise.

Oops I missed that, yes the experiment name is handled separately since it's a link.  Just pushed up the fix now.

etrapnell-nist:
Thanks! Looks good now.

etrapnell-nist (approved):

keithmanville:
squashed and merged to dev

PR #578, opened 2024-07-31
Fix build-breaking typo in cookiecutter.json

cdstelly:
The tag for the `argbash` docker container set within cookiecutter.json is currently set to `sha-0d046a`, which doesn't exist and causes a build failure. This PR updates to `sha-0d046ad`, which is a valid tag.

Before fix: 
Screenshot 2024-07-31 161520

After fix: 
Screenshot 2024-07-31 161541

Maintainers: double check `matejak/argbash:sha-0d046ad` is the intended version from the [argbash Docker Hub repo](https://hub.docker.com/r/matejak/argbash/tags).


keithmanville:
Thank you for identifying the issue and for the fix!

I modified your commit message to pass gitlint and merged to main.

PR #583, opened 2024-08-06
fix: return correct createdOn timestamp for all resources

keithmanville:
The API was returning the snapshot's created_on timestamp instead of the resource's created_on timestamp. This commit fixes the issue in the response build functions in `restapi/v1/utils.py`.


jkglasbrenner (approved):
LGTM, I'll merge this to dev and also cherry-pick it into main.

PR #613, opened 2024-08-16
V1 fix entrypoint schema

andrewhand:
Allows for registration of entrypoints without including queues in request. Test coverage added.

### DIFF
--- a/src/dioptra/restapi/v1/entrypoints/schema.py
+++ b/src/dioptra/restapi/v1/entrypoints/schema.py
@@ -154,8 +154,10 @@ class EntrypointMutableFieldsSchema(Schema):
         fields.Integer(),
         attribute="queue_ids",
         data_key="queues",
+        allow_none=True,
         metadata=dict(description="The queue for the entrypoint."),
         load_only=True,
+        load_default=list(),
     )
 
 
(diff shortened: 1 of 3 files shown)


src/dioptra/restapi/v1/entrypoints/schema.py
```diff
@@ -151,8 +151,10 @@ class EntrypointMutableFieldsSchema(Schema):
         fields.Integer(),
         attribute="queue_ids",
         data_key="queues",
+        allow_none=True,
```
keithmanville, line 154: not necessary

src/dioptra/restapi/v1/entrypoints/schema.py
```diff
@@ -151,8 +151,10 @@ class EntrypointMutableFieldsSchema(Schema):
...
         metadata=dict(description="The queue for the entrypoint."),
         load_only=True,
+        load_default=list(),
```
keithmanville, line 157: Not quite right, should be just `list`.

Part of docs showing that a callable can be used in load_default/dump_default argument: https://marshmallow.readthedocs.io/en/stable/quickstart.html#specifying-defaults

keithmanville (changes requested):
tests need to be updated to incorporate the python client.

Add the similar fix for plugins and parameters.

keithmanville:
fixed in #783

PR #614, opened 2024-08-16
fix: incorrect plugin task parameter order

keithmanville:
This commit fixes an issue where plugin task parameters were returned sorted alphabetically by name instead of respecting the specified order. It modifies an existing test to check for this case.


jkglasbrenner (approved):
LGTM

PR #641, opened 2024-09-12
fix: fix issue in modifying plugins associated with an entrypoint

keithmanville:
This commit fixes an issue when appending or deleting a plugin from an entrypoint. The services were re-using EntrypointPluginFile and EntrypointParameter ORM objects when creating new Entrypoint snapshots. This led to consistency errors in the database. The service implementations were corrected to construct new ORM objects to attach to the new Entrypoint snapshot


jkglasbrenner (approved):
LGTM, I'm going to also propagate this change to main since its a straightforward fix.

PR #652, opened 2024-10-07
[WIP] V1 entrypoint workflow yaml validation

andrewhand:


andrewhand:
Need to resolve merge conflicts with new error handling methods before rebase.

src/dioptra/restapi/v1/workflows/errors.py
```diff
@@ -39,3 +43,10 @@ def handle_experiment_does_not_exist_error(error):
             "message": "Not Found - The experiment associated with the job does not "
             "exist"
         }, 404
+
+    @api.errorhandler(EntrypointWorkflowValidationIssue)
+    def handle_entrypoint_workflow_validation_error(error):
+        issues = error.args
+        print (issues)
+        message = "\n".join(str(issue) for issue in issues)
```
keithmanville, line 51: We should create a "detail" dictionary that extracts the relevant information from task engine issues.

See Harold's error handling refactor: https://github.com/usnistgov/dioptra/blob/a66cad4d524c733a936ec0afe6f6242d4a96c137/src/dioptra/restapi/errors.py#L341

This lets use return more structured information instead of just string-ifiying the issues directly.

src/dioptra/restapi/v1/workflows/service.py
```diff
@@ -64,3 +83,87 @@ def get(self, job_id: int, file_type: FileTypes, **kwargs) -> IO[bytes]:
...
+
+        Raises:
+            
+        """
+        log: BoundLogger = kwargs.get("log", LOGGER.new())
+        log.debug("Validate a entrypoint workflow", task_graph=task_graph, plugin_ids=plugin_ids, entrypoint_parameters=entrypoint_parameters)
+
+        parameters = {param['name']: param['default_value'] for param in entrypoint_parameters}
+
+        tasks: dict[str, Any] = {}
```
keithmanville, line 132: I think we should move this block (line 132 to 161) to `export_task_engine_yaml.py`.  It would be a function similar to `export_task_engine_yaml` but with different inputs.

tests/unit/restapi/v1/test_workflows.py
```diff
@@ -0,0 +1,234 @@
...
+
+    plugin_ids = [plugin_response["id"]]
+    entrypoint_parameters = [
+        {
+            "name" : "name",
+            "defaultValue": "User",
+            "parameterType": "string",
+        },
+    ]
+    expected_message = "[ValidationIssue(IssueType.SEMANTIC, IssueSeverity.ERROR, 'In step \"hello_step\": unrecognized task plugin: hello_wrld')]"
```
keithmanville, line 227: I don't think we should match the error text exactly. Maybe just check response code and the issue type.

keithmanville (changes requested):
Did an initial pass on this and left comments.

The additional feature to add to this PR is using the validation in the entrypoint create and modify services to prevent invalid entrypoints from being created.

andrewhand:
Moved to new PR https://github.com/usnistgov/dioptra/pull/673 after rebase issues.

PR #653, opened 2024-10-08
Save and reload form

henrychoy:
- When navigating away from an edit form, a popup should warn the user that their modifications will be lost (unless the user made no changes or hit cancel)

- When navigating away from a create form, the data should be preserved for the remainder of the current session, so when the user navigates back, they have the option of reloading or discarding their un-saved values.

- Job forms are temporarily saved in a experiment context, meaning that a saved job form for Experiment A have no effect on the job forms for experiment B.  Similarly, plugin files are saved in a plugin context.

- If a saved form contains resource selections (entrypoints, queues, plugins) that have since been deleted, those values will not be re-loaded into the form.


henrychoy:
> When navigating away from the create form, it warns me that I will lose my data. I hit confirm. When I navigate back it offers to fill in my previous inputs.
> 
> I think we should drop the warning when navigating away since my inputs are recoverable.
> 
> I was testing this for experiments

@keithmanville ok just pushed the change.  When leaving a create form, it will temp save and leave without warning the user.

PR #657, opened 2024-10-16
feat: add a depth-limited repr function

### LINKED ISSUE #656: [BUG] ORM object reprs are excessively large
**Describe the bug**
The repr of a single object can be enormous.

**To Reproduce**

You can try the following script:

```python
# This creates 10 queues each of which has 3 tags randomly chosen from a
# set of 10.
import random
import dioptra.restapi.db.models as m

user = m.User("name", "password", "user@example.org")
group = m.Group("groupname", user)

tags = [m.Tag(f"tag_{i}", group, user) for i in range(10)]

queues = []
for i in range(10):
    res = m.Resource("queue", group)
    queues.append(m.Queue("a queue", res, user, f"queue_{i}"))

for queue in queues:
    queue.tags.extend(random.sample(tags, k=3))

print(repr(queues[0]))
```

This will repr one of the queues so created.  You'll want to redirect the output to a file and wait a bit.  I've run this a couple times; both times, the resulting file was over **1GB** in size!

**Expected behavior**

A reasonably sized repr.

chisholm:
Fixes #656 .

This PR shows a way to reduce the size of reprs of ORM objects.  It implements a custom repr function which is depth-limited.  It's a simple generic way to "prune" the repr.  If you try again the script given in the issue, the result should *much* smaller.

This PR also overrides \_\_repr\_\_() for three ORM classes: Queue, Resource, Tag as an example.

The function is implemented in a pretty generic way, so you could use it with simple types and generic data structures as well, but the depth limiting is specifically implemented in terms of nested ORM objects, since it is written to solve the ORM object repr problem.

Update: I moved the \_\_repr\_\_() override to a base class and removed the overrides for Queue/Resource/Tag.  This should enable depth-limited reprs for all ORM classes automatically.

### DIFF
--- a/src/dioptra/restapi/db/db.py
+++ b/src/dioptra/restapi/db/db.py
@@ -34,6 +34,8 @@ from sqlalchemy.engine import Engine
 from sqlalchemy.ext.mutable import MutableDict
 from sqlalchemy.orm import DeclarativeBase, MappedAsDataclass, mapped_column
 
+from dioptra.restapi.db.models.utils import depth_limited_repr
+
 from .custom_types import GUID, TZDateTime
 
 intpk = Annotated[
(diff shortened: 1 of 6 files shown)


jkglasbrenner (approved):
LGTM

PR #658, opened 2024-10-16
WIP: V1 artifact files endpoint changes

mtrapnell-nist:


src/dioptra/restapi/v1/artifacts/controller.py
```diff
@@ -46,6 +47,7 @@
 from .service import (
     RESOURCE_TYPE,
     SEARCHABLE_FIELDS,
+    ArtifactIdContentsService,
```
keithmanville, line 50: causes an import error right now. I realize you haven't worked on the service layer yet, but I found it useful to create a class stub, so flask would run and I could view the endpoint and schema in swagger.

src/dioptra/restapi/v1/artifacts/schema.py
```diff
@@ -42,6 +42,33 @@ class ArtifactRefSchema(ArtifactRefBaseSchema):  # type: ignore
...
 
 
+class ArtifactFileSchema(Schema):
+    """The schema for the files stored in an Artifact resource."""
+
+    relativePath = fields.String(
+        attribute="relative_path",
+        metadata=dict(description="Relative path to the Artifact File."),
+    )
+    artifactType = fields.String(
```
keithmanville, line 52: will need to coordinate artifact types with @hbooth 's work. Probably should be an enum see here for an example:
https://github.com/usnistgov/dioptra/blob/a661aaa168788dffb1d7dcb1c95952b4106e9b7e/src/dioptra/restapi/v1/schemas.py#L211C1-L229C1

src/dioptra/restapi/v1/artifacts/schema.py
```diff
@@ -81,10 +128,40 @@ class ArtifactPageSchema(BasePageSchema):
...
+            description="Determines whether the file will be downloaded or viewed."
+        ),
+        load_default=None,
+    )
+
+
 class ArtifactGetQueryParameters(
     PagingQueryParametersSchema,
     GroupIdQueryParametersSchema,
     SearchQueryParametersSchema,
```
keithmanville, line 162: I think we should leave out Search and SortBy query parameters at least for now. These would require a custom implementation since search and sort for resources are implemented via the sqlalchemy.

src/dioptra/restapi/v1/artifacts/schema.py
```diff
@@ -69,6 +96,26 @@ class ArtifactSchema(ArtifactMutableFieldsSchema, ArtifactBaseSchema):  # type:
...
     )
+    isDir = fields.Boolean(
+        attribute="is_dir",
+        metadata=dict(
+            description="Indicates whether or not the Artifact is a directory."
+        ),
+        required=True,
+    )
+    files = fields.Nested(
+        ArtifactFileSchema,
```
keithmanville, line 107: should be the paged schema.

src/dioptra/restapi/v1/artifacts/schema.py
```diff
@@ -69,6 +96,26 @@ class ArtifactSchema(ArtifactMutableFieldsSchema, ArtifactBaseSchema):  # type:
...
+        attribute="files",
+        metadata=dict(description="The list of files contained in an Artifact."),
+        required=True,
+    )
+    fileTree = fields.Nested(
+        ArtifactFileSchema,
```
keithmanville, line 113: need a schema for the file tree view. I'm not sure if we defined this yet. I'm also not sure how it would work with the paging parameters.

The intent of the file tree was to provide the frontend with a nested file tree structure that makes it easy to display in a file browser like UI. If we are concerned with the amount of data that might be returned and need paging, maybe we can instead remove `fileTree` and have the UI use `files`, but add a `recurse` boolean query param that determines if we want a flat list of all files, or if we just want the files in the directory in the `path` query param. This way the UI would work by making a call to `/artifacts/{id}/contents?path=PATH&recurse=False`, updating the `path` each time the user clicks a file/directory in the browser.

src/dioptra/restapi/v1/artifacts/schema.py
```diff
@@ -81,10 +128,40 @@ class ArtifactPageSchema(BasePageSchema):
...
+    resource endpoint."""
+
+    showFileTree = fields.Boolean(
+        attribute="show_file_tree",
+        metadata=dict(description="Display files as a directory."),
+        load_default=None,
+    )
+
+
+class DownloadPathQueryParametersSchema(Schema):
```
keithmanville, line 142: Consider renaming to something like `ArtifactContentsQueryParameterSchema` (especially if we add additional fields like `recurse`) since we are grouping a couple different parameters all related to the contents or files associated with the artifact.

src/dioptra/restapi/v1/artifacts/schema.py
```diff
@@ -81,10 +128,40 @@ class ArtifactPageSchema(BasePageSchema):
...
+    """A schema for adding path and downlaod query parameters to a resource endpoint."""
```
keithmanville, line 143: minor typo: downlaod -> download

keithmanville (changes requested):

PR #663, opened 2024-10-24
Respect package-lock.json when building the frontend with init-deployment.sh

### LINKED ISSUE #662: Pin npm dependencies
Pin the npm dependencies by generating a lock file to prevent frontend builds from breaking unexpectedly.

jkglasbrenner:
This updates the init-deployment.sh script so that it copies the frontend's package-lock.json file into the build context. The command "npm install" is replaced with "npm ci", which will do a "clean install" of the exact versions in package-lock.json and will fail if anything breaks.

Closes #662


jkglasbrenner:
Several people on our team tested this out and confirmed it worked.

PR #672, opened 2024-10-30
feat(restapi): add metrics capabilities to RESTAPI

### LINKED ISSUE #671: Proxy MLFlow metrics requests through the RESTAPI
**Definition of Done**
* Creation of 3 new endpoints:
  * /jobs/{id}/metrics
    * should return one of each metric with the latest timestamp for this job
  * /jobs/{id}/metrics/{metric_name}/snapshots
    * should return all metrics for this job
  * /experiments/{id}/metrics
    * should return one of each metric for each job in the experiment
* Tests for the above endpoints

jtsextonMITRE:
Closes #671.

Additionally:
Fixes some incorrect function descriptions.
Fixes some mypy linting issues in related files.
Adds a restapi test for the /jobs/{id}/mlflowRun endpoint using a fixture added to help the metrics test.

### DIFF
--- a/src/dioptra/restapi/v1/jobs/schema.py
+++ b/src/dioptra/restapi/v1/jobs/schema.py
@@ -15,6 +15,8 @@
 # ACCESS THE FULL CC BY 4.0 LICENSE HERE:
 # https://creativecommons.org/licenses/by/4.0/legalcode
 """The schemas for serializing/deserializing Job resources."""
+import re
+
 from marshmallow import Schema, fields, validate
 
 from dioptra.restapi.v1.artifacts.schema import ArtifactRefSchema
@@ -27,6 +29,9 @@ from dioptra.restapi.v1.schemas import (
     generate_base_resource_ref_schema,
     generate_base_resource_schema,
 )
+
+ALLOWED_METRIC_NAME_REGEX = re.compile(r"^([A-Z]|[A-Z_][A-Z0-9_]+)$", flags=re.IGNORECASE)  # noqa: B950; fmt: skip
+
 
 JobRefSchema = generate_base_resource_ref_schema("Job")
 JobSnapshotRefSchema = generate_base_resource_ref_schema("Job", keep_snapshot_id=True)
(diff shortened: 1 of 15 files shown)


src/dioptra/restapi/v1/jobs/schema.py
```diff
@@ -41,14 +41,68 @@ class JobMlflowRunSchema(Schema):
...
+    name = fields.String(
+        attribute="name",
+        metadata=dict(description="The name of the metric."),
+        required=True,
+    )
+    value = fields.Float(
+        attribute="value",
+        metadata=dict(description="The value of the metric."),
+        required=True,
+    )
```
keithmanville, line 62: let's add `step` as an `Int` defaulting to `0`.
This mimics mlflow client's `log_metric` function and we can pass it through there in the service layer. 
https://mlflow.org/docs/latest/python_api/mlflow.client.html#mlflow.client.MlflowClient.log_metric

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -527,6 +525,157 @@ def get(
...
+        log.debug(
+            "Get job metric history by id and name",
+            job_id=job_id,
+            metric_name=metric_name,
+        )
+
+        run_id: UUID = self._job_id_mlflowrun_service.get(job_id=job_id, **kwargs)[
+            "mlflow_run_id"
+        ]
+        client = MlflowClient()
```
keithmanville, line 657: If it can't connect to the mlflow server it attempts to re-connect several times, so this request can hang for a bit. Not sure if this is the best behavior? Maybe we should set number of retries lower?

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -527,6 +525,157 @@ def get(
...
+
+        try:
+            history = client.get_metric_history(run_id=run_id.hex, key=metric_name)
+        except MlflowException as e:
+            raise EntityDoesNotExistError("MlFlowRun", run_id=run_id.hex) from e
+
+        metrics_page = [
```
keithmanville, line 664: recommend re-using the `build_paging_envelope` function in `utils` for this.

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -1055,3 +1204,68 @@ def create(
             )
 
         return {"mlflow_run_id": job.mlflow_run.mlflow_run_id}
+
+
+class ExperimentMetricsService(object):
```
keithmanville, line 1209: I think this should go under `experiments/service.py`

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -1055,3 +1204,68 @@ def create(
...
+            page_length=page_length,
+            sort_by_string=sort_by_string,
+            descending=descending,
+            **kwargs,
+        )
+
+        job_ids = [job["job"].resource_id for job in jobs]
+
+        metrics_for_jobs = [
+            {"id": job_id, "metrics": self._job_id_metrics_service.get(job_id)}
```
keithmanville, line 1268: If even one of these is missing an mlflowRunId, an error is raised. This can happen when a job errors out before attaching an mlflowRunId.

Recommend adding a check for this and returning an empty list of metrics for that job.

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -527,6 +525,157 @@ def get(
...
+
+        Returns:
+            The metrics for the requested job if found, otherwise an error message.
+        """
+        from mlflow.tracking import MlflowClient
+
+        log: BoundLogger = kwargs.get("log", LOGGER.new())
+        log.debug("Get job metrics by id", job_id=job_id)
+
+        run_id: UUID = self._job_id_mlflowrun_service.get(job_id=job_id, **kwargs)[
```
keithmanville, line 561: If the job is missing mlflowRunId, an error is raised. This can happen when a job errors out before attaching an mlflowRunId.

Recommend adding a check for this and returning an empty list of metrics instead of raising an error.

keithmanville (changes requested):

src/dioptra/restapi/v1/jobs/service.py
jtsextonMITRE: I put this here because it relies on both ExperimentJobService and JobIdMetricsService which causes a circular import if you put it in experiments/service.py - I tried a local import to no avail, same issue

src/dioptra/restapi/v1/jobs/schema.py
jtsextonMITRE: I added timestamp as well since that is an option in mlflow

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -527,6 +525,172 @@ def get(
...
+        )["mlflow_run_id"]
+        client = MlflowClient()
+
+        if run_id is None:
+            raise EntityDoesNotExistError("MlFlowRun", run_id=None)
+
+        try:
+            history = client.get_metric_history(run_id=run_id.hex, key=metric_name)
+        except MlflowException as e:
+            raise EntityDoesNotExistError("MlFlowRun", run_id=run_id.hex) from e
```
keithmanville, line 677: How do we want to handle errors here?

`get_metric_history` returns an empty list if the `run_id` is not found or the there are no metrics with `key`. It doesn't raise an error. So we don't actually get an error raised by mlflow. It does raise an error if `run_id` is None, but it's a `INVALID_PARAMETER_VALUE_ERROR`

Do we want to raise an error if the metric with the specified name is not found? When we retrieve all metrics and there are none, we simply return an empty list, but maybe requesting a specific name should operate differently. Should this be a new error type? I thought `EntityDoesNotExist` was more for top level entities (e.g. Resources, Users, Groups).

We still might want a try/except around any mlflow client invocations to catch any kind of connectivity errors. Should we have some new error type for errors with an internal dioptra service?

keithmanville: @jkglasbrenner @jtsextonMITRE

keithmanville:
@jtsextonMITRE I made a couple commits to this branch:
- I didn't have an issue with `import uuid` causing a circular import. Can you confirm this is working for you?
- I updated the Metrics schemas - hopefully these changes make sense.

src/dioptra/restapi/v1/jobs/service.py
keithmanville: @jkglasbrenner these questions about error handling are my remaining unresolved issues with this PR. Once we resolve this I would approve the PR.

keithmanville (approved):

PR #673, opened 2024-10-31
New restapi v1 entrypoint workflow yaml validation

andrewhand:
New branch for entrypoint workflow yaml validation


jkglasbrenner:
This branch is significantly out of sync with the dev branch despite what the Files changed tab implies. This ORM models still include the 3-way joining table that we separated a while ago. I think the issue occurred because a git pull was used instead of a git rebase.

I went ahead and imported as much of the current state as I could into a new branch: https://github.com/usnistgov/dioptra/tree/cleanup-restapi-entrypoint-workflow-yaml-validation. My subsequent work on this will continue there for now.

The only file I held off on importing were the changes to the entrypoint service. While I was reviewing, I realized that we need to include groupId in the endpoint as a parameter, otherwise you have no way to know what plugin parameter types you need to pull from the database (this is unfortunately necessary to do). That change will break that changes in the entrypoint service, so I'm going to delay those changes for now.

To try and reduce code duplication, my plan is to build dataclass adapters that follow the same form as the ORM classes. In the validation service, I'll migrate the input data into these adapters, and then we can reuse all of the tooling that's currently used in the job download workflow endpoint.

jkglasbrenner:
Superseded by #835

PR #674, opened 2024-11-01
Sync to latest plugin

henrychoy:
In the entrypoint table and edit form, plugins should be labeled as outdated if it isn't the latestSnapshot.  Beside each outdated plugin should be a "sync to latest plugin" button that should update the plugin to the latest snapshot.  This is available in the entrypoint edit form and the assign plugins dialog in the entrypoints table.


keithmanville:
resolved by #678

Issue #620, opened 2024-08-16
Rework job creation UI


keithmanville:
The Job creation interface should be updated to support creating jobs from the jobs page as well as the experiments page. The interface should have drop downs for Experiment, Entrypoint, and Queue. Entrypoint and Queue should be grayed out and only populated with acceptable values once the previous values have been selected.

### Definition of done:
- [ ] The Job creation interface has been updated
- [ ] This feature is merged into dev

PR #683, opened 2024-12-04
feat: new common job form

henrychoy:
Issue: https://github.com/usnistgov/dioptra/issues/620

- Jobs can now be created through the all-jobs page and the experiment -> job page.
- Entrypoint dropdown will be greyed out until experiment is selected, and will only show entrypoints that are linked to the selected experiment
- Queue dropdown will be greyed out until entrypoint is selected, and will only show queues that are linked to the selected entrypoint
- Returning to an unsaved form should give you the option to load previous values

### DIFF
--- a/src/frontend/src/dialogs/AppendResource.vue
+++ b/src/frontend/src/dialogs/AppendResource.vue
@@ -1,0 +1,125 @@
+<template>
+  <DialogComponent 
+    v-model="showDialog"
+    @emitSubmit="submit"
+    :hideDraftBtn="true"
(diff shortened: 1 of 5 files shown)


keithmanville (commented):
This properly implements the changes in the issue.

One thing I noticed is that if there is no associated entrypoint or queue, the drop down is correctly empty, but it might not be clear to the user why this is the case or what to do to correct it. We should consider including an inline message or maybe even a "quick add" dialog to make these associations.

PR #697, opened 2024-12-19
feat: browse queue history

henrychoy:
In the edit queue dialog, add view history toggle to browse snapshots.  Latest snapshot will be selected by default.  Click or tab over to the list, and then you can use up/down arrow keys.  Submit button will be disabled when viewing history.

Timestamps are currently all the same because snapshots inherit timestamp from the parent resource, this will be fixed separately.

### DIFF
--- a/src/frontend/src/dialogs/DialogComponent.vue
+++ b/src/frontend/src/dialogs/DialogComponent.vue
@@ -1,10 +1,18 @@
 <template>
   <q-dialog v-model="showDialog" aria-labelledby="modalTitle" :persistent="persistent">
-    <q-card style="width: 95%" flat >
+    <q-card flat style="min-width: 500px; max-width: 80%;">
       <q-form @submit="$emit('emitSubmit')">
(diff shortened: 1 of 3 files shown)


keithmanville (changes requested):
I believe the GET for snapshots is only retrieving a single page of results, so if you have more than 10 snapshots, you only get the 10 oldest ones in the list.

Is it possible to retrieve the first page ordered by date descending, and then if the user scrolls down, it retrieves additional pages as needed?

henrychoy:
> I believe the GET for snapshots is only retrieving a single page of results, so if you have more than 10 snapshots, you only get the 10 oldest ones in the list.
> 
> Is it possible to retrieve the first page ordered by date descending, and then if the user scrolls down, it retrieves additional pages as needed?

Thanks for catching this.  As discussed on the call, we're going to have the snapshot list GET ALL results for now, and in a future task I can look into adding a sort option and/or loading more on scroll.

PR #699, opened 2024-12-23
feat: add snapshot_created_on field to resource response schemas

keithmanville:
This commit adds the snapshot_created_on datetime field to all Resource response schemas. It also updates the `build_` helper functions to include the field in responses and updates the test suite to validate the updated response schemas.


henrychoy (approved):
Works as expected.  Snapshots now have `snapshotCreatedOn` which is timestamp when it was created.

PR #703, opened 2024-12-25
Db up downgrade snapshot

### LINKED ISSUE #474: Remove the "resource_snapshot" type from the resource_snapshot_types Enum table
This will prevent the creation of a "bare" `ResourceSnapshot` entry with type "resource_snapshot" in the database. All snapshots are supposed to be an inherited type, so we want to prevent this at the data integrity level.

A couple hints for this. First, to generate a blank migration script (is necessary here), use 

```sh
dioptra-db revision -m "<Migration Script Message Goes Here>"
```

That will create a new migration script file in `src/dioptra/restapi/db/alembic/versions/`.

A migration script that does something similar to what we need here is https://github.com/usnistgov/dioptra/blob/dev/src/dioptra/restapi/db/alembic/versions/10f9e72e72aa_add_readonly_resource_lock_type_to_db_.py. However, there are a couple of key differences:

1. The linked migration script is adding a record to the Enum table. This change requires deleting an existing value.
2. The linked migration script does not have a downgrade script. This change requires us to have both a working upgrade and downgrade script. Upgrade deletes the `"resource_snapshot"` record, downgrade inserts it back into the table.

The main information to take from the linked migration script is that it shows how you can set up the migration ORM models to handle CRUD operations on the table. Note that the migration ORM models are supposed to be standalone, we should not import the models under `src/dioptra/restapi/db/models` in a migration script.

This migration script has a little more complexity, and shows the need for both an upgrade and downgrade: https://github.com/usnistgov/dioptra/blob/dev/src/dioptra/restapi/db/alembic/versions/6a75ede23821_update_entry_point_parameter_types_list.py

### Definition of Done
- [x] The "resource_snapshot" enum entry is removed from the table, and a new migration script is created
- [ ] All existing tests pass
- [ ] The bug fix is merged into dev

JustKuzya:
The code changes fixes #474 and has up-/down-grade logic for alembic up-/down-grades

### DIFF
--- a/src/dioptra/restapi/db/alembic/versions/0ca6eca33569_upgrade_resource_snapshot_types.py
+++ b/src/dioptra/restapi/db/alembic/versions/0ca6eca33569_upgrade_resource_snapshot_types.py
@@ -1,0 +1,89 @@
+"""Up/Down-grade resource snapshot types as delete/insert 'resource_snapshot' [GH-Issue #474]
+
+Revision ID: 0ca6eca33569
+Revises: f1d231f7ef15
+Create Date: 2024-12-12 00:47:23.575103
+
+"""  # noqa: B950
+
+from typing import Annotated
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.orm import (
+    DeclarativeBase,
+    Mapped,
+    MappedAsDataclass,
+    mapped_column,
+    sessionmaker,
+)
+
+# revision identifiers, used by Alembic.
+revision = "0ca6eca33569"
+down_revision = "f1d231f7ef15"
+branch_labels = None
+depends_on = None
+
+
+text_ = Annotated[str, mapped_column(sa.Text())]
+
+
+# The data to up/down-grade
+ENUM_TABLE = "resource_types"
+TABLE_ENTRY = "resource_snapshot"
+
+
+class UpgradeBase(DeclarativeBase, MappedAsDataclass):
+    pass
+
+
+class DowngradeBase(DeclarativeBase, MappedAsDataclass):
+    pass
+
+
+class ResourceTypeUpgrade(UpgradeBase):
+    __tablename__ = ENUM_TABLE
+    resource_type: Mapped[text_] = mapped_column(primary_key=True)
+
+
+class ResourceTypeDowngrade(DowngradeBase):
+    __tablename__ = ENUM_TABLE
+    resource_type: Mapped[text_] = mapped_column(primary_key=True)
+
+
+def upgrade():
+    """Alembic Upgrade Hook-Point that deletes the entry from
+    resource_types table with value 'resource_snapshot'
+    """
+    bind = op.get_bind()
+    Session = sessionmaker(bind=bind)
+
+    with Session() as session:
+        stmt = sa.select(ResourceTypeUpgrade).where(
+            ResourceTypeUpgrade.resource_type == TABLE_ENTRY
+        )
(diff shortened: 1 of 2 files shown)


jkglasbrenner:
The migration script is mixing the ORM and Core styles of SQLAlchemy in a way that I don't think you need to do here, and its making it harder to understand what is happening. Also, in `downgrade()`, you shouldn't need that counting operation because the `resource_type` field is a primary key already. The way the migration scripts are applied (e.g. you cannot apply the same upgrade or downgrade twice in a row) should also prevent a duplicate row from being inserted even if the database engine ignores the uniqueness constraint.

jkglasbrenner:
Here's how you can write this migration script using just the ORM. Note that you need to create both Upgrade and Downgrade versions of your ORM classes in order for things to work correctly (based on the comments in your code, I suspect that you ran into this and that's why you started using operations from SQLAlchemy Core). Both the `upgrade()` and the `downgrade()` migrations query and check for the record where `resource_type = 'resource_snapshot'`, and use that as a guard against deleting or inserting the record when its unexpectedly missing/present.

```python
"""Up/Down-grade resource snapshot types as delete/insert 'resource_snapshot' [GH-Issue #474]

Revision ID: 0ca6eca33569
Revises: 6a75ede23821
Create Date: 2024-12-12 00:47:23.575103

"""

from typing import Annotated

import sqlalchemy as sa
from alembic import op
from sqlalchemy.orm import (
    DeclarativeBase,
    Mapped,
    MappedAsDataclass,
    mapped_column,
    sessionmaker,
)

# revision identifiers, used by Alembic.
revision = "0ca6eca33569"
down_revision = "6a75ede23821"
branch_labels = None
depends_on = None

text_ = Annotated[str, mapped_column(sa.Text())]


class UpgradeBase(DeclarativeBase, MappedAsDataclass):
    pass


class DowngradeBase(DeclarativeBase, MappedAsDataclass):
    pass


class ResourceTypeUpgrade(UpgradeBase):
    __tablename__ = "resource_types"
    resource_type: Mapped[text_] = mapped_column(primary_key=True)


class ResourceTypeDowngrade(DowngradeBase):
    __tablename__ = "resource_types"
    resource_type: Mapped[text_] = mapped_column(primary_key=True)


def upgrade():
    """Alembic Upgrade Hook-Point that deletes the entry from
    resource_types table with value 'resource_snapshot'
    """
    bind = op.get_bind()
    Session = sessionmaker(bind=bind)

    with Session() as session:
        stmt = sa.select(ResourceTypeUpgrade).where(
            ResourceTypeUpgrade.resource_type == "resource_snapshot"
        )
        resource_snapshot_type = session.scalar(stmt)

        if resource_snapshot_type is not None:
            session.delete(resource_snapshot_type)

        session.commit()


def downgrade():
    """Alembic Downgrade Hook-Point that reinstates the entry to
    resource_types table with value 'resource_snapshot'
    """
    bind = op.get_bind()
    Session = sessionmaker(bind=bind)

    with Session() as session:
        stmt = sa.select(ResourceTypeUpgrade).where(
            ResourceTypeUpgrade.resource_type == "resource_snapshot"
        )
        resource_snapshot_type = session.scalar(stmt)

        if resource_snapshot_type is None:
            session.add(ResourceTypeUpgrade(resource_type="resource_snapshot"))

        session.commit()

```

jkglasbrenner (changes requested):
See https://github.com/usnistgov/dioptra/pull/703#issuecomment-2605969070.

PR #706, opened 2024-12-30
feat: browse experiment history

henrychoy:
Add ability for user to browse Experiment history in the edit form.  Clicking the toggle will bring out the right drawer containing the snapshot list, which can eventually be used in other form pages.  Selecting a row loads the values into the form.  Up/down keyboard navigation should work when you are clicked into the table.  Table can be sorted.  The queue dialog was also updated to use this SnapshotList component.


keithmanville (approved):
This looks good to me.

I think the greyed out entrypoint section with the "Entrypoints are not part of Experiment snapshots" message is nicer than hiding that panel.

Adding @hbooth and @jkglasbrenner as reviewers as well since we haven't had a chance go over the design together.

PR #707, opened 2024-12-31
feat(restapi): add snapshot id to draft modifications PUT endpoints

keithmanville:
This commit adds an optional query parameter for changing a draft modifications base snapshot ID to the PUT endpoint. If a snapshot ID is provided the value is updated, otherwise it is left unchanged.

This feature is designed to enable the reconciliation of drafts by allowing the user to signal that their changes are based on a more recent snapshot.

An existing test is modified to test this feature.

### DIFF
--- a/src/dioptra/client/drafts.py
+++ b/src/dioptra/client/drafts.py
@@ -373,11 +373,14 @@ class ModifyResourceDraftsSubCollectionClient(SubCollectionClient[T]):
             json_=self._validate_fields(kwargs),
         )
 
-    def modify(self, *resource_ids: str | int, **kwargs) -> T:
+    def modify(
+        self, *resource_ids: str | int, resource_snapshot_id: str | int, **kwargs
+    ) -> T:
         """Modify a resource modification draft.
 
         Args:
             *resource_ids: The parent resource ids that own the sub-collection.
+            resource_snapshot_id: The resource snapshot id the draft is based on.
             **kwargs: The draft fields to modify.
 
         Returns:
@@ -389,7 +392,10 @@ class ModifyResourceDraftsSubCollectionClient(SubCollectionClient[T]):
         """
         return self._session.put(
             self.build_sub_collection_url(*resource_ids),
-            json_=self._validate_fields(kwargs),
+            json_={
+                "resourceSnapshot": int(resource_snapshot_id),
+                "resourceData": self._validate_fields(kwargs),
+            },
         )
 
     def delete(self, *resource_ids: str | int) -> T:
--- a/src/dioptra/restapi/v1/shared/drafts/service.py
+++ b/src/dioptra/restapi/v1/shared/drafts/service.py
@@ -32,6 +32,7 @@ from dioptra.restapi.errors import (
     DraftAlreadyExistsError,
     DraftDoesNotExistError,
     EntityDoesNotExistError,
+    InvalidDraftBaseResourceSnapshotError,
 )
 from dioptra.restapi.v1.groups.service import GroupIdService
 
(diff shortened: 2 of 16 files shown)


keithmanville:
I ended up implementing this feature with an optional query parameter instead of adding a field alongside the draft data fields. The originally approach ended up feeling messy and adding complexity. It required dynamically modifying schemas in the draft controller. It also put the snapshot ID along with the data fields, which I felt was a bit confusing and makes it so the PUT data for drafts and their underlying resources not longer matched. Internally, resource snapshot is stored in the payload field, but in the responses it is broken out and only the draft data is in the payload. This is another reason it felt more natural to keep the snapshot ID separate for the request.

Do we want to validate that the submitted snapshot actually exists? This would require an extra query in the service layer. Right now it will accept any integer (validated by the schema). Once the draft commit workflow is merged, an attempt to commit a draft based on a non-existent snapshot would result in a 404 error from the commit endpoint.

Right now the test simply sets an arbitrary non-existent snapshot ID to verify it was changed. I think the test can be improved once commit workflow is done since that is the real user story for this feature (attempt to commit, update base snapshot ID, re-attempt commit).

keithmanville:
After group discussion we determined that the implementation should be changed so that the snapshot ID be provided as part of the payload, not a query param. There should be a new draft schema that envelops the resource schemas and includes the snapshot ID external to the resource data.

keithmanville:
This is a breaking change to the draft modifications PUT endpoints.

The client is updated for this change, but the UI needs to be updated as well.

keithmanville:
For now the UI simply passes the current snapshot ID when modifying draft modifications to a queue. In the future we will want to make this part of the form. I recommend leaving this for when we design the UI around reconciling drafts.

src/dioptra/restapi/v1/queues/controller.py
```diff
@@ -207,6 +207,7 @@ def put(self, id: int):
     request_schema=QueueMutableFieldsSchema,
 )
 
+
```
keithmanville, line 210: remove this extra line

src/dioptra/client/drafts.py
```diff
@@ -373,7 +373,9 @@ def create(self, *resource_ids: str | int, **kwargs) -> T:
             json_=self._validate_fields(kwargs),
         )
 
-    def modify(self, *resource_ids: str | int, **kwargs) -> T:
+    def modify(
+        self, *resource_ids: str | int, resource_snapshot_id: str | int, **kwargs
+    ) -> T:
         """Modify a resource modification draft.
 
         Args:
```
jkglasbrenner, line 381: Client docstring needs to be updated.

tests/unit/restapi/v1/test_queue.py
```diff
@@ -587,8 +587,14 @@ def test_manage_existing_queue_draft(
     description = "description"
 
     # test creation
-    draft = {"name": name, "description": description}
-    draft_mod = {"name": new_name, "description": description}
+    draft = {
```
keithmanville, line 590: revert formatting changes

src/dioptra/restapi/v1/shared/drafts/service.py
```diff
@@ -467,7 +467,8 @@ def modify(
             return None, num_other_drafts
 
         current_timestamp = datetime.datetime.now(tz=datetime.timezone.utc)
-        draft.payload["resource_data"] = payload
+        draft.payload["resource_snapshot_id"] = payload["resource_snapshot_id"]
```
keithmanville, line 470: add check here to query for resource snapshot id to ensure it exists and that it is of the correct resource type. Also add check to enforce that the provided resource snapshot ID is >= the existing resource snapshot ID.

keithmanville: add temporary check that forces the new snapshot ID to be the same. This effectively prevents this new feature from being used. This check will be removed when the draft commit workflow feature is added.

jkglasbrenner (changes requested):

keithmanville:
Make sure to update docstrings to reflect new keyword args for ModifyResourceDraftsSubCollectionClient modify function.
This can be found in most subcollection clients.

Example for queues:
https://github.com/usnistgov/dioptra/blob/d307312b6a9d90c6c0b8e51a5a9fd569088d7d09/src/dioptra/client/queues.py#L116

jkglasbrenner (approved):
LGTM

Issue #625, opened 2024-08-16
Modify the behavior of the create resource button


keithmanville:
The create resource button currently opens a drawer with a single option to create the resource for the current page. The behavior should be modified so that it provides an option to create any resource. If the creation is canceled, it should go back to the previous page.

### Blocked by:
- #621

### Definition of done:
- [ ] The behavior of the create resource button is updated
- [ ] This feature is merged into dev

PR #715, opened 2025-01-16
feat: new create resource button

henrychoy:
issue: https://github.com/usnistgov/dioptra/issues/625

- The round fixed create resource button in the bottom right corner should give options to create any resource
- In the edit pages, the cancel button should go back to the previous page you were on


keithmanville (approved):
LGTM. @henrychoy  can you handle rebasing on dev?

PR #719, opened 2025-01-24
feat(restapi): added signature analysis workflow endpoint & tests

### LINKED ISSUE #691: Add a plugin task type inference workflow for missing types
This issue should implement the plugin task type inference workflow without checking the database for matches.

See documentation in #618 

### Definition of Done
- [x] Signature analysis functionality is added
- [ ] Document the new workflow endpoint in google docs
- [x] A new workflow that uses the signature analysis functionality is added
- [x] New tests are added for this workflow
- [x] All tests pass

jtsextonMITRE:
Closes #691

### DIFF
--- a/src/dioptra/client/workflows.py
+++ b/src/dioptra/client/workflows.py
@@ -22,6 +22,7 @@ from .base import CollectionClient, IllegalArgumentError
 T = TypeVar("T")
 
 JOB_FILES_DOWNLOAD: Final[str] = "jobFilesDownload"
+SIGNATURE_ANALYSIS: Final[str] = "pluginTaskSignatureAnalysis"
 
 
 class WorkflowsCollectionClient(CollectionClient[T]):
@@ -85,4 +86,23 @@ class WorkflowsCollectionClient(CollectionClient[T]):
 
         return self._session.download(
             self.url, JOB_FILES_DOWNLOAD, output_path=job_files_path, params=params
+        )
+
+    def analyze_plugin_task_signatures(self, python_code: str) -> T:
+        """
+        Requests signature analysis for the functions in an annotated python file.
+
+        Args:
+            python_code: The contents of the python file.
+            filename: The name of the file.
+
+        Returns:
+            The response from the Dioptra API.
+
+        """
+
+        return self._session.post(
+            self.url,
+            SIGNATURE_ANALYSIS,
+            json_={"pythonCode": python_code},
         )
--- a/src/dioptra/restapi/v1/workflows/controller.py
+++ b/src/dioptra/restapi/v1/workflows/controller.py
@@ -19,14 +19,19 @@ import uuid
 
 import structlog
 from flask import request, send_file
-from flask_accepts import accepts
+from flask_accepts import accepts, responds
 from flask_login import login_required
 from flask_restx import Namespace, Resource
 from injector import inject
 from structlog.stdlib import BoundLogger
 
-from .schema import FileTypes, JobFilesDownloadQueryParametersSchema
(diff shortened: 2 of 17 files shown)


src/dioptra/client/workflows.py
```diff
@@ -86,3 +87,41 @@ def download_job_files(
         return self._session.download(
             self.url, JOB_FILES_DOWNLOAD, output_path=job_files_path, params=params
         )
+
+    def signature_analysis_contents(
```
keithmanville, line 91: For the other workflow we rename this to use a verb, so `analyze_signature`.

It's also not very clear what kind of signature analysis we are talking about, so maybe we can be a bit more verbose with something like: `analyze_plugin_task_signatures`. Open to better suggestions.

src/dioptra/client/workflows.py
```diff
@@ -86,3 +87,41 @@ def download_job_files(
...
+            The response from the Dioptra API.
+
+        """
+        return self._session.post(
+            self.url,
+            SIGNATURE_ANALYSIS,
+            json_={"filename": filename, "fileContents": fileContents},
+        )
+
+    def signature_analysis_file(self, filename: str) -> T:
```
keithmanville, line 111: instead of having separate methods, can we overload a single method and either read from file or from contents depending on which args are passed?

This was the guidance James gave me for my import workflow. See here for an example:
https://github.com/usnistgov/dioptra/blob/16d340156d9bd544e4b567c46ebe9ca82aae88fd/src/dioptra/client/workflows.py#L96

src/dioptra/restapi/v1/lib/signature_analysis.py
```diff
@@ -0,0 +1,736 @@
...
+            This is an optional arg passed on to the underlying compile()
+            function, which documents it as:  "The filename argument should
+            give the file from which the code was read; pass some recognizable
+            value if it wasn't read from a file ('<string>' is commonly used)."
+
+    Yields:
+        Function signature information data structures, as dicts
+    """
+    if filepath:
+        ast = ast_module.parse(python_source, filename=filepath, feature_version=(3, 9))
```
keithmanville, line 687: Change to `feature_version=sys.version_info[0:2]` so this stays current to the version of python we are using.

We are currently using 3.11 in the worker containers if for some reason we don't want to stick to the version of python being used to run the function.

src/dioptra/restapi/v1/lib/signature_analysis.py
```diff
@@ -0,0 +1,736 @@
...
+    else:
+        ast = ast_module.parse(python_source, feature_version=(3, 9))
```
keithmanville, line 689: same as line 687

examples/DEPRECATED-pytorch-detectron2-demo/src/poison_train.py
keithmanville, line 1: why are these files in examples showing as changed?

src/dioptra/restapi/v1/workflows/controller.py
```diff
@@ -78,3 +83,36 @@ def get(self):
             mimetype=mimetype[parsed_query_params["file_type"]],
             download_name=download_name[parsed_query_params["file_type"]],
         )
+
+
+@api.route("/signatureAnalysis")
```
keithmanville, line 88: do we want to be more descriptive with the endpoint name?  `taskPluginSignatureAnalysis`?

would need the same change in the client.

keithmanville (changes requested):
Want to have a discussion to make sure we are capturing the types how we want in the output.

Issue #680, opened 2024-11-19
Figure out how to create a dioptra deployment with cruft with no user prompts


keithmanville:
Figure out how to pre-define cookiecutter variables and bypass the interactive prompt.

The test should invoke cruft as a cli command as a subprocess.

### Definition of Done:
- [ ] cruft prompts can be bypassed
- [ ] the new feature is documented as part of a technical documentation section in the sphinx docs
- [ ] tests are added to ensure the correct cookiecutter parameters were set properly
- [ ] The feature is merged into dev

keithmanville:
resolved by #720

PR #720, opened 2025-01-27
Cruft deployment with no user prompts

etrapnell-nist:
Tests for bypassing cruft prompts as specified in #680. The "Running Dioptra" doc has been updated to include the prompt-less cruft invocation.


keithmanville:
Is it possible to provide a (partial) configuration via list of parameters as part of the `cruft create` command or as a parameter that points to a config file? Or is the only option to skip the prompts and use the defaults from the repository?

For example, `cruft update` command has a `--variables-to-update` argument which takes a JSON string of parameters to set:
```
cruft update --checkout dev --variables-to-update '{ "docker_compose_path": "docker compose" }'
```

keithmanville:
I tested the promptless deployment as instructed in the new docs, and it worked as expected. I also pulled the branch and verified the new tests pass.

keithmanville:
The failing mypy tests should be resolved after rebasing on dev.

etrapnell-nist:
> Is it possible to provide a (partial) configuration via list of parameters as part of the `cruft create` command or as a parameter that points to a config file? Or is the only option to skip the prompts and use the defaults from the repository?
> 
> For example, `cruft update` command has a `--variables-to-update` argument which takes a JSON string of parameters to set:
> 
> ```
> cruft update --checkout dev --variables-to-update '{ "docker_compose_path": "docker compose" }'
> ```

Yes it is possible to skip the prompts while overriding the defaults. Use `cruft create` with these options as an example: `--no-input --extra-context '{"nginx_server_name": "dioptra2.example.com"}'`

keithmanville:
Nice! I confirmed that works. Can we add it to the documentation?

Can we also suggest the user may want to put their deployment config in a file and provide it like this:
```
cruft create https://github.com/usnistgov/dioptra --checkout dev \
  --directory cookiecutter-templates/cookiecutter-dioptra-deployment --no-input --extra-context "`cat deployment-config.json`"
```

etrapnell-nist:
I confirmed that an overrides file can be passed to cruft via the `--extra-context-file` option. I added both commands to the "Running Dioptra" doc. I tried to keep it concise, but having 4 choices for a single command feels a little bulky for a quickstart guide - what do you think?

etrapnell-nist:
I simplified the quickstart, keeping only the promptless cruft invocation and moved the other methods to the "Applying the template" section.

Issue #579, opened 2024-08-01
[BUG] Refreshing page on frontend when using containers leads to 404 error.


keithmanville:
**Describe the bug**
When running the frontend in the containers, refreshing the UI causes a 404 error, and the user needs to manually navigate back to the root URL.

This issue is documented here: 
https://router.vuejs.org/guide/essentials/history-mode.html#HTML5-Mode

Proposed fix is to add this to the nginx configuration for the REST API fallback service:
```
location / {
  try_files $uri $uri/ /index.html;
}
```

**To Reproduce**
1. Run the frontend in the container
2. refresh the page
3. The user will see a 404 error

ashley-rice:
I'm also having this issue. I have 2 questions: 

1. To what nginx configuration file should this fix be added? 
2. I am unable to submit a job as it seems like I am not getting a response from the API, despite the ports being exposed. Could this 404 error be related?

keithmanville:
resolved by #726

PR #726, opened 2025-01-30
fix: update nginx conf to fix frontend container refresh error

mtrapnell-nist:
Fixes the frontend refresh problem raised by issue #579.


jtsextonMITRE (approved):
works and looks good to me

Issue #724, opened 2025-01-29
[BUG] test_metrics test failing in the Windows GitHub runner


keithmanville:
`test_metrics` is failing in the Windows GitHub runner.

When a second value for the accuracy metric is posted, a subsequent GET returns the first value instead of the newer value.

see:
https://github.com/usnistgov/dioptra/actions/runs/13035219363/job/36366305054?pr=722

```
     def assert_job_metrics_matches_expectations(
        dioptra_client: DioptraClient[DioptraResponseProtocol],
        job_id: int,
        expected: list[dict[str, Any]],
    ) -> None:
        response = dioptra_client.jobs.get_metrics_by_id(job_id=job_id)
>       assert response.status_code == HTTPStatus.OK and response.json() == expected
E       AssertionError: assert (200 == <HTTPStatus.OK: 200> and [{'name': 'ac...value': 0.99}] == [{'name': 'ac...value': 0.99}]
E        +  where 200 = <tests.unit.restapi.lib.client.DioptraTestResponse object at 0x00000183AFF19850>.status_code
E        +  and   <HTTPStatus.OK: 200> = HTTPStatus.OK
E         
E         At index 0 diff: {'name': 'accuracy', 'value': 4.1} != {'name': 'accuracy', 'value': 4.2}
E         Use -v to get more diff)
```

keithmanville:
This behavior is not consistent.

Here the same test passes:
https://github.com/usnistgov/dioptra/actions/runs/13037442728/job/36371278927

jkglasbrenner:
It does not seem like #728 solved the issue, this GitHub Action ran after I rebased a branch on dev to include this change: https://github.com/usnistgov/dioptra/actions/runs/13062292399/job/36447856877

keithmanville:
resolved by #728

PR #728, opened 2025-01-30
fix: choose latest added metric if their timestamps are the same

jtsextonMITRE:
This should fix at least one of the scenarios that could cause #724, by ensuring that we choose the metric most recently added to the list if two of them have the same timestamp.


keithmanville:
add a comment explaining the reason for the `>=`

Issue #763, opened 2025-02-17
The exceptions raised from the python client configured for JSON responses do not include enough information for a user to debug.


keithmanville:
The exceptions include the response code, but no other details about the error. When the client is configured for requests responses, the response has the more detailed information from the REST API. Can we include this more detailed message in the exceptions raised by the client configured for JSON responses?

jkglasbrenner:
Closed by #764

PR #764, opened 2025-02-17
feat(client): add restapi error message to exceptions raised by client

keithmanville:
addresses #763

### DIFF
--- a/src/dioptra/client/sessions.py
+++ b/src/dioptra/client/sessions.py
@@ -118,7 +118,7 @@ def convert_response_to_dict(response: DioptraResponseProtocol) -> dict[str, Any
             response.request.url,
             response.text,
         )
-        raise StatusCodeError(f"Error code returned: {response.status_code}")
+        raise StatusCodeError(response.status_code, response.text)
 
     try:
         response_dict = response.json()
@@ -387,7 +387,7 @@ class BaseDioptraRequestsSession(DioptraSession[T], ABC, Generic[T]):
                 response.request.url,
                 response.text,
             )
-            raise StatusCodeError(f"Error code returned: {response.status_code}")
+            raise StatusCodeError(response.status_code, response.text)
 
         with output_path.open(mode="wb") as f:
             for chunk in response.iter_content(chunk_size=self.DOWNLOAD_CHUNK_SIZE):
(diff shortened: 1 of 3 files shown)


src/dioptra/client/sessions.py
```diff
@@ -118,7 +118,9 @@ def convert_response_to_dict(response: DioptraResponseProtocol) -> dict[str, Any
             response.request.url,
             response.text,
         )
-        raise StatusCodeError(f"Error code returned: {response.status_code}")
+        raise StatusCodeError(
+            f"Error code returned: {response.status_code}", response.text
```
jkglasbrenner, line 122: Recommend that we prepend a statement to contextualize what `response.text` should be understood as:

```suggestion
            f"Error code returned: {response.status_code}",
            f"Error message returned: {response.text}",
```

src/dioptra/client/sessions.py
```diff
@@ -387,7 +389,9 @@ def download(
                 response.request.url,
                 response.text,
             )
-            raise StatusCodeError(f"Error code returned: {response.status_code}")
+            raise StatusCodeError(
+                f"Error code returned: {response.status_code}", response.text
```
jkglasbrenner, line 393: Recommend that we prepend a statement to contextualize what `response.text` should be understood as:

```suggestion
                f"Error code returned: {response.status_code}",
                f"Error message returned: {response.text}",
```

jkglasbrenner (changes requested):
For context, I think my reasoning for omitting `response.text` from the raised error message was because the full details would be captured and returned if you turn on the logger and set the logging level to DEBUG. However, that requires you to do this in the first place, which isn't common in things like Jupyter notebooks, so it makes sense to also include it in the raised Exception. I like the solution of having it be a second argument to the Exception, since that returns a tuple of error messages.

In addition, you'll want to also make these changes in the test client, such as here:

https://github.com/usnistgov/dioptra/blob/aab45ae9b1914ddcf789bf178ca055ddf3180532/tests/unit/restapi/lib/client.py#L378-L379

keithmanville:
@jkglasbrenner I think I resolved your comment. Can you take a second look and merge to dev if it looks good?

tests/unit/restapi/lib/client.py
```diff
@@ -375,7 +375,10 @@ def download(
                 response.request.url,
                 response.text,
             )
-            raise StatusCodeError(f"Error code returned: {response.status_code}")
+            raise StatusCodeError(
+                f"Error code returned: {response.status_code}"
```
jkglasbrenner: This doesn't have a comma, but the others do. I think we want the comma, but should double-check to be sure @hbooth

Issue #709, opened 2025-01-13
Single click to open resource on tables


henrychoy:
**New Feature**
Replace current table selection functionality with allowing a single click to open a resource

**Definition of Done**
- Single click to open existing "edit" page or dialog
- Remove table radio buttons, edit button
- Move delete button from table to edit page
- Buttons inside table rows should behave as before and not open resource

PR #765, opened 2025-02-18
feat: Single click to open resource

henrychoy:
Closes issue #709, #781, #774

Single click to open resource.  Also added delete button on each table row.


keithmanville (approved):
Overall looks good to me and I think meets the definition of done in the issue card.

When viewing history, can we make it so the delete button is disabled (just like the submit button and edit fields)? It might not be clear to the user they would be deleting the whole resource and not a specific snapshot and would make it consistent with disabling any modifications when viewing history.

Also, curious to see what you think about using a different color for the cancel button to make it more distinct from deletion.

henrychoy:
> Overall looks good to me and I think meets the definition of done in the issue card.
> 
> When viewing history, can we make it so the delete button is disabled (just like the submit button and edit fields)? It might not be clear to the user they would be deleting the whole resource and not a specific snapshot and would make it consistent with disabling any modifications when viewing history.
> 
> Also, curious to see what you think about using a different color for the cancel button to make it more distinct from deletion.

Just updated to make the delete button disabled when viewing history.  I can do the same when I add history to the rest of the pages.  Also I made all cancel buttons more neutral looking so that the delete buttons stand out.

henrychoy:
FYI this PR also includes this small fix: #774

keithmanville:
@henrychoy I am squashing your commits and will merge these changes shortly.

One thing I wanted to note is about the link to the Entrypoint on the Jobs page. See the new issue card here: https://github.com/usnistgov/dioptra/issues/801

I think this is outside the scope of the current issues for this PR since we do not currently have the ability to view Entrypoint history

Issue #748, opened 2025-02-06
Update Model service layer to call MLFlow model version registration


keithmanville:
The `ModelIdVersionsService.create` method should also register the new model version with MLFlow via the client's [create_model_version](https://mlflow.org/docs/latest/python_api/mlflow.client.html#mlflow.client.MlflowClient.create_model_version) method.  The name is derived from the Model's resource ID. The `source` is the artifact URI and the `run_id` is the mlflow run ID attached to the job that created the artifact.

The `ModelIdVersionsNumberService.modify` method does not need to be updated since the only thing that can be changed is the description which does not affect the registered model in MLFlow.

Example of mocking MLFlow in the tests:
https://github.com/usnistgov/dioptra/blob/dev/tests/unit/restapi/lib/mock_mlflow.py
https://github.com/usnistgov/dioptra/blob/8e1688a037d8c0f6f1b775433a519be957a7ada0/tests/unit/restapi/v1/test_job.py#L463

### Definition of Done
- [ ] Model Version creation in the Dioptra REST API properly registers the new version with MLFlow
- [ ] New tests are added to cover the new functionality (via mocking MLFlow)
- [ ] The feature is merged into dev

Issue #747, opened 2025-02-06
Update Model service layer to register the model in MLFlow


keithmanville:
The `ModelService.create` method should also create a registered model in MLFlow using the client [create_registered_model](https://mlflow.org/docs/latest/python_api/mlflow.client.html#mlflow.client.MlflowClient.create_registered_model) method. The model name should be based on the resource ID of the model, e.g. f"model_{resource_id}". The implementation will need to use `db.session.flush()` so that the newly created model resource ID can be accessed before committing staged changes to the database.

We need to ensure that in the case of an error in registering the model namespace in either Diotpra or MLFlow, the changes can be rolled back to the original state in both Dioptra and MLFlow.

`ModelIdService.modify` and `ModelIdService.delete` do not require changes. modify actions do not need to change the underlying registered model in MLFlow and delete actions do not delete the registered MLFlow model.

Example of mocking MLFlow in the tests:
https://github.com/usnistgov/dioptra/blob/dev/tests/unit/restapi/lib/mock_mlflow.py
https://github.com/usnistgov/dioptra/blob/8e1688a037d8c0f6f1b775433a519be957a7ada0/tests/unit/restapi/v1/test_job.py#L463

### Definition of Done:
- [ ] Model creation in the DIoptra REST API properly registers the model namespace in MLFlow
- [ ] New tests are added to cover the new functionality (by mocking mlflow client)
- [ ] The feature is merged into dev

PR #768, opened 2025-02-19
feat: update model service layer to register mlflow model

keithmanville:
resolves:
- #747
- #748

### DIFF
--- a/src/dioptra/restapi/v1/models/service.py
+++ b/src/dioptra/restapi/v1/models/service.py
@@ -136,7 +136,16 @@ class ModelService(object):
         except MlflowException as e:
             raise MLFlowError(e.message) from e
 
-        db.session.commit()
+        try:
+            db.session.commit()
+        except Exception as e:
+            # if the commit fails, attempt to roll back mlflow changes
+            try:
+                self._mlflow_client.delete_registered_model(mlflow_model_name)
+            except MlflowException as mlflow_error:
+                raise MLFlowError(str(e)) from mlflow_error
+            raise e
+
         log.debug(
             "Model registration successful",
             model_id=ml_model.resource_id,
--- a/src/dioptra/restapi/v1/models/service.py
+++ b/src/dioptra/restapi/v1/models/service.py
@@ -21,8 +21,6 @@ from typing import Any, Final, cast
 import structlog
 from flask_login import current_user
 from injector import inject
-from mlflow.exceptions import MlflowException
-from mlflow.tracking import MlflowClient
 from sqlalchemy import Integer, func, select
 from sqlalchemy.orm import aliased
 from structlog.stdlib import BoundLogger
@@ -39,6 +37,8 @@ from dioptra.restapi.v1 import utils
 from dioptra.restapi.v1.artifacts.service import ArtifactIdService
 from dioptra.restapi.v1.groups.service import GroupIdService
 from dioptra.restapi.v1.shared.search_parser import construct_sql_query_filters
+from mlflow.exceptions import MlflowException
+from mlflow.tracking import MlflowClient
 
 LOGGER: BoundLogger = structlog.stdlib.get_logger()
 
--- a/src/dioptra/restapi/v1/models/service.py
+++ b/src/dioptra/restapi/v1/models/service.py
@@ -129,10 +129,8 @@ class ModelService(object):
 
         db.session.flush()
 
-        mlflow_model_name = f"resource_{ml_model.resource_id:09d}"
-
         try:
-            self._mlflow_client.create_registered_model(mlflow_model_name)
+            self._mlflow_client.create_registered_model(str(ml_model.alternative_id))
         except MlflowException as e:
             raise MlflowError(e.message) from e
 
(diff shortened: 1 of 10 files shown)


src/dioptra/restapi/v1/models/service.py
```diff
@@ -118,16 +122,32 @@ def create(
             resource=resource,
             creator=current_user,
         )
-
         db.session.add(ml_model)
 
-        if commit:
+        db.session.flush()
+
+        mlflow_model_name = f"resource_{ml_model.resource_id:09d}"
```
jkglasbrenner, line 129: Is there a need to have the zero-padding in the mlflow model name, e.g. `{ml_model.resource_id:09d}` instead of `{ml_model.resource_id:d}`?

src/dioptra/restapi/v1/models/service.py
```diff
@@ -65,6 +68,7 @@ def __init__(
         self,
         model_name_service: ModelNameService,
         group_id_service: GroupIdService,
+        mlflow_client: MlflowClient,
```
jkglasbrenner, line 71: As a first pass, I'm okay with injecting the third party `MlflowClient` directly, but as a technical debt follow-up issue I'd like to see its behavior wrapped similar to what we did with the `rq` library, where we make a dedicated class in the `shared` folder to create a standardized interface for the behaviors we want: https://github.com/usnistgov/dioptra/blob/71a646084266e379815d032fc48fea680e97d639/src/dioptra/restapi/v1/shared/rq_service.py

That also allows us to encapsulate logging and try/except handling in a single place that's outside of the services used directly in the endpoints.

So yeah, only action for this comment is to create an issue so we don't lose track of this.

src/dioptra/restapi/v1/models/service.py
```diff
@@ -550,14 +571,23 @@ def create(
 
         ml_model.resource.children.append(new_version.resource)
         db.session.add(new_version)
+        db.session.flush()
```
jkglasbrenner, line 574: Why is the flush needed here? Isn't `ml_model.resource_id` already accessible without flushing?

src/dioptra/restapi/v1/models/service.py
```diff
@@ -550,14 +571,23 @@ def create(
...
 
-        if commit:
-            db.session.commit()
-            log.debug(
-                "Model registration successful",
-                model_id=ml_model.resource_id,
-                name=ml_model.name,
+        mlflow_model_name = f"resource_{ml_model.resource_id:09d}"
```
jkglasbrenner, line 576: Same question as before on whether the `:09d` padding is necessary or not.

src/dioptra/restapi/v1/models/service.py
```diff
@@ -550,14 +571,23 @@ def create(
...
+        try:
+            self._mlflow_client.create_model_version(
+                mlflow_model_name, source=artifact.uri
             )
+        except MlflowException as e:
+            raise MLFlowError(e.message) from e
+
+        db.session.commit()
```
jkglasbrenner, line 584: Is it possible to delete a model version after you've created it in Mlflow? If so, then we need to wrap this in an try/except to allow rollbacks? If not, then I'm not sure how we handle the inconsistent state.

src/dioptra/restapi/v1/models/service.py
```diff
@@ -550,14 +571,23 @@ def create(
...
-        if commit:
-            db.session.commit()
-            log.debug(
-                "Model registration successful",
-                model_id=ml_model.resource_id,
-                name=ml_model.name,
+        mlflow_model_name = f"resource_{ml_model.resource_id:09d}"
+        try:
+            self._mlflow_client.create_model_version(
+                mlflow_model_name, source=artifact.uri
```
jkglasbrenner, line 579: How are we linking our version numbers with the ones in the MLflow database? Are we assuming a 1-to-1 correspondence in version numbers? What do we do if something happens and the version numbers suddenly are off-by-1 (or more) between the databases?

src/dioptra/restapi/v1/models/service.py
jkglasbrenner, line 576: I am also wondering whether it's better to just generate a random UUID for each model name, this feels brittle because its a naming convention as opposed to an explicit declaration in the database.

jkglasbrenner (changes requested):
I went through this, most comments are small, but the bigger thing I noticed is how we're linking our database <-> mlflow tracking database. It seems to be based around naming and numbering conventions that feel brittle, I could imagine someone going into the database and deleting something manually, and if you're using SQlite, it will recycle older primary IDs, meaning a name collision is possible.

We could punt and add database tracking later, although that could make for a complex migration script (it could have to talk to Mlflow Tracking server).

src/dioptra/restapi/v1/models/service.py
keithmanville: no, not necessary.

src/dioptra/restapi/v1/models/service.py
keithmanville: I think I was trying to be as sure as possible that the transaction could be commited before doing anything with mlflow that might need a more complicated rollback.

jkglasbrenner:
Likely to be affected by changes in #658, so this is blocked until that PR is fully resolved.

keithmanville:
OBE

PR #771, opened 2025-02-26
(GUI) Set a new resource's default group selection to the user's group

mtrapnell-nist:
This GUI change sets the group field on the different resource creation pages to automatically populate with the user's group.

### DIFF
--- a/src/frontend/src/dialogs/AddTagsDialog.vue
+++ b/src/frontend/src/dialogs/AddTagsDialog.vue
@@ -75,6 +75,9 @@
       name.value = ''
       locked.value = true
     }
+    if (!group.value) {
+      group.value = store.loggedInGroup.id
+    }
   })
 
   function emitAddOrEdit() {
--- a/src/frontend/src/dialogs/ModelsDialog.vue
+++ b/src/frontend/src/dialogs/ModelsDialog.vue
@@ -88,6 +88,9 @@
       name.value = ''
       description.value = ''
     }
+    if (!group.value) {
+      group.value = store.loggedInGroup.id
+    }
   })
 
   function emitAddOrEdit() {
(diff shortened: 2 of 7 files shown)


henrychoy:
Looks great to me!  One minor detail is that when editing an experiment or entrypoint, then leaving without saving, then returning, it'll prompt you if you want to load the previous values, if you say cancel, then we probably shouldn't clear out the group.  You can fix this by editing the `clearForm()` function in those files.

keithmanville:
LGTM as well. I'd recommend this for merging once you address Henry's comment.

The git commit message will need to be re-written to pass gitlint.

I saw the sorting tags test fail. I saw this crop up somewhere before, so I created an issue on it here with some thoughts on the problem and suggestions for fixing it: https://github.com/usnistgov/dioptra/issues/772

mtrapnell-nist:
Thanks for the review! When attempting to implement the suggestion to not clear the group field in clearForm(), I noticed a different unexpected behavior which I am now looking into. 

- Click create new entrypoint and enter data, 
- leave without saving, 
- click create an entrypoint again, 
- click cancel on the prompt to load previous values, 
- then immediately leave the page without saving again,
- click create entrypoint a third time,
- it prompts a second time to load saved values

It only happens on the first 2 times after trying to create the entrypoint, then everything works as expected. This is consistent for both experiments and entrypoints.

henrychoy:
> Thanks for the review! When attempting to implement the suggestion to not clear the group field in clearForm(), I noticed a different unexpected behavior which I am now looking into.

As discussed on the call, this PR is good to merge as-is.  And I can look at not clearing out the group separately here: https://github.com/usnistgov/dioptra/issues/774

henrychoy (approved):

PR #777, opened 2025-03-05
fix: update frontend npm packages

henrychoy:
This PR updates all npm packages for the frontend.  There were some dependency errors I had to fix manually.  Also there was only one breaking change which had to do with how `quasar-variables.sass` was being imported, this includes that fix.  This PR also resolves some minor warnings on the misuse of the `<caption>` tag.

As before, all that is needed is to `npm install`.  If you see any errors, please let me know.


mtrapnell-nist:
This looks good to merge. No errors occurred after running `npm install`.

mtrapnell-nist (approved):
This looks good to merge. No errors occurred after running `npm install`.

keithmanville (approved):

PR #780, opened 2025-03-06
feat: add script for importing resources to the deployment directory

keithmanville:


JustKuzya:
1. I'm completely OK with either Click and ArgParse Pythonic ways of handling docu-manuals, parameters, and defaults, but in Dioptra (it seems to me) we have predominantly used the Click way. I think click (arguably) gives cleaner code. I’m certainly NOT asking for a rewrite, just want to have a more clear project policy guidance for the future: “We use only Click” or “We use what developer finds easier to use in each context” - something of that sort. 🤷

2. OK looking docu-manual with --help option 👍

3. Weird behavior when executed **without input parameters** `python ./import_resources.py` - it interactively asks for password (yet despite neither DIOPTRA_USERNAME nor DIOPTRA_PASSWORD are present in env) the script doesn't ask for user name and the rest of the parameters, only password.

chisholm:
I prefer argparse, and used it in the [run-experiment](https://github.com/usnistgov/dioptra/blob/0fef796574d462081827539ee2dfa17a4e4eb3c1/src/dioptra/task_engine/run_experiment.py) and [validate-experiment](https://github.com/usnistgov/dioptra/blob/0fef796574d462081827539ee2dfa17a4e4eb3c1/src/dioptra/task_engine/validate.py) scripts.

I'm not such a fan of the mile high stack of decorators approach of click.  It is not good code structure for me.

hbooth:
> 1. I chose argparse since it is part of the standard library and we are recommending this to be used from the deployment venv and I wanted to limit any additional dependencies. I agree we should have some kind of policy here. Maybe something to be added to the dev docs.

I generally prefer using the same library throughout to minimize the learning curve of a project. On the other hand, I like the idea of using what is already in the standard library as that minimizes on dependencies, but I do realize that click is already a dependency for other parts of the project. Perhaps we continue to use click for the Flask related parts (where click is already a dependency and that is how you need to integrate), but for any of our own command-line scripts we use argparse?

JustKuzya (approved):
It looks logically OK and seems to work as expected (though strangely, unless you have not the environment set) 👍

Issue #772, opened 2025-02-27
test_tag_sort fails sometimes


keithmanville:
The `test_tag_sort` is inconsistent and fails some times.

I suspect it may fail when no sort criteria is provided and we can't expect a specific order ( the `(None, None, ["tag1", "tag2", "tag3"])` case below).

```python
    @pytest.mark.parametrize(
        "sort_by, descending , expected",
        [
            (None, None, ["tag1", "tag2", "tag3"]),
            ("name", True, ["tag2", "tag1", "tag3"]),
            ("name", False, ["tag3", "tag1", "tag2"]),
            ("createdOn", True, ["tag3", "tag2", "tag1"]),
            ("createdOn", False, ["tag1", "tag2", "tag3"]),
        ],
    )
    def test_tag_sort(
    ...
```

I think Andy encountered something similar here with the returned order of Queues:
https://github.com/usnistgov/dioptra/blob/f5f355896304b1187d67bc34e08e451efc90f8c2/tests/unit/restapi/v1/test_queue.py#L158

Since we don't have a guarantee on the returned order, the test should only check that the correct tags are present and not assume an order.

Currently known tests that check based on order but don't specify one are: Queue Retrieval, Queue Sort (with None, None for sort field and order) and Tags Sort (with None None for sort and order). As part of this issue, check tests the look based on sort_by and do not provide a field and any tests that perform retrieval and check based on a python lists instead of a set.

keithmanville:
resolved by #788

PR #788, opened 2025-03-19
Test resource sort sometimes fails

etrapnell-nist:
Address the test sort issue as described in #772

### DIFF
--- a/tests/unit/restapi/v1/test_artifact.py
+++ b/tests/unit/restapi/v1/test_artifact.py
@@ -30,6 +30,7 @@ from dioptra.client.base import DioptraResponseProtocol
 from dioptra.client.client import DioptraClient
 
 from ..lib import helpers
+from ..test_utils import assert_retrieving_resource_works
 
 # -- Assertions ------------------------------------------------------------------------
 
@@ -127,6 +128,8 @@ def assert_retrieving_artifacts_works(
     dioptra_client: DioptraClient[DioptraResponseProtocol],
     expected: list[dict[str, Any]],
     group_id: int | None = None,
+    sort_by: str | None = None,
+    descending: bool | None = None,
     search: str | None = None,
     paging_info: dict[str, Any] | None = None,
 ) -> None:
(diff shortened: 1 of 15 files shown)


JustKuzya (approved):
The changes and the test/model changes merging the 711 into 772 PR look OK.

The only open question is the ordering of nodes in JSON objects do we want to maintain test independence, or we want to have some universally available method to compare **well-ordered** JSON objects? (It is more of a general project test-philosophy question)

etrapnell-nist:
There was a conflict with the dev branch that I had to resolve manually.

tests/unit/restapi/v1/test_artifact.py
```diff
@@ -205,6 +211,49 @@ def assert_sorting_artifact_works(
...
+
+    if group_id is not None:
+        query_string["group_id"] = group_id
+
+    if search is not None:
+        query_string["search"] = search
+
+    if paging_info is not None:
+        query_string["index"] = paging_info["index"]
+        query_string["page_length"] = paging_info["page_length"]
```
hbooth, line 249: The code here seems repeated several times throughout this commit. Would it make sense to create utility function to perform this logic?

tests/unit/restapi/v1/test_artifact.py
```diff
@@ -295,6 +344,43 @@ def test_artifacts_get_all(
...
+    - The returned list of artifacts matches regardless of the order in
+        the parametrize lists above.
+    """
+    expected_ids = set(
+        [registered_artifacts[expected_name]["id"] for expected_name in expected]
+    )
+    assert_sorting_artifact_by_none_works(
+        dioptra_client, sort_by=sortBy, descending=descending, expected=expected_ids
+    )
+
```
hbooth, line 379: Same comment for here. This logic seems repeated.

hbooth (commented):
Please take a look to see if it makes sense to create a test utility function for validation the sort and none operations.

tests/unit/restapi/v1/test_artifact.py
JustKuzya: These place collect the local information for building a query passed as `**query_string` into `artifacts.get` call. All three(3) places (where it occurs) have different sub-sets of the query_string indexers. So, while it is possible to refactor, I'm not sure that it would improve either the code readability or will be easy to reuse. I suggest to leave these as is.

tests/unit/restapi/v1/test_artifact.py
JustKuzya: All the repetitive code is locked inside of the assertions, which are the reusable code aggregations already.

keithmanville (changes requested):
`assert_sorting_queue_works`, `assert_sorting_queue_by_none_works`, and `assert_retrieving_queues_works` are all very similar in that they call the `GET /queues` and check to see if the response is correct.

I'm proposing a refactor in which we can use a assert function (`assert_retrieving_queues_works`) to handle all of the tests. I added the sorting params (`sort_by` and `descending`). Then if `sort_by` is not provided, it uses @JustKuzya's normalized compare function, otherwise it does direct dict comparison to ensure the order of the response is correct.

Additionally, I don't think we need `test_queue_sort_by_none`. This is already handled by the existing retrieval test. The parameterized version is just sorting the expected values, not actually exercising different calls to the API.

keithmanville:
I made the proposed changes to `test_queue.py` here: https://github.com/usnistgov/dioptra/pull/788/commits/eef92c34edebae0efa07e35ab4b2dad820a313fb

Can you take a look and see if you agree with these changes, and confirm I didn't miss something in what we were trying to test? If you concur with these changes, the same pattern can be followed in the other tests.

etrapnell-nist:
All tests are now passing.

Issue #655, opened 2024-10-11
As a user I would like to be able to store output from a single task plugin or from a combination of task plugins as artifacts for later use and analysis.


hbooth:
How to store an artifact is not obvious and requires each plug-in that wishes to store data to implement this functionality. In addition, it requires access to/knowledge of MLFlow. Providing built-in functionality that allows for a declarative model for artifacts would address these issues.

**Blocked By**:
- #575 

**Definition of Done**
- [x] Extension of Entry Point YAML file to support declarations of artifacts
- [x] Task Engine can validate this new file
- [x] Task engine can produce output appropriate for running a job which appropriately breaks apart the task graph and artifacts
- [x] While running a job the appropriate artifact creation hooks are called which store the artifacts as appropriate
- [x] All new functionality has appropriate tests
- [x] Merged into dev branch

**Describe alternatives you've considered**
A potential alternative would be to create a "utility" plug-in that implements this functionality that could be called, but it would not address the concerns around MLFlow and other resource access.

jkglasbrenner:
Closed by #899

PR #790, opened 2025-03-20
Adding RESTAPI & DB changes for Artifacts

jtsextonMITRE:
Part of #655


jtsextonMITRE:
Note that the "Files Changed" is large because it includes updates from dev which are needed for the database update

PR #796, opened 2025-03-25
fix: fix job_timeout parameter in rq

### LINKED ISSUE #787: [BUG] Timeout parameter not being respected for jobs, instead defaults to 24h
**Describe the bug**
No matter what timeout you pass in you get 24h as the timeout, which is the default timeout we give to RQ.

**Additional context**
The parameter **timeout** [here
](https://github.com/usnistgov/dioptra/blob/main/src/dioptra/restapi/v1/shared/rq_service.py#L61)
should be changed to **job_timeout** to fix this.

jtsextonMITRE:
Closes #787.


keithmanville:
I tested this with a job that included a `time.sleep`. I confirmed that on dev, the timeout is not triggered and that this branch fixes the issue.

PR #797, opened 2025-03-25
feat(frontend): add component to import resources from git repo

### LINKED ISSUE #718: Add new UI elements to enable plugin import functionality to the frontend
Blocked by:
- #588

keithmanville:
This is a draft PR that demonstrates the basic functionality. The import form is a new Vue component. Need to figure out where in the frontend it should go. The form should be expanded to accept the other options for importing resources (archive file and files).

closes #718


keithmanville:
couple notes on testing this:
- to import from git you have to use the dev branch since the plugin import workflow is not in main. I have it default to main so we don't forget to switch it when we do a release
- if you want to test the archive upload you can create a tar like this from the root dioptra dir and it should work: `tar -czf plugins.tar.gz extra`

henrychoy:
May just be me, but I'm getting a unicode error when trying to import the archive file?  Using `tar -czf plugins.tar.gz extra` in the root directory.

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa3 in position 45: invalid start byte [werkzeug]
```

src/dioptra/restapi/v1/workflows/controller.py
```diff
@@ -162,7 +162,7 @@ def post(self):
             source_type=parsed_form["source_type"],
             git_url=parsed_form.get("git_url", None),
             archive_file=request.files.get("archiveFile", None),
-            files=request.files.getlist("files", None),
+            files=request.files.getlist("files") + request.files.getlist("files[]"),
```
jkglasbrenner, line 165: Is `files[]` a Vue-ism that can't be avoided?

src/frontend/src/dialogs/ImportResourcesDialog.vue
```diff
@@ -0,0 +1,145 @@
...
+  const isMobile = inject('isMobile')
+  const isExtraSmall = inject('isExtraSmall')
+
+  const sourceTypeOptions = [
+    {slot: 'git', value: 'git'},
+    {slot: 'upload_archive', value: 'upload_archive'},
+	  {slot: 'upload_files', value: 'upload_files'},
+  ];
+  const sourceType = ref(sourceTypeOptions[0].value);
+  const gitUrl = ref('https://github.com/usnistgov/dioptra.git;extra#main');
```
jkglasbrenner, line 113: What's the `;extra` for? It's not a syntax I'm familiar with, and it didn't seem to be needed when I tried the code out.

jkglasbrenner (commented):
The import interface looks nice, and is pretty intuitive to use. I left a couple of questions, but they're minor.

src/frontend/src/dialogs/ImportResourcesDialog.vue
keithmanville: it's a comma separated list of paths I use for sparse-checkout. See here:
https://github.com/usnistgov/dioptra/blob/8bda1136798db761db3fea06a85559c4c46e018d/src/dioptra/restapi/v1/workflows/lib/clone_git_repository.py#L50

One of the things I figured out when trying to make the cloning as fast as possible. It's not necessary and in this case maybe the potential confusion of including it as a default outweighs a small speed up?

src/dioptra/restapi/v1/workflows/controller.py
keithmanville: I couldn't figure out how to get it to send without the `[]`. I'm not sure if one one way or the other is more standard. I've seen that notation before when sending files with curl, but I'm not sure if the server receives the `[]` as well.

I can look into it a little more. Honestly, I just took the easy route of accepting both options instead of looking into it deeper.

jkglasbrenner: Yeah, on the Python side the `files` field without `[]` is possibly set by the `MultipartEncoder` from `requests_toolbelt` or its a setting used within vanilla `requests` itself. It doesn't look like the Python client explicitly chooses to use it (otherwise we could just refactor the Python client to be consistent with what Vue.js does).

I think both options are valid from a conventions point-of-view, although I wish it was possible to just use one instead of both. Maybe Henry knows if there's an easy way to change this on the Vue side? If not, then its fine and isn't a big deal, just wanted to double check.

src/frontend/src/dialogs/ImportResourcesDialog.vue
jkglasbrenner: I'm fine with having it as an option if its documented somewhere. Since you're using `urlparse` on it, its a valid web URL, but my confusion came from knowing that `git clone https://github.com/usnistgov/dioptra.git;extra#main` won't work, so its not a valid git url.

Main thing is documentation in the UI I think. This is where having a little ⓘ icon that provides a tooltip explaining the non-standard syntax when moused over would be nice.

keithmanville:
@jkglasbrenner this is ready for a followup review.

I rebased and squashed commits. I kept some of the service layer fixes that I exposed in developing this separate from the frontend changes. I also kept the filename regex fix separate.

henrychoy (approved):
Looks good, tested importing from url, archive file, and directory, and everything looked good.

jkglasbrenner (approved):
I think it looks good now, I defer any other minor Vue-related changes to @henrychoy's knowledge.

PR #800, opened 2025-03-26
docs: update API docs for paging

### LINKED ISSUE #793: Page index API documentation can be misleading
**Describe the documentation issue**
In the API docs, it says index is the "Index of the current page." From my own misunderstanding, I thought this meant, for example, `70 results / 10 per page = 7 pages`, and `index=2` selects the second page, i.e. results `11` through `20`. What it actually means is it will select 10 results starting from index 2, i.e. results `2`-`11`.

We should update this so that it says the index is the "Starting index of the current page." to avoid confusion.

**Section of Documentation**
This is in the API docs for the page index.

jtsextonMITRE:
Closes #793.

### DIFF
--- a/src/dioptra/restapi/v1/schemas.py
+++ b/src/dioptra/restapi/v1/schemas.py
@@ -152,7 +152,7 @@ class BasePageSchema(Schema):
 
     index = fields.Integer(
         attribute="index",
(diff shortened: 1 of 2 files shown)


keithmanville:
Resolved by updating documentation per Harold's recommendation in slack conversation.

Issue #742, opened 2025-02-05
Display metrics in Jobs UI


keithmanville:
This feature will expose metrics information in the Jobs page of the UI. It should be a new tab alongside overview. This issue is scoped to metrics for a single job. Displaying metrics across multiple jobs in an experiment is a future feature.

It should have a table of metrics from `/api/v1/jobs/{id}/metrics`:
```
[
  {
    "name": "string",
    "value": 0
  }
]
```

And it should be possible to select a metric to see it's history:
`/api/v1/jobs/{id}/metrics/{name}/snapshots`
```
{
  "index": 0,
  "isComplete": true,
  "totalNumResults": 0,
  "first": "string",
  "next": "string",
  "prev": "string",
  "data": [
    {
      "name": "string",
      "value": 0,
      "step": 0,
      "timestamp": 0
    }
  ]
}
```

Start by mocking up ideas for how this could look in the UI.

Look at plotly (https://plotly.com/javascript/) to handle plotting.

### Blocked by:
- #781

PR #812, opened 2025-04-02
feat: Plotly graph, updated

henrychoy:
Closes: #742 

This PR adds a re-useable `PlotlyGraph` component that is used in the metrics tab of the job dashboard.  This and dev has many newly updated packages.  Please do `npm install` or better yet `npm ci` for a clean install.

### DIFF
--- a/src/frontend/package-lock.json
+++ b/src/frontend/package-lock.json
@@ -17,6 +17,7 @@
         "codemirror": "^6.0.1",
         "js-yaml": "^4.1.0",
         "pinia": "^3.0.1",
+        "plotly.js-dist-min": "^3.0.1",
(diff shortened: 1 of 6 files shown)


keithmanville (commented):
Plots look good to me. The add data buttons need to be removed before merging

A couple thoughts:
- it might be nice to be able to click on a metric on the overview page which goes to the plot on the other tab
- it might be nice to plot multiple metrics on the same plot. But not sure how we want to handle that
- maybe we could have a way to download the metrics data? but probably not a high priority since it can be done via the python client.

I would be fine with this PR as is (after debug buttons are removed). But want to see what others think about potential design changes or improvements.

Issue #559, opened 2024-07-24
GUI- Add ability to rerun a job


mtrapnell-nist:
**Is your feature request related to a problem? Please describe.**
When a job fails to run, it would be helpful to have an option to either rerun the job, or having a button that copies the parameters of an existing job into a new job creation window. This would allow for easier testing, modification, and rerunning of jobs.

The entrypoint may have changed between runs, so some kind of reconciliation is required.
If a name isn't matched, it should be omitted with warnings to the user.
If an dependent resource has been deleted an error should be returned.

blocked by #674

jkglasbrenner:
Closed by #833

PR #833, opened 2025-04-16
GUI- Add ability to rerun a job

mtrapnell-nist:
This will resolve issue #559.

### DIFF
--- a/src/frontend/src/views/CreateJob.vue
+++ b/src/frontend/src/views/CreateJob.vue
@@ -529,14 +529,39 @@
     }
   }
 
+  async function getResource(type, id) {
+    try {
+      const res = await api.getItem(type, id)
+      return res.data
+    } catch(err) {
+      console.log(err)
+    }
+  }
+
   onMounted(async () => {
+    if(history.state.oldJobId) {
+      const oldJob = await getResource('jobs', history.state.oldJobId)
+      console.log('oldJob = ', oldJob)
+      await getExperiment(oldJob.experiment.id)
+      if(allowableEntrypointIds.value.includes(oldJob.entrypoint.id)) {
+        await getEntrypoint(oldJob.entrypoint.id)
+      } else {
+        notify.error(`${oldJob.entrypoint.name} is no longer linked to ${oldJob.experiment.name}`)
+      }
+      if(allowableQueueIds.value.includes(oldJob.queue.id)) {
+        job.value.queue = await getResource('queues', oldJob.queue.id)
+      } else {
+        notify.error(`${oldJob.queue.name} is no longer linked to ${oldJob.entrypoint.name}`)
+      }
+      job.value.description = oldJob.description
+    }
     if(Object.hasOwn(route.params, 'id')) {
       await getExperiment(route.params.id)
     } else {
       await getExperiments()
     }
 
-    if(store.savedForms.jobs[expJobOrAllJobs.value]) {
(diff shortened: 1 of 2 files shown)


keithmanville:
@mtrapnell-nist Can we make it so that the entrypoint parameter values are also set to the same values used in the job?

I ran a hello world job with name changed to something else. When I click re-run, I get the correct entrypoint and queue in the job creation screen, but the value for name is the default value for the entrypoint ("world") instead of the value I used in the job.

henrychoy:
> @mtrapnell-nist Can we make it so that the entrypoint parameter values are also set to the same values used in the job?
> 
> I ran a hello world job with name changed to something else. When I click re-run, I get the correct entrypoint and queue in the job creation screen, but the value for name is the default value for the entrypoint ("world") instead of the value I used in the job.

@mtrapnell-nist To get the original parameter default values from the old job, I think we need to pass `snapshotId` of the entrypoint of the old job in history (in addition to the oldJobId that you're already passing), and when the job form re-loads, you need to get that specific snapshot (new endpoint is needed in `dataApi`) of the entrypoint and overwrite the default param values with the old ones.  Let me know if you need any help!

henrychoy:
> @henrychoy This should be ready for another review when you have a chance. Thanks!

This works great, however looks like there's an edge case we didn't think of yet.  If you re-run a job with an entrypoint where a param has been added since the old job ran, the default value doesn't show in the chart, and the job is submitted without that new param.  And also, if you switch to a new entrypoint, it will show the params but not the default values.  

I think this can all be fixed from the `job.value.entrypoint` watcher.  Esentially we need a if check to check if the selected entrypoint is different from the oldJob one, and if it is, just use the old logic.  If it's the same entrypoint, then also use the old logic, but check if it has the same params of the oldJob, and if so swap out the value.  It's a little complicated explaining through text but let me know if that made sense.

mtrapnell-nist:
@henrychoy I made some more changes. 

The issue I am running into now is that even when submitting the job with the old parameters and values in the case of a changed entrypoint, any new parameters are still included even when not choosing to update the parameters. It seems that we might need the ability to create the job with the specific entrypoint snapshot matching when it was submitted with the original job if the user wants to re-run a job exactly as it was. Unless I am missing something, I don't see a way to do this currently.

jtsextonMITRE:
Am I doing something wrong? I get POST errors when I click Submit Job, though the fields do appear to be populated correctly and the python client is still able to run the job. 

image

jtsextonMITRE (approved):
works for me. was able to rerun an MNIST job with the same and changed parameters as the old one

henrychoy:
@mtrapnell-nist Sorry finally got a chance to review this again.  I just see 2 issues below, but otherwise it works great.

- `entrypointSnapshotId` is not needed in the payload, it causes the POST call to error out.  Works when removed.
- It errors out when trying to clear out the entrypoint dropdown.  To fix, you need a ? before each job.entrypoint.property in both the button if's below.

```
          v-if="!updateEntrypoint && job.entrypoint.id === oldEntrypoint?.id && 
          oldEntrypoint?.snapshot !== latestEntrypoint?.snapshot"
```

```
          v-if="updateEntrypoint && job.entrypoint.id === oldEntrypoint?.id && 
          job.entrypoint.snapshot === latestEntrypoint?.snapshot"
```

jkglasbrenner (approved):
Made a couple of final changes to fix the broken unit tests, I updated the Python client to include `entrypoint_snapshot_id` as an optional parameter when creating a new job, and I added a new unit test to cover this new functionality. Looks good now.

PR #834, opened 2025-04-17
fix: Convert NaN's to None and vice-versa in the metrics client/endpoint

### LINKED ISSUE #810: [BUG] Metrics endpoint and client should be able to accept NaN
**Describe the bug**
The client does not allow the passing of NaN into `append_metrics_by_id` due to it being disallowed by JSON.

We should modify this so it transforms NaN into a string and the receiving endpoint recognizes it as NaN.

jtsextonMITRE:
Closes #810.

### DIFF
--- a/src/dioptra/restapi/v1/jobs/service_to_migrate.py
+++ b/src/dioptra/restapi/v1/jobs/service_to_migrate.py
@@ -1,0 +1,1308 @@
+# This Software (Dioptra) is being made available as a public service by the
+# National Institute of Standards and Technology (NIST), an Agency of the United
+# States Department of Commerce. This software was developed in part by employees of
+# NIST and in part by NIST contractors. Copyright in portions of this software that
+# were developed by NIST contractors has been licensed or assigned to NIST. Pursuant
+# to Title 17 United States Code Section 105, works of NIST employees are not
+# subject to copyright protection in the United States. However, NIST may hold
+# international copyright in software created by its employees and domestic
+# copyright (or licensing rights) in portions of software that were assigned or
+# licensed to NIST. To the extent that NIST holds copyright in this software, it is
+# being made available under the Creative Commons Attribution 4.0 International
+# license (CC BY 4.0). The disclaimers of the CC BY 4.0 license apply to all parts
+# of the software developed or licensed by NIST.
+#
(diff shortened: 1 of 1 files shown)


keithmanville (approved):
This works as expected for me.

In the UI table, a NaN value shows up blank. This should probably have the text "NaN" so it is clear to the user. I'm guessing this is because the frontend gets None from the API since we only have the conversion in the python client.

In the graphs, NaNs show as gaps in the line plot, which seems like a reasonable visualization to me.
image

@henrychoy

jkglasbrenner:
Rebased this on `dev` as best I could. There's been a significant amount of drift in `src/dioptra/restapi/v1/jobs/service.py` due to the artifacts handler merge, so this isn't ready to merge. The biggest difference is that we now have `src/dioptra/restapi/v1/shared/job_run_store.py`, so most of the changes I see in the `service.py` file belong here.

What I've done is put a copy of the pre-rebased version of the `service.py` file at `src/dioptra/restapi/v1/jobs/service_to_migrate.py`. I resolved the minor linting and formatting errors that came up, and incorporated the error class rename `MLFlowError -> JobStoreError`. So here's what we have:

- `src/dioptra/restapi/v1/jobs/service.py` - This file was reset to the version currently on dev as of commit 73b02418eb413b95501359b4ed5b1c89ee77de66.
- `src/dioptra/restapi/v1/jobs/service_to_migrate.py` - This is a copy of `service.py` from commit 413b389759b31ea1b6ee104249e482eebb6c10f1 in the `jtsexton-metrics-nan-to-none` branch (pre-rebase and force push). Minor linting, formatting, and name changes have been incorporated, but otherwise it is unchanged. 

## TODO

The spirit of the changes in `service_to_migrate.py` need to be migrated into `src/dioptra/restapi/v1/shared/job_run_store.py` and `src/dioptra/restapi/v1/jobs/service.py` in order to bring things up to date.

jkglasbrenner:
Unrelated to the TODO in my first message, I also want to revisit the `metrics_wrapper` in `src/dioptra/client/jobs.py`. Having been away from this for a couple months, it's hard to understand how it's working at a glance. More generally, this approach feels like a quick bandaid rather than a proper fix. I want to think about this again, it would be nice if we could handle this kind of issue in a more integrated way.

keithmanville:
superseded by #946

PR #835, opened 2025-04-17
Add validateEntrypoint workflow

### LINKED ISSUE #615: Add workflow for entrypoint yaml validation
Add a new workflow for validating the task_graph portion of the entrypoint yaml

This should make use of the existing validation functions here:
https://github.com/usnistgov/dioptra/blob/main/src/dioptra/task_engine/validation.py

The endpoint should look something like: `POST /workflows/entrypointValidate { "task_graph": "contents", "plugin_ids" [...], "entrypoint_parameters": { .. } }`
With the response being a message stating the either the yaml is valid or not (with the reason provided by the validation function).

We want to be able to validate an entrypoint yaml before an Entrypoint Resource has been created by the user. To accomplish this the user will need to provide some details of the entrypoint.

See here for an example of constructing the full task engine yaml:
https://github.com/usnistgov/dioptra/blob/03d295ea1d433bf67e6dc89f0fbd9e006abed9de/src/dioptra/restapi/v1/workflows/lib/export_task_engine_yaml.py#L67

Additionally, the entrypoint services should be updated to prevent the creation of snapshots with invalid yaml. They should return an appropriate error message containing the validation errors from the validation function.

### Definition of Done:
- [x] The schema, controller, service, and errors are added to implement the yaml validation workflow
- [ ] task graph yaml is integrated into the entrypoint services to prevent invalid yaml from being commited in a snapshot
- [ ] New tests are added to cover this feature
- [ ] All tests pass
- [ ] Feature is merged into dev

jkglasbrenner:
Replaces PR #673

Closes #615

### DIFF
--- a/src/dioptra/client/workflows.py
+++ b/src/dioptra/client/workflows.py
@@ -25,6 +25,7 @@ from .base import (
 
 T = TypeVar("T")
 
+VALIDATE_ENTRYPOINT: Final[str] = "validateEntrypoint"
 JOB_FILES_DOWNLOAD: Final[str] = "jobFilesDownload"
 SIGNATURE_ANALYSIS: Final[str] = "pluginTaskSignatureAnalysis"
 RESOURCE_IMPORT: Final[str] = "resourceImport"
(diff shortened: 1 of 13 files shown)


jkglasbrenner:
In the interest of getting this into dev and unblocking other PRs, I'm merging this. Tests all pass, including the new ones.

PR #838, opened 2025-04-22
Local MLFlow Guidance and Script. Fixes #815

### LINKED ISSUE #815: Add information about setting up MLflow to developer KB
Need instructions on local set-up (and potentially docker container set-up) for running MLFlow as part of a developer set-up/configuration

JustKuzya:
- Added auto-creation of the MLFlow directory to the script. 

Maybe the non-linear mlFlow URI part needs to be done as well inside the script (currently there is just warning in the MD-guide). On the other hand, it seems that the script's behavior would be less intuitive with the partial path auto-included into the URI.


JustKuzya:
Changed base that defaulted to main for some reason.

dev-kb/ml-flow/ml-flow-with-dioptra.md
```diff
@@ -0,0 +1,36 @@
+## [MLFlow](https://mlflow.org/docs/latest/) is an open source platform for Machine Learning(ML)
+This Knowledge Base document describes how to utilize MLFlow with Dioptra locally
+
+- To use the `start-mlflow.sh` script (after obtaining the script from the source-control), please remember to add the executable attribute to the script:
```
keithmanville, line 4: We should be able to commit the script with executable permissions, so this wouldn't be necessary.

keithmanville (commented):
I believe James is the original author of the mlflow script, so let's be sure to credit him properly in the commit log.

hbooth (approved):
after a few minor changes it looks good to go

PR #840, opened 2025-04-25
fix: client typing changes to accommodate JSON lists

### LINKED ISSUE #836: [BUG] The current JSON client accepts dict[str, Any] only, but should also accept list[dict[str, Any]]
Because endpoints like `/api/v1/jobs/{id}/metrics`, `/api/v1/entrypoints/{id}/queues`, and `/api/v1/entrypoints/{id}/tags` exist, the client's typing should be updated to be able to handle `list[dict[str, Any]]` types, instead of just `dict[str, Any]`. 

This can be done by adding an addition type variable T2, such that T1 = `dict[str, Any]` and T2 = `list[dict[str, Any]]` for the JSON client, and T1 == T2 for the Response client.

jtsextonMITRE:
Closes #836

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -431,3 +431,4 @@ node_modules/
 !tests/**/lib
 !src/**/tags
 !src/**/lib
+/examples/dioptra-venv
--- a/.gitignore
+++ b/.gitignore
@@ -431,4 +431,3 @@ node_modules/
 !tests/**/lib
 !src/**/tags
 !src/**/lib
-/examples/dioptra-venv
--- a/src/dioptra/client/artifacts.py
+++ b/src/dioptra/client/artifacts.py
@@ -19,10 +19,11 @@ from typing import Any, ClassVar, TypeVar
 from .base import CollectionClient, DioptraSession
 from .snapshots import SnapshotsSubCollectionClient
 
-T = TypeVar("T")
+T1 = TypeVar("T1")
+T2 = TypeVar("T2")
 
 
-class ArtifactsCollectionClient(CollectionClient[T]):
+class ArtifactsCollectionClient(CollectionClient[T1, T2]):
     """The client for managing Dioptra's /artifacts collection.
 
     Attributes:
@@ -31,19 +32,19 @@ class ArtifactsCollectionClient(CollectionClient[T]):
 
     name: ClassVar[str] = "artifacts"
 
-    def __init__(self, session: DioptraSession[T]) -> None:
+    def __init__(self, session: DioptraSession[T1, T2]) -> None:
         """Initialize the ArtifactsCollectionClient instance.
 
         Args:
             session: The Dioptra API session object.
         """
         super().__init__(session)
-        self._snapshots = SnapshotsSubCollectionClient[T](
+        self._snapshots = SnapshotsSubCollectionClient[T1, T2](
             session=session, root_collection=self
         )
 
     @property
-    def snapshots(self) -> SnapshotsSubCollectionClient[T]:
+    def snapshots(self) -> SnapshotsSubCollectionClient[T1, T2]:
         """The client for retrieving artifact resource snapshots.
 
         Each client method in the sub-collection accepts an arbitrary number of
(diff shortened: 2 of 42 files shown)


jkglasbrenner:
I finally got around to revisiting the approach proposed in this PR, and I think the concern I raised in August still applies. The general problem is that adding more generic type slots to the Dioptra client does not scale well. If we knew that the client would only ever return either a dictionary or a list of dictionaries, then it would be fine, but because JSON can also produce scalar values, arrays of scalars, nested arrays, and null, trying to model that with more shared generic parameters becomes untenable.

Instead, a better direction may be to separate the implementation type annotations from the public typing surface using `.pyi` stubs. In our client source code, we would update the JSON client annotations to return something like `JsonValue`, which provides the full range of types that `json.loads` can return rather than pretending every JSON response is `dict[str, Any]`. That keeps the runtime implementation simple and avoids multiplying generic slots. Then, we can use generated `.pyi` stubs to provide the more precise public interface that we want. The stubs can describe `connect_response_dioptra_client()` as returning a response-client protocol and `connect_json_dioptra_client()` as returning a JSON-client protocol. Those protocols can give endpoint-specific return types without duplicating the implementation bodies.

Before we commit to this, I think it is worth doing a small spike off the current state of our `dev` branch. The purpose of the spike is to prove that a stub for at least one collection client, say queues, is picked up by mypy, that VS Code or Pylance follows the stubbed factory return type, and that the runtime source can remain a single implementation. If that works, then we can figure out how we want to handle the rest of the client codebase.

One idea I have to make this maintainable is to generate the `.pyi` stubs from the source code, with a way to configure exceptions to the JSON output shape. Most of those return types are going to be dictionaries, and only a few are going to be exceptions. If we can specify those exceptions, override them, and otherwise auto-generate the `.pyi` stubs from the source code signatures, then this becomes much more maintainable.

So the plan would be to first prove that the stubs work at all. If they do, then we can come up with a method for generating them so this does not become a massive maintenance burden, because this client has quite a large surface area.

jkglasbrenner:
I ran the proposed `.pyi` stub spike from a fresh branch off current `dev`, and the result looks promising.

The question I wanted to answer was whether we can keep the Python client runtime simple while giving users more precise static types for the JSON-returning client. In other words, the runtime source can honestly say that JSON responses return a broad `JsonValue`, while `.pyi` stubs describe the more specific public interface that users and language servers should see for each endpoint.

For the spike, I kept the runtime implementation as a single client. I added source-level JSON aliases, including `JsonValue`, and updated the JSON session and JSON factory source annotations to use that broad JSON type. Then I added a minimal stub surface for the factory functions and one client path, `queues`, including the `queues.tags` subcollection. That gave us two useful cases to test:

- queue endpoints that return object-shaped JSON
- queue tag endpoints that return a top-level JSON list

Both mypy and Pyright picked up the stubbed factory return types. In the typing probe, the JSON factory was treated as returning a JSON-client protocol, queue methods such as `queues.get_by_id()` were treated as returning object-shaped JSON, and queue tag methods such as `queues.tags.get(1)` were treated as returning a list of JSON objects. The response-object factory was also treated separately, with queue methods still returning `DioptraResponseProtocol`.

I also included a negative assignment in the probe to make sure the result was not just collapsing to `Any` or one generic JSON type. That check worked, so the object-shaped and list-shaped JSON responses are genuinely distinct to the type checker.

The runtime smoke test passed. Both factories still instantiate the existing runtime `DioptraClient`, and `client.queues` is still the existing `QueuesCollectionClient`. No duplicate collection-client method bodies were introduced.

Packaging also looks workable. A wheel build included the new stubs and the typed-package marker:

- `dioptra/client/client.pyi`
- `dioptra/client/__init__.pyi`
- `dioptra/py.typed`

Commands that passed:

```bash
uv run mypy src/dioptra/client/base.py src/dioptra/client/sessions.py src/dioptra/client/client.py
uvx tox run -e mypy -- src/dioptra/client tests/typing/client_json_stub_probe.py
uvx pyright tests/typing/client_json_stub_probe.py
uv build --wheel
```

My read is that the basic premise is viable. The next design question is how to generate the stubs from the runtime source plus a small endpoint-shape catalog, because hand-maintaining the full client surface would be too much. One caveat from the spike is that colocated `.pyi` files become the type checker’s public view of a module, so we should keep implementation-source checks and consumer-surface checks distinct in the workflow.

I pushed the code for this spike to this branch: <https://github.com/usnistgov/dioptra/tree/client-json-typing-stub-spike>.

Issue #841, opened 2025-05-05
add worker user registration to deployment config


etrapnell-nist:

jkglasbrenner:
Closed by #842

PR #842, opened 2025-05-05
add worker user registration to deployment config

etrapnell-nist:
Automatically registers the dioptra-worker account as proposed in #841


jkglasbrenner (approved):
LGTM

Issue #816, opened 2025-04-04
document/script how to setup/use TMUX within developer KB


hbooth:
Provide instructions/scripts for the use of TMUXS within developer environment

hbooth:
closed by PR #843

PR #843, opened 2025-05-12
Jk tmux start i816

JustKuzya:
Creates integrated launch-script leveraging all existing scripts for starting particular servers and services for dioptra dev-environment. Fixes issue #816


JustKuzya:
Overwritten the last commit message, added more doc-comments to the script, and resolved dev ↔ jk-tmux-start-i816  merge conflicts in ml-flow directory. Namely the changes added in this conflict resolve: readme typo, extra informational logic in the script.

mtrapnell-nist:
Everything works for me. In reviewing the scripts, I noted the @Q syntax won't work on earlier versions of bash. Do we want to consider changing this or just have the requirement be that bash 5 is installed?

etrapnell-nist:
I was able to get this running with one caveat: I had to specify this branch name in the config file, and then manually call `npm run dev` in one of the tmux windows. Since the `run-front.sh` script uses the `${DIOPTRA_BRANCH}` variable, it will fail on arbitrary branch names. Perhaps add a new config variable to define the npm workspace and decouple it from the branch name?

JustKuzya:
> Everything works for me. In reviewing the scripts, I noted the `@Q` syntax won't work on earlier versions of bash. Do we want to consider changing this or just have the requirement be that bash 5 is installed?

Added the bash version 5+ check logic into start-all.sh

hbooth (approved):
reviewed and tested script files

PR #856, opened 2025-05-21
feat: add restapi job logging endpoint

### LINKED ISSUE #773: Add logging endpoint to Jobs
Use the prototype in #467 to configure logging to send to and endpoint.

Design a new endpoint under Jobs to accept logs. Logs should be able to be appended. Will need to store logs in some way. There should be a new GET endpoint for retrieving logs.

In a separate issue, we should make the logs visible in the frontend.

chisholm:
Fixes #773 

Unit tests are pretty simple and minimal so far.  Need to approve/finalize the design first.

### DIFF
--- a/src/dioptra/restapi/db/models/jobs.py
+++ b/src/dioptra/restapi/db/models/jobs.py
@@ -19,7 +19,15 @@ from typing import TYPE_CHECKING
 from sqlalchemy import Column, ForeignKey, ForeignKeyConstraint, Index, Text, select
 from sqlalchemy.orm import Mapped, column_property, mapped_column, relationship
 
-from dioptra.restapi.db.db import bigint, db, guid, intpk, text_
+from dioptra.restapi.db.db import (
+    bigint,
+    datetimetz,
+    db,
+    guid,
+    intpk,
+    optionalstr,
+    text_,
+)
 
 from .entry_points import EntryPoint
 from .experiments import Experiment
@@ -35,6 +43,11 @@ if TYPE_CHECKING:
 job_status_types_table = db.Table(
     "job_status_types",
     Column("status", Text(), primary_key=True),
+)
+
+job_log_severity_table = db.Table(
+    "job_log_severity",
+    Column("severity", Text(), primary_key=True),
 )
 
 # -- ORM Classes -----------------------------------------------------------------------
@@ -210,3 +223,22 @@ class Job(ResourceSnapshot):
     __mapper_args__ = {
         "polymorphic_identity": "job",
     }
+
+
+class JobLog(db.Model):  # type: ignore[name-defined]
+    __tablename__ = "job_logs"
+
+    # Database fields
+
+    # "id" also acts as an ordinal for recording log entry order
+    id: Mapped[intpk] = mapped_column(init=False)
+    job_resource_id: Mapped[bigint] = mapped_column(
+        ForeignKey("resources.resource_id"), init=False
+    )
+    severity: Mapped[text_] = mapped_column(ForeignKey("job_log_severity.severity"))
+    step_name: Mapped[optionalstr]
+    timestamp: Mapped[datetimetz]
+    message: Mapped[text_]
+
+    # Relationships
+    job_resource: Mapped["Resource"] = relationship()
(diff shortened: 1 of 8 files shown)


jkglasbrenner (changes requested):
The database table design for capturing the log messages for a Job looks good, I had a comment regarding adding an additional relationship to it.

The other comments are mostly about building a paging envelope to match what is done in other endpoints.

Other than the above, the general approach here is what I'd expect to see.

src/dioptra/restapi/db/models/jobs.py
```diff
@@ -210,3 +223,22 @@ class Job(ResourceSnapshot):
...
+    job_resource_id: Mapped[bigint] = mapped_column(
+        ForeignKey("resources.resource_id"), init=False
+    )
+    severity: Mapped[text_] = mapped_column(ForeignKey("job_log_severity.severity"))
+    step_name: Mapped[optionalstr]
+    timestamp: Mapped[datetimetz]
+    message: Mapped[text_]
+
+    # Relationships
+    job_resource: Mapped["Resource"] = relationship()
```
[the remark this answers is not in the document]
chisholm: I couldn't find a good way to do this.  You could add a plain python `@property` which triggers a query, but sqlalchemy docs say every access of the property would trigger another query.  Seems kind of simplistic and inefficient.  So I'll just leave the classes as-is and resolve this for now.

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -1307,3 +1310,86 @@ def get(
...
+            )
+            db.session.add(job_log)
+
+        db.session.commit()
+
+    def get_logs(
+        self,
+        job_resource_id: int,
+        index: int | None,
+        page_length: int | None,
```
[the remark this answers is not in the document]
chisholm: This should be done, but isn't specific to job logs.  All paging should have limits.  Maybe we should open a separate issue to address paging limits in all places paging is supported?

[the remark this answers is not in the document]
chisholm: I added #872.  So I will close this.

src/dioptra/restapi/v1/jobs/schema.py
```diff
@@ -262,3 +264,86 @@ class ExperimentJobGetQueryParameters(
...
+    A logging record within the context of a running job.
+    """
+
+    step_name = fields.String(
+        data_key="stepName",
+        metadata={
+            "description": "The name of the step which was executing when the"
+            " log record was emitted, if known/applicable."
+        },
+    )
```
jkglasbrenner: The snake_case and camelCase versions are reversed, and should use `attribute` instead of `data_key` in this situation:

```suggestion
    stepName = fields.String(
        attribute="step_name",
        metadata={
            "description": "The name of the step which was executing when the"
            " log record was emitted, if known/applicable."
        },
    )
```

src/dioptra/restapi/v1/jobs/schema.py
```diff
@@ -262,3 +264,86 @@ class ExperimentJobGetQueryParameters(
...
+    severity = fields.Enum(
+        JobLogSeverity,
+        required=True,
+        metadata={
+            "description": "Log severity level: {}".format(
+                ", ".join(e.name for e in JobLogSeverity)
+            )
+        },
+    )
```
jkglasbrenner, line 343: It seems like the Enum field type doesn't show up in the generated Swagger Docs:

image

Enum schema fields do seem to work on POST endpoints that are forms (see `/workflows/jobFilesDownload`), but not with JSON-based POST endpoints. Strange. It clearly works if you make the POST, so validation is fine, so its only an issue with the Swagger Docs.

It's not a dealbreaker, but if you're able to figure out what's going on in flask-accepts that is causing this issue, then maybe we can figure out a workaround.

src/dioptra/restapi/v1/jobs/schema.py
chisholm: Oh, I wrote it according to advice in the marshmallow docs:

https://marshmallow.readthedocs.io/en/latest/marshmallow.fields.html#marshmallow.fields.Field

For `attribute`:

> Note: This should only be used for very specific use cases such as outputting multiple fields for a single attribute, or using keys/attributes that are invalid variable names, unsuitable for field names. In most cases, you should use data_key instead.

So I used `data_key`.

jkglasbrenner (approved):
Everything continued to work as expected after rebasing on `dev` and integrating with the latest changes. I updated the parent hash for the alembic script and added an index to the job logs table. Going to merge to dev shortly.

src/dioptra/restapi/v1/jobs/schema.py
jkglasbrenner: Spun this off into a separate issue card.

PR #861, opened 2025-05-22
feat: Validate entrypoint button

### LINKED ISSUE #857: Add UI elements to allow user to validate their entrypoint inputs (both for new and existing resources)
Add a button to the entrypoint creation/edit page for doing an entrypoint validation based on the current input. The response form the endpoint will either be a success message without any errors listed, or a failure with one or more errors listed. The errors, if present, should be shown in a way that's more prominent and easy to review.

See changes in #835 for info on what the requests and responses will look like.

henrychoy:
closes #857 

Add validate inputs button to entrypoint create/edit page.  List errors in a popup dialog if there are any.

### DIFF
--- a/src/dioptra/restapi/errors.py
+++ b/src/dioptra/restapi/errors.py
@@ -347,6 +347,13 @@ class PluginParameterTypeMatchesBuiltinTypeError(DioptraError):
         )
 
 
+class InvalidYamlError(DioptraError):
+    """Raised when the provided YAML is invalid and fails to parse."""
+
+    def __init__(self, message: str):
+        super().__init__(message)
+
+
 # User Errors
 class UserDoesNotExistError(DioptraError):
     """The entered username does not exist."""
--- a/src/dioptra/restapi/v1/shared/task_engine_yaml/service.py
+++ b/src/dioptra/restapi/v1/shared/task_engine_yaml/service.py
@@ -21,7 +21,10 @@ from typing import Any, Final, cast
 import structlog
 import yaml
 from structlog.stdlib import BoundLogger
+from yaml.parser import ParserError
+from yaml.scanner import ScannerError
 
+from dioptra.restapi.errors import InvalidYamlError
 from dioptra.restapi.v1.workflows.lib.type_coercions import (
     BOOLEAN_PARAM_TYPE,
     FLOAT_PARAM_TYPE,
@@ -249,7 +252,10 @@ class TaskEngineYamlService(object):
             A dictionary representation of the entry point's task graph.
         """
         log = logger or LOGGER.new()  # noqa: F841
-        return cast(dict[str, Any], yaml.safe_load(entry_point.task_graph))
+        try:
+            return cast(dict[str, Any], yaml.safe_load(entry_point.task_graph))
+        except (ParserError, ScannerError) as e:
+            raise InvalidYamlError(str(e)) from e
 
     def build_plugin_field(
         self,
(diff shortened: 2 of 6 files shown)


henrychoy:
I noticed the 2 issues below.  Wondering if I should handle both from the frontend or better to modify the endpoint?

- The validate endpoint returns a 500 when submitting invalid YAML
- The validate endpoint doesn't consider our placeholders <step-name> and <input-value> as errors, is this ok?  Or should I add them as errors in the dialog from the frontend?

jkglasbrenner:
The error handlers are all defined here: https://github.com/usnistgov/dioptra/blob/dev/src/dioptra/restapi/errors.py

What we want is to create a DioptraError (should be a new one) that will be raised when the YAML fails to parse. Then, the caller can decide whether the error can just bubble up like usual, or in the validation endpoint case, catch it and repackage the message to match the other validation responses.

henrychoy:
> The error handlers are all defined here: https://github.com/usnistgov/dioptra/blob/dev/src/dioptra/restapi/errors.py
> 
> What we want is to create a DioptraError (should be a new one) that will be raised when the YAML fails to parse. Then, the caller can decide whether the error can just bubble up like usual, or in the validation endpoint case, catch it and repackage the message to match the other validation responses.

Thanks, just created the new error below and used it instead of ValueError.

```
class InvalidYamlError(DioptraError):
    """Raised when the provided YAML is invalid and fails to parse."""

    def __init__(self, message: str):
        super().__init__(message)
```

jkglasbrenner (changes requested):
@henrychoy The validate button appears to work as expected when editing an existing entrypoint, but it is failing (at least for me) when I try to use it when creating a new entrypoint. Here's the setup I was validating:

image

When I click validate, nothing shows up in the UI. When I check the console, I get this AxiosError:

```json
{
    "message": "Request failed with status code 400",
    "name": "AxiosError",
    "stack": "AxiosError: Request failed with status code 400\n    at settle (http://localhost:5173/node_modules/.vite/deps/axios.js?v=7e0c4121:1218:12)\n    at XMLHttpRequest.onloadend (http://localhost:5173/node_modules/.vite/deps/axios.js?v=7e0c4121:1550:7)\n    at Axios.request (http://localhost:5173/node_modules/.vite/deps/axios.js?v=7e0c4121:2108:41)\n    at async Module.validateEntrypoint (http://localhost:5173/src/services/dataApi.ts:218:10)\n    at async Proxy.validateInputs (http://localhost:5173/src/views/CreateEntryPoint.vue:410:19)",
    "config": {
        "transitional": {
            "silentJSONParsing": true,
            "forcedJSONParsing": true,
            "clarifyTimeoutError": false
        },
        "adapter": [
            "xhr",
            "http",
            "fetch"
        ],
        "transformRequest": [
            null
        ],
        "transformResponse": [
            null
        ],
        "timeout": 0,
        "xsrfCookieName": "XSRF-TOKEN",
        "xsrfHeaderName": "X-XSRF-TOKEN",
        "maxContentLength": -1,
        "maxBodyLength": -1,
        "env": {},
        "headers": {
            "Accept": "application/json, text/plain, */*",
            "Content-Type": "application/json"
        },
        "method": "post",
        "url": "/api/v1/workflows/validateEntrypoint",
        "data": "{\"taskGraph\":\"hello_step:\\n  hello:\\n    name: $name\",\"pluginSnapshots\":[null],\"parameters\":[]}",
        "allowAbsoluteUrls": true
    },
    "code": "ERR_BAD_REQUEST",
    "status": 400
}
```

The response is:

```json
{"errors": {"group": ["Missing data for required field."], "pluginSnapshots": {"0": ["Field may not be null."]}}}
```

My assumption is that the plugin snapshot ids aren't being pulled into the request. Can you check and update?

henrychoy:
@jkglasbrenner Thanks for catching that.  You're right it was the group and pluginSnapshots not being passed correctly.  Just pushed up a fix.

PR #874, opened 2025-06-14
fix: make a task engine job stoppable (rebased version)

chisholm:
This a kind of "rebase" of #444 (see that PR for technical details), but not simply a `git rebase` since dev branch has changed so much since the original changes were made.  I just made the changes by hand, based on the older branch.  run_task_engine_stoppable.py is brought forward unchanged, and the same updates were applied to task_engine.py to add the stop flag.  But the integration into the job execution procedure is changed to work with current dev branch.

Calls to run_experiment() were replaced with run_experiment_stoppable(), but the change was made in run_dioptra_job.py.tmpl.  Its main() function now propagates the stopped flag back to run_v1_dioptra_job.py, where the dioptra job status is set.

Normal job execution was tested with the hello world example.  Premature stopping was not tested since we still don't have a mechanism to signal a job to stop (I did test this in the original branch, but had to make some awkward contrivances to do so).

### DIFF
--- a/src/dioptra/restapi/v1/workflows/lib/run_dioptra_job.py.tmpl
+++ b/src/dioptra/restapi/v1/workflows/lib/run_dioptra_job.py.tmpl
@@ -28,6 +28,8 @@ from dioptra.sdk.utilities.contexts import env_vars, sys_path_dirs
 from dioptra.task_engine.issues import IssueSeverity
 from dioptra.task_engine.task_engine import run_experiment
 from dioptra.task_engine.validation import validate
+from dioptra.rq.tasks.run_task_engine_stoppable import run_experiment_stoppable
+
 
 LOGGER: BoundLogger = structlog.stdlib.get_logger()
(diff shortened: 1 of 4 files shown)


jkglasbrenner:
@hbooth Would be helpful to get your feedback on this regarding the following questions:

1. Based on my understanding of your incoming changes for the artifact handler to resolve the initial work in #658, the changes to `src/dioptra/restapi/v1/workflows/lib/run_dioptra_job.py.tmpl` in this PR might rendered obsolete. Should this PR omit the changes to the `.tmpl` file entirely? (We can capture the proposed changes in a comment here so that it's not lost).
2. The changes in `src/dioptra/rq/tasks/run_v1_dioptra_job.py` may also interfere with the artifact handlers changes. Also, as the code comments left here suggest, it may be better to add additional statuses to capture the difference between a stopped/killed job and a failed job. Should we remove these changes from the PR? (and capture in a comment, like suggested in 1?)

PR #879, opened 2025-06-18
feat: add import plugin tasks button

### LINKED ISSUE #626: Incorporate plugin param type inference into UI
The UI should make use of the plugin task inference workflow to assist the user in defining plugin tasks.

Add a button to trigger the type inference when composing a Plugin task. It would send the Python file contents into the appropriate workflow endpoint when clicked, and then import the response back into the GUI.

### Blocked by:
- #618

### Definition of Done:
- [ ] The plugin file creation and editing forms uses the workflow to infer task definitions

henrychoy:
Closes #626 

This PR adds a "Import Plugin Tasks" button on the plugin file form.

- Uses the PluginTaskSignatureAnaylsis endpoint to present a list of plugin tasks for the user to import.
- The endpoint was returning a 500 when invalid python was provided.  Changed it to raise a InvalidPythonError instead.
- When importing a task that already exists, the task will be highlighted and the user is shown what params will be overwritten.  The user can then decide whether to import that task or not.
- Params with missing types will be flagged on the parent create plugin file form page.

### DIFF
--- a/src/dioptra/restapi/errors.py
+++ b/src/dioptra/restapi/errors.py
@@ -396,6 +396,13 @@ class InvalidYamlError(DioptraError):
         super().__init__(message)
 
 
+class InvalidPythonError(DioptraError):
+    """Raised when the provided python code is invalid and fails to parse."""
+
+    def __init__(self, message: str):
+        super().__init__(message)
+
+
 # User Errors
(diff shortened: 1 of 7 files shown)


keithmanville (commented):
approved after fixing linting checks and squashing commits.

PR #880, opened 2025-06-19
fix: add more get_one() methods, ensure they raise the right exceptions

chisholm:
This PR adds get_one() methods to user, group, draft repositories. Exception behavior is updated across all implementations to ensure the correct exceptions are thrown, depending on existence and deletion policy.  Unit tests are added/updated.

These changes are necessary to address #877 .

(Also fixed a minor docstring bug in the drafts repo.)

PR #882, opened 2025-06-19
refactor: split repo utils into smaller parts

### LINKED ISSUE #875: Break up the repository utils file into a module with smaller sub-units
The existing utils file at: src/dioptra/restapi/db/repository/utils.py is quite large and would benefit from being divided into smaller sub-units that are closely related. Externally, the module will look the same with functions/classes exported through __init__.py.

chisholm:
Fixes #875 .

This PR shows a finer grained organization to repo utils:
- common: things used across repos, repo utils, services, etc.  These are very general, e.g. DeletionPolicy.
- resources: commonalities specific to the resource repositories, e.g. adding/removing child resources
- checks: this has the assert and non-assert checker functions (I didn't want to call it "asserts" since it has more types of functions than that).
- search: this has some functionality common to all the field-based and fuzzy keyword searches

All symbols are exported to the parent utils module.

(Also added the standard license comment to repository/\_\_init\_\_.py, which I'd forgotten before.)

This will conflict with PR #880, since it also modifies repo utils.  If that PR is merged first, I'll have to update and rebase this one (and vice versa).


jkglasbrenner:
@chisholm As a heads up I merged PR #880, so this one will need to be rebased.

chisholm:
Rebase is done.

PR #884, opened 2025-06-23
fix(restapi): allow name conflicts with deleted resources/snapshots

keithmanville:
This commit allows for new resources and resource snapshots to be created with the name of a previously deleted resource. It applies to resources that are using the new repository pattern implementations, reverting the behavior back to how name conflicts were previously handled.

It fixes the issue where if a resource was deleted, that resource's name could never be used again. This issue was especially prevalent when using the resource import workflow to overwrite (delete and re-create) existing resources.

It updates a resource import workflow test by first creating resources with conflicting names before importing resources with the overwrite strategy.

### DIFF
--- a/src/dioptra/restapi/db/repository/utils/checks.py
+++ b/src/dioptra/restapi/db/repository/utils/checks.py
@@ -1227,6 +1227,7 @@ def assert_resource_name_available(
             resource_type_class.resource_snapshot_id == m.Resource.latest_snapshot_id,
             # Dunno why mypy has trouble with this expression...
             m.Resource.owner == snap.resource.owner,  # type: ignore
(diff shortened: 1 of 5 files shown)


chisholm (changes requested):
I think the changes are fine.  It makes sense to me to not add deletion policy to the assert_*_name_exists functions, until there is a use case.

But I think repo utils unit tests should probably be added, which include deletion.  The tests which are there don't delete anything; it probably never occurred to me to make behavior depend on deletion state, so that variability is not in the tests.  It sounds like an important thing to check.

https://github.com/usnistgov/dioptra/blob/6d757771170f11456e1a28f2aff9623d0ac234c9/tests/unit/restapi/test_repo_utils.py#L1215-L1239

PR #886, opened 2025-06-25
refactor: change .get() calls to .get_one() in services

### LINKED ISSUE #877: Any call to retrieve a resource that can use get_one should do so
To simplify the client code, any client that is able to use the get_one method should favor that method where a "None" result would be an error.

chisholm:
Fixes #877 .

... where possible.  Also remove some mypy type guard asserts in queue services which are no longer necessary. Changing .get() method signature to an `@overload` style helped mypy understand it better, so the type guards became superfluous.


jkglasbrenner (approved):
Looks good to me.

PR #888, opened 2025-06-25
feat: default sort by job id descending

### LINKED ISSUE #870: Add ability to sort jobs on the jobs page by ID and change the default to descending
The current default for viewing jobs is to sort by ascending job ID, which ends up making it so that the first page always displays the oldest jobs. Most people using Dioptra will want to see the newest jobs first, so we should reverse this, and probably also make more of the fields on this page sortable. 

Currently you can sort by only description and status, but it would be useful to be able to sort by Job ID, Experiment, Entrypoint, and Queue as well.

henrychoy:
closes #870 

- sort by job id descending by default so that newest jobs are listed first
- add ability to sort by job experiment, entrypoint, and queue

### DIFF
--- a/src/dioptra/restapi/v1/jobs/service.py
+++ b/src/dioptra/restapi/v1/jobs/service.py
@@ -69,10 +69,14 @@ SEARCHABLE_FIELDS: Final[dict[str, Any]] = {
     "tag": lambda x: models.Job.tags.any(models.Tag.name.like(x, escape="/")),
 }
 SORTABLE_FIELDS: Final[dict[str, Any]] = {
+    "id": models.Resource.resource_id,
     "description": models.Job.description,
     "createdOn": models.Job.created_on,
     "lastModifiedOn": models.Resource.last_modified_on,
     "status": models.Job.status,
+    "experiment": models.Experiment.name,
+    "entrypoint": models.EntryPoint.name,
+    "queue": models.Queue.name,
 }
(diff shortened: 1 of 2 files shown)


keithmanville:
Recommend moving description after Queue in the column ordering so it would be:
JobID, Experiment, EntryPoint, Queue, Description, ...

Description is most likely to have varying length text. When you change the sort order and longer descriptions appear, the columns re-size and all all columns end up shifted to the right, which can be a little disorienting. I think moving the Description column as recommended would help alleviate this. Open to other ideas.

keithmanville (commented):
Recommend moving description after Queue in the column ordering so it would be:
JobID, Experiment, EntryPoint, Queue, Description, ...

Description is most likely to have varying length text. When you change the sort order and longer descriptions appear, the columns re-size and all all columns end up shifted to the right, which can be a little disorienting. I think moving the Description column as recommended would help alleviate this. Open to other ideas.

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -354,6 +358,12 @@ def get(
         jobs_stmt = (
             select(models.Job)
             .join(models.Resource)
+            .join(models.Job.experiment_job)
```
keithmanville, line 361: @jkglasbrenner These joins are needed to support sorting by Experiment/EntryPoint/Queue. Is there a better way to do this? Anything problematic with joining so many tables here?

PR #892, opened 2025-06-30
fix: avoid error if last character of search is colon or comma

### LINKED ISSUE #859: Add de-bounce to table search box in UI
Currently, the UI is sending the search string immediatley to the API as its being typed, and it returns an error message when using the query language when you type a `:`, e.g., `name:`. There should be a brief delay before sending the query, after the user has not registered input for a set amount of time, for example 500ms.

henrychoy:
Closes #859 

There already is a 300ms debounce on the search field.  The issue is that when searching with a colon, it naturally takes longer and thus you see the error.  The same issue happens when using a comma.  The fix in this PR is to pass in an empty string whenever the last character of the search is a colon.  Similarly if the last character is a comma, it will strip out that trailing comma.  Now the user wont see an error until they finish typing their search.  I'm not sure this workaround is better because it gives the user a technically wrong result and assume they're gonna continue typing after the colon or comma, but let me know what you think.  Another idea is to have an "advanced search" button next to the search field which brings up a modal where you build these more advance searches through a form.

Other fixes in this PR:
- Remove unnecessary parsing done on the search string in the frontend.
- Currently the \ escape character can't be used to escape a colon or comma.  This adds those as escapable characters.
- In the docs, clarify that using commas searches by AND and not OR

Note: Currently experiments. queues, and param types return a 500 when the user inputs an invalid search string, unlike the other types that return a 422 with a message.  I'm assuming this is part of the ongoing migration to the repository pattern.

### DIFF
--- a/src/dioptra/restapi/db/repository/utils/search.py
+++ b/src/dioptra/restapi/db/repository/utils/search.py
@@ -50,6 +50,8 @@ def _construct_sql_search_value(search_term: str) -> str:
         search_term = search_term.replace(r"\\", "\\")
         search_term = search_term.replace(r"\*", "*")
         search_term = search_term.replace(r"\?", "?")
+        search_term = search_term.replace(r"\:", ":")
+        search_term = search_term.replace(r"\,", ",")
         search_term = search_term.replace(r"\"", '"')
         search_term = search_term.replace(r"\'", "'")
         search_term = search_term.replace(r"\n", "\n")
--- a/src/dioptra/restapi/v1/shared/search_parser.py
+++ b/src/dioptra/restapi/v1/shared/search_parser.py
@@ -44,7 +44,7 @@ def _define_query_grammar() -> pp.ParserElement:
     wildcard = ~pp.Literal("\\") + (pp.Literal("*") | pp.Literal("?"))
 
     # An escaped character is a '\' followed by a character that needs to be escaped
-    escape = pp.Literal("\\") + pp.Word("\\?*\"'n", exact=1)
+    escape = pp.Literal("\\") + pp.Word("\\?*\"'n:,", exact=1)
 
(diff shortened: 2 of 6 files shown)


src/dioptra/restapi/v1/shared/search_parser.py
```diff
@@ -44,7 +44,7 @@ def _define_query_grammar() -> pp.ParserElement:
     wildcard = ~pp.Literal("\\") + (pp.Literal("*") | pp.Literal("?"))
 
     # An escaped character is a '\' followed by a character that needs to be escaped
-    escape = pp.Literal("\\") + pp.Word("\\?*\"'n", exact=1)
+    escape = pp.Literal("\\") + pp.Word("\\?*\"'n:,", exact=1)
```
keithmanville, line 47: Is this change needed? My understanding was this change is only a frontend change that checks for `:` or `,` and holds the request to the backend.

src/dioptra/restapi/db/repository/utils/search.py
```diff
@@ -50,6 +50,8 @@ def _construct_sql_search_value(search_term: str) -> str:
         search_term = search_term.replace(r"\\", "\\")
         search_term = search_term.replace(r"\*", "*")
         search_term = search_term.replace(r"\?", "?")
+        search_term = search_term.replace(r"\:", ":")
+        search_term = search_term.replace(r"\,", ",")
```
keithmanville, line 54: Is this change needed? My understanding was this change is only a frontend change that checks for : or , and holds the request to the backend.

henrychoy: Yea I only made these changes because the docs had an example of being able to escape a colon character.  I can revert if that's not needed.

keithmanville (approved):
Approve after squashing commits.

Hold on merging to dev until we get #899 in

PR #894, opened 2025-07-01
fix: urlencode generated password

### LINKED ISSUE #867: [BUG] The random word generator used in the deployment can generate words with apostrophes
Yesterday I had the lucky experience of getting the word `mustn't` out of the random word generator which generates our .env file in the docker deployment. Unfortunately we also use those words in the URI for the postgresql database which can cause the container to be unable to start.

jtsextonMITRE:
Closes #867.


JustKuzya (approved):
The changes are exactly what was originally suggested in the comments for the issue. Good job us. 👍

Issue #575, opened 2024-07-29
Implement new API endpoints for browsing and retrieving artifact files


keithmanville:

jkglasbrenner:
Closed by #899

PR #899, opened 2025-07-07
Implementation of Artifact Handling Feature

### LINKED ISSUE #655: As a user I would like to be able to store output from a single task plugin or from a combination of task plugins as artifacts for later use and analysis.

hbooth:
Resolves #655 #575


src/dioptra/sdk/tasks/artifact.py
```diff
@@ -0,0 +1,271 @@
...
+    def deserialize(working_dir: Path, path: str, **kwargs) -> str:
+        return Path(working_dir, path).read_text()
+
+    @staticmethod
+    def validation() -> dict[str, Any] | None:
+        return {"type": {"enum": ["txt", "html", "md"]}}
+
+    @staticmethod
+    def name() -> str:
+        return "string"
```
keithmanville: Consider renaming some of these. I think it can be a little confusing when the name of the task is a primitive type. Maybe `string_artifact`?

src/dioptra/sdk/tasks/artifact.py
```diff
@@ -0,0 +1,271 @@
...
+        result.write_text(contents, encoding="UTF-8", newline="")
+        return result
+
+    @staticmethod
+    def deserialize(working_dir: Path, path: str, **kwargs) -> str:
+        return Path(working_dir, path).read_text()
+
+    @staticmethod
+    def validation() -> dict[str, Any] | None:
+        return {"type": {"enum": ["txt", "html", "md"]}}
```
keithmanville: Overly limiting? What purpose does the type serve if you are always getting in/out a string?

henrychoy:
With GET plugins, it doesn't provide the list of plugin.files.tasks[artifacts or functions][inputParams or outputParams].  I only get the id and name of each task.  So when filling out the entrypoint form, I cant populate the input and output param columns of the task table without making an additional call to get each pluginFile.  Can the input and output params also be included in the GET plugins?

src/dioptra/restapi/v1/plugins/schema.py
```diff
@@ -167,10 +209,10 @@ class PluginFileRefSchema(PluginFileRefBaseSchema):  # type: ignore
         metadata=dict(description="Filename of the PluginFile resource."),
     )
     tasks = fields.Nested(
```
keithmanville, line 211: in entrypoints we have `graph`/`artifact_graph` and `parameters`/`artifact_parameters`, but here we nest `functions` and `artifacts` under `tasks` instead of having `tasks`/`artifact_tasks`. So I think this is the only place we introduce the `function` terminology. Also the task schemas are different for both. Does it make sense to make `tasks` and `artifact_tasks` two top level fields instead?

keithmanville: Did we stop using the "artifact handler" terminology? I thought that was a nice way of differentiating between tasks and artifact serialization / deserialization.

src/dioptra/sdk/tasks/artifact.py
hbooth: when i first put together, it was mostly an example. i am happy to relax, and remove the restriction.

src/dioptra/restapi/v1/plugins/schema.py
hbooth: yes, handler was switched to "task" to better align the naming with the data structures where "function tasks" and "artifact tasks" both exist within a "plugin file" and we did not want to fully replicate the plugin file hierarchy for "artifact tasks" (or handlers). using the design that "plugin files" contain "tasks" and there can be different kinds of tasks allowed for the sharing of much of the code and machinery that was already there. currently "function" and "artifact" are two different kinds of tasks and allows for easier future extension to other kinds of "tasks" as needed.

src/dioptra/client/artifacts.py
```diff
@@ -14,10 +14,17 @@
 #
 # ACCESS THE FULL CC BY 4.0 LICENSE HERE:
 # https://creativecommons.org/licenses/by/4.0/legalcode
-from typing import Any, ClassVar, TypeVar
+from __future__ import annotations
```
jkglasbrenner: I've noticed the `from __future__ import annotations` has been restored in a number of files. The main reason we had it in the code before was because it provided support for the new style of type annotations in Python 3.9, e.g. `str | None` instead of `Optional[str]`. The new style of annotations work natively on Python 3.10 and above, and we dropped support for 3.9 a while ago.

It's worth pointing out that this `__future__` import still changes behavior in Python 3.10 and above, so it's not a no-op. It can cause unexpected bugs with libraries that use type annotations to determine runtime behavior. It's unclear if this behavior will ever be merged into main Python because of this, see https://docs.python.org/3/library/__future__.html#id1, it's already been delayed twice. My understanding is it breaks popular libraries like pydantic and FastAPI, and for a time it caused issues for injector.

Thoughts?

src/dioptra/client/artifacts.py
```diff
@@ -38,12 +45,12 @@ def __init__(self, session: DioptraSession[T]) -> None:
...
         """
         super().__init__(session)
-        self._snapshots = SnapshotsSubCollectionClient[T](
+        self._snapshots = ArtifactsSnapshotCollectionClient[T](
             session=session, root_collection=self
         )
 
     @property
-    def snapshots(self) -> SnapshotsSubCollectionClient[T]:
+    def snapshots(self) -> ArtifactsSnapshotCollectionClient[T]:
```
jkglasbrenner: Maybe this is why you added `from __future__ import annotations`? If so, the solution that removes the need for it is this:

```suggestion
    def snapshots(self) -> "ArtifactsSnapshotCollectionClient[T]":
```

`mypy` knows how to resolve the above.

src/dioptra/client/artifacts.py
```diff
@@ -121,36 +128,61 @@ def get_by_id(self, artifact_id: str | int) -> T:
...
+            "group": int(group_id),
+            "job": int(job_id),
+            "artifactUri": artifact_uri,
         }
 
+        if plugin_snapshot_id is not None:
+            json_["pluginSnapshotId"] = int(plugin_snapshot_id)
+
+        if task_id is not None:
+            json_["taskId"] = int(task_id)
```
jkglasbrenner, line 174: What does it mean for `plugin_snapshot_id` and/or `task_id` to be `None` here? Shouldn't serialization/deserialization definitions be mandatory?

Similarly, what happens if I just provide `plugin_snapshot_id` or `task_id` but not both? Is that currently allowed?

(This comment applies anywhere else this same pattern appears)

src/dioptra/client/artifacts.py
```diff
@@ -159,7 +191,117 @@ def modify_by_id(self, artifact_id: str | int, description: str | None) -> T:
...
+        artifact snapshot id.
+
+        Args:
+            artifact_id: The artifact resource id, an integer.
+            artifact_snapshot_id: The artifact snapshot id, an integer.
+            file_type: the file type for the bundle.
+            artifact_path: if the artifact is a directory, then a value other None
+                indicates the path when the directory structure to retrieve. if the
+                artifact is a file, None must be provided.
+            output_dir: the directory to put the downloaded artifact
```
jkglasbrenner: The docstring should explain what happens if you leave these as None. For example, `output_dir` should mention that the current working directory will be used if None.

src/dioptra/client/artifacts.py
```diff
@@ -38,12 +45,12 @@ def __init__(self, session: DioptraSession[T]) -> None:
...
+        self._snapshots = ArtifactsSnapshotCollectionClient[T](
             session=session, root_collection=self
         )
 
     @property
-    def snapshots(self) -> SnapshotsSubCollectionClient[T]:
+    def snapshots(self) -> ArtifactsSnapshotCollectionClient[T]:
         """The client for retrieving artifact resource snapshots.
 
         Each client method in the sub-collection accepts an arbitrary number of
```
jkglasbrenner, line 54: The examples at the bottom of the snapshots docstring should include something about the `get_contents` method.

src/dioptra/client/entrypoints.py
```diff
@@ -262,6 +380,123 @@ def delete_by_id(self, entrypoint_id: str | int, queue_id: str | int) -> T:
...
+        file_type: FileTypes = FileTypes.TAR_GZ,
+        output_dir: Path | None = None,
+        file_stem: str = "plugins",
+    ) -> Path:
+        """Get the plugins bundle for the entrypoint matching the provided snapshot id.
+
+        Args:
+            entrypoint_id: The entrypoint id, an integer.
+            entrypoint_snapshot_id: The entrypoint snapshot id, an integer.
+            file_type: the file type for the bundle.
```
jkglasbrenner: Some of the new arguments aren't documented in the docstring. What happens if something is left as `None` should also be mentioned as appropriate.

(This comment applies to other methods in this class)

src/dioptra/client/entrypoints.py
```diff
@@ -384,7 +627,7 @@ def modify_resource_drafts(self) -> ModifyResourceDraftsSubCollectionClient[T]:
         return self._modify_resource_drafts
 
     @property
-    def snapshots(self) -> SnapshotsSubCollectionClient[T]:
+    def snapshots(self) -> EntrypointsSnapshotCollectionClient[T]:
         """The client for retrieving entrypoint resource snapshots.
 
         Each client method in the sub-collection accepts an arbitrary number of
```
jkglasbrenner, line 643: The new methods due to switching from `SnapshotsSubCollectionClient` to `EntrypointsSnapshotCollectionClient` should be documented in the examples.

src/dioptra/client/entrypoints.py
```diff
@@ -556,10 +819,14 @@ def modify_by_id(
...
             task_graph: The new task graph for the entrypoint as a YAML-formatted
                 string.
+            artifact_graph: The artifact graph for the new entrypoint as a
+                YAML-formatted string.
             description: The new description of the entrypoint. To remove the
                 description, pass None.
             parameters: The new list of parameters for the entrypoint. To remove all
                 parameters, pass None.
+            artifact_parameters: The list of artifact parameters for the new entrypoint.
+                Optional, defaults to None.
```
jkglasbrenner: Are these accurate? With the modify calls, setting something to None tends to imply deletion, see the "parameters" description for example. Is that the case here too for artifact_graph and artifact_parameters?

src/dioptra/client/experiments.py
```diff
@@ -260,6 +260,7 @@ def create(
         queue_id: int,
         entrypoint_snapshot_id: int | None = None,
         values: dict[str, Any] | None = None,
+        artifact_values: dict[str, Any] | None = None,
```
jkglasbrenner, line 263: Missing a docstring for this new argument.

src/dioptra/client/plugins.py
```diff
@@ -292,24 +299,41 @@ def create(
...
         contents: str,
-        tasks: list[dict[str, Any]],
+        function_tasks: list[dict[str, Any]] | None = None,
+        artifact_tasks: list[dict[str, Any]] | None = None,
         description: str | None = None,
     ) -> T:
         """Creates a plugin file.
 
+        Either function_tasks or artifact_tasks should be provided. Mixing the two
+        types of tasks is not recommended.
```
jkglasbrenner, line 307: By this, do we mean that we're suggesting that a plugin file should only contain `function_tasks` OR `artifact_tasks`? If we're worried about mixing and matching causing problems, do we want to elevate this from a suggestion to a requirement and reject requests that provide non-empty lists for both kinds of tasks?

(This comment applies to other methods that have the same arguments)

src/dioptra/client/plugins.py
```diff
@@ -483,7 +568,7 @@ def modify_resource_drafts(self) -> ModifyResourceDraftsSubCollectionClient[T]:
         return self._modify_resource_drafts
 
     @property
-    def snapshots(self) -> SnapshotsSubCollectionClient[T]:
+    def snapshots(self) -> PluginsSnapshotCollectionClient[T]:
         """The client for retrieving plugin resource snapshots.
 
         Each client method in the sub-collection accepts an arbitrary number of
```
jkglasbrenner, line 577: The docstring should be updated to document the extra methods that are available due to having `PluginsSnapshotCollectionClient`.

src/dioptra/restapi/db/alembic/versions/fa39987673e0_modifications_to_support_artifact_.py
jkglasbrenner, line 1: When adding this many new related tables with additions and changes to foreign key relationships, it's best to delay the setting of the foreign keys until the end of the upgrade script. So that means not using `sa.ForeignKeyConstraint()` in the `op.create_table` calls and instead handling it in an alter table context. I can take the lead on handling this since I did it before for the v1 table migration file.

src/dioptra/restapi/db/alembic/versions/fa39987673e0_modifications_to_support_artifact_.py
```diff
@@ -0,0 +1,503 @@
...
+        # need to manually update the primary key
+        batch_op.drop_constraint(
+            constraint_name=op.f("pk_plugin_task_output_parameters"), type_="primary"
+        )
+        batch_op.create_primary_key(
+            constraint_name=op.f("pk_plugin_task_output_parameters"),
+            columns=["plugin_task_id", "name"],
+        )
+        batch_op.drop_constraint(
+            "fk_plugin_task_output_parameters_plugin_file_resource_snapshot_id_plugin_tasks",
```
jkglasbrenner: I think this might cause issues for Postgres since it's longer than 63 characters. These are usually wrapped with `batch_op.f()` or `op.f()` (depending on whether you're in an alter table context), which handles truncation of the constraint names correctly.

```suggestion
            batch_op.f("fk_plugin_task_output_parameters_plugin_file_resource_snapshot_id_plugin_tasks"),
```

There's other places it is probably needed too, this is just the particular one that caught my eye.

src/dioptra/restapi/db/models/plugins.py
```diff
@@ -135,21 +135,64 @@ class PluginPluginFile(db.Model):  # type: ignore[name-defined]
...
     file: Mapped["PluginFile"] = relationship(back_populates="tasks")
+
+    # other settings
+    __table_args__ = (
+        Index(
+            None, "plugin_file_resource_snapshot_id", "plugin_task_name", unique=True
+        ),
+    )
+    __mapper_args__ = {
+        "polymorphic_identity": "employee",
```
jkglasbrenner: I think "employee" here has to be a copy-paste error from a tutorial or the SQLAlchemy docs.

src/dioptra/restapi/db/alembic/versions/fa39987673e0_modifications_to_support_artifact_.py
```diff
@@ -0,0 +1,503 @@
...
+
+        # need to manually handle the addition of a new primary key -- perform after the
+        # alter table to allow foreign key relationships to not break
+        # creating the index above allows those relationships to continue
+        batch_op.drop_constraint(
+            constraint_name=op.f("pk_plugin_tasks"), type_="primary"
+        )
+        batch_op.create_primary_key(
+            constraint_name=op.f("pk_plugin_tasks"), columns=["task_id"]
+        )
```
jkglasbrenner: This is very likely going to fail if you have a `plugin_tasks` table that contains one or more records. This will happen for a couple of reasons:

- The `task_id` column, which is replacing the original primary key, is going to be empty when added to any non-empty plugin_tasks table. Without backfilling in a primary key value, the migration will fail when you try to set the new primary key.
- The `type` column is also going to be empty, we should fill in the function task type for any pre-existing records.

The `function_tasks` table also needs to be appropriately populated with existing values based on the above.

The reverse operation needs to be in the downgrade method as well.

src/dioptra/restapi/db/models/plugins.py
```diff
@@ -135,21 +135,64 @@ class PluginPluginFile(db.Model):  # type: ignore[name-defined]
...
+    plugin_task_name: Mapped[text_] = mapped_column(nullable=False)
+    type: Mapped[str] = mapped_column(init=False, nullable=False)
 
     # Relationships
     file: Mapped["PluginFile"] = relationship(back_populates="tasks")
+
+    # other settings
+    __table_args__ = (
+        Index(
+            None, "plugin_file_resource_snapshot_id", "plugin_task_name", unique=True
```
jkglasbrenner, line 152: This will mean that a function task and an artifact task cannot share the same name. I _think_ that's correct (you can't use the same function name twice in the same Python file), but we should be sure that there isn't something we're missing.

src/dioptra/restapi/db/models/plugins.py
```diff
@@ -135,21 +135,64 @@ class PluginPluginFile(db.Model):  # type: ignore[name-defined]
...
+        "polymorphic_identity": "employee",
+        "polymorphic_on": "type",
+    }
+
+
+class FunctionTask(PluginTask):
+    __tablename__ = "function_tasks"
+    function_task_id: Mapped[intpk] = mapped_column(
+        ForeignKey("plugin_tasks.task_id"), init=False
+    )
```
jkglasbrenner: The convention that we used for `ResourceSnapshot` polymorphism is to use the same field name for the foreign key managing the polymorphic one-to-one mappings to avoid confusion.

```suggestion
    task_id: Mapped[intpk] = mapped_column(
        ForeignKey("plugin_tasks.task_id"), init=False
    )
```

src/dioptra/restapi/db/models/plugins.py
```diff
@@ -135,21 +135,64 @@ class PluginPluginFile(db.Model):  # type: ignore[name-defined]
...
+        "polymorphic_load": "selectin",
+        "polymorphic_identity": "function",
+    }
+
+
+class ArtifactTask(PluginTask):
+    __tablename__ = "artifact_tasks"
+    artifact_task_id: Mapped[intpk] = mapped_column(
+        ForeignKey("plugin_tasks.task_id"), init=False
     )
```
jkglasbrenner: The convention that we used for `ResourceSnapshot` polymorphism is to use the same field name for the foreign key managing the polymorphic one-to-one mappings to avoid confusion.

```suggestion
    task_id: Mapped[intpk] = mapped_column(
        ForeignKey("plugin_tasks.task_id"), init=False
    )
```

src/dioptra/restapi/db/models/plugins.py
```diff
@@ -135,21 +135,64 @@ class PluginPluginFile(db.Model):  # type: ignore[name-defined]
...
-        back_populates="task", lazy="joined"
+        lazy="joined"
     )
     output_parameters: Mapped[list["PluginTaskOutputParameter"]] = relationship(
-        back_populates="task", lazy="joined"
+        lazy="joined"
+    )
+
+    __mapper_args__ = {
+        "polymorphic_load": "selectin",
```
jkglasbrenner, line 179: Why `"selectin"`? `ResourceSnapshot` doesn't specify this, and it seems like you can get the same behavior on demand using `selectin_polymorphic`, which is less than mandating it in the ORM definition itself: https://docs.sqlalchemy.org/en/20/orm/queryguide/inheritance.html#polymorphic-selectin.

src/dioptra/restapi/db/models/plugins.py
```diff
@@ -160,7 +203,7 @@ class PluginTaskParameterType(ResourceSnapshot):
     name: Mapped[text_] = mapped_column(nullable=False, index=True)
     structure: Mapped[optionaljson_] = mapped_column(nullable=True)
 
-    # Relationships
+    # Relationships -- TODO: are these really needed?
```
jkglasbrenner, line 213: Regarding the question, in theory these relationships might be overkill, and we could check if any code currently accesses them.

Looking forward a bit, it might be helpful to have this when asking the question "What tasks accept this type as input/output?" I can see that being valuable if we have a tool for interactively building the task graph and providing a list of valid options.

src/dioptra/restapi/db/models/plugins.py
```diff
@@ -201,38 +245,18 @@ class PluginTaskInputParameter(db.Model):  # type: ignore[name-defined]
...
 
-    # Additional settings
-    __table_args__ = (
-        Index(
-            None,
-            "plugin_file_resource_snapshot_id",
-            "plugin_task_name",
-            "name",
-            unique=True,
-        ),
```
jkglasbrenner, line 219: This unique index that was removed was to make sure you didn't use the same keyword argument name twice in the same plugin task definition, since that's not defined behavior in a function definition. Any reason to not use something like:

```python
        Index(
            None,
            "plugin_task_id",
            "name",
            unique=True,
        ),
```

src/dioptra/restapi/db/models/plugins.py
```diff
@@ -188,8 +231,9 @@ class PluginTaskInputParameter(db.Model):  # type: ignore[name-defined]
     __tablename__ = "plugin_task_input_parameters"
 
     # Database fields
-    plugin_file_resource_snapshot_id: Mapped[intpk] = mapped_column(init=False)
-    plugin_task_name: Mapped[text_] = mapped_column(init=False, primary_key=True)
+    plugin_task_id: Mapped[intpk] = mapped_column(
+        ForeignKey("plugin_tasks.task_id"), init=False
+    )
```
jkglasbrenner: Same as before, recommend to keep the column names consistent between tables.

```suggestion
    task_id: Mapped[intpk] = mapped_column(
        ForeignKey("plugin_tasks.task_id"), init=False
    )
```

src/dioptra/restapi/db/models/plugins.py
```diff
@@ -201,38 +245,18 @@ class PluginTaskInputParameter(db.Model):  # type: ignore[name-defined]
...
 
 class PluginTaskOutputParameter(db.Model):  # type: ignore[name-defined]
     __tablename__ = "plugin_task_output_parameters"
 
     # Database fields
-    plugin_file_resource_snapshot_id: Mapped[intpk] = mapped_column(init=False)
-    plugin_task_name: Mapped[text_] = mapped_column(init=False, primary_key=True)
+    plugin_task_id: Mapped[intpk] = mapped_column(
+        ForeignKey("plugin_tasks.task_id"), init=False
+    )
```
jkglasbrenner: Same as before, recommend to keep the column names consistent between tables.

```suggestion
    task_id: Mapped[intpk] = mapped_column(
        ForeignKey("plugin_tasks.task_id"), init=False
    )
```

src/dioptra/restapi/db/models/entry_points.py
```diff
@@ -47,17 +48,23 @@ class EntryPoint(ResourceSnapshot):
...
     task_graph: Mapped[text_] = mapped_column(nullable=False)
-
+    artifact_graph: Mapped[text_] = mapped_column(nullable=False)
     # Relationships
     parameters: Mapped[list["EntryPointParameter"]] = relationship(
         back_populates="entry_point"
     )
+    artifact_parameters: Mapped[list["EntryPointArtifact"]] = relationship(
+        back_populates="entry_point"
+    )
```
jkglasbrenner: Why is this relationship called `artifact_parameters`, but the ORM class is named `EntryPointArtifact`? It seems inconsistent.

src/dioptra/restapi/db/models/artifacts.py
```diff
@@ -31,6 +34,12 @@ class Artifact(ResourceSnapshot):
     resource_snapshot_id: Mapped[intpk] = mapped_column(init=False)
     resource_id: Mapped[bigint] = mapped_column(init=False, nullable=False, index=True)
     uri: Mapped[text_] = mapped_column(nullable=False)
+    is_dir: Mapped[bool] = mapped_column(nullable=False)
+    file_size: Mapped[Optional[int]] = mapped_column(nullable=True)
```
jkglasbrenner: There's an `optionalbigint` you can use here. It's imported from `dioptra.restapi.db.db`.


```suggestion
    file_size: Mapped[optionalbigint] = mapped_column(nullable=True)
```

src/dioptra/restapi/db/models/entry_points.py
```diff
@@ -146,6 +153,117 @@ class EntryPointParameterValue(db.Model):  # type: ignore[name-defined]
...
+
+    # Database fields
+    job_resource_id: Mapped[intpk] = mapped_column(
+        ForeignKey("resources.resource_id"), init=False
+    )
+    entry_point_resource_snapshot_id: Mapped[intpk] = mapped_column(init=False)
+    artifact_number: Mapped[intpk] = mapped_column(init=False)
+    value: Mapped[bigint] = mapped_column(
+        ForeignKey("artifacts.resource_snapshot_id"), nullable=False, init=False
+    )
```
jkglasbrenner: I found naming this column `value`, but it's actually an ID that refers to `artifacts.resource_snapshot_id` to be a little confusing (I was trying to understand how the relationship to `Artifact` was possible). My guess is that the `artifact` relationship is the main point of access once you get this object, so why not just have this be clear that it's a snapshot id?


```suggestion
    artifact_resource_snapshot_id: Mapped[bigint] = mapped_column(
        ForeignKey("artifacts.resource_snapshot_id"), nullable=False, init=False
    )
```

src/dioptra/restapi/v1/shared/job_run_store.py
```diff
@@ -0,0 +1,270 @@
...
+
+    @abstractmethod
+    def get_artifact_file_list(
+        self, base_uri: str, subfolder_path: str
+    ) -> list[ArtifactFile]:
+        raise NotImplementedError
+
+    @abstractmethod
+    def find_artifact(self, run_id: str, uri: str) -> ArtifactFile:
+        raise NotImplementedError
```
jkglasbrenner: Protocols use `...` in the function body instead of `raise NotImplementedError` (which is more appropriate for an abstract base class). Protocols also shouldn't use `@abstractmethod` since you're not inheriting from them. Having the methods listed already tells you the method must exist in the interface.


```suggestion
class JobRunStoreProtocol(Protocol):
    def get_metrics(self, run_id: str | None) -> list[dict[str, Any]]: ...  # noqa: E704

    def get_metric_history(  # noqa: E704
        self, run_id: str, name: str, page_index: int, page_length: int
    ) -> tuple[list[dict[str, Any]], int]: ...

    def log_metric(  # noqa: E704
        self, run_id: str, name: str, value: float, step: int | None
    ) -> None: ...

    def download_artifacts(  # noqa: E704
        self, artifact_uri: str, path: str | None, destination: Path
    ) -> Path: ...

    def get_artifact_file_list(  # noqa: E704
        self, base_uri: str, subfolder_path: str
    ) -> list[ArtifactFile]: ...

    def find_artifact(self, run_id: str, uri: str) -> ArtifactFile: ...  # noqa: E704
```

src/dioptra/restapi/v1/plugin_parameter_types/service.py
```diff
@@ -533,3 +535,59 @@ def create_all(
...
+        returned_parameter_type_ids = set([x.resource_id for x in parameter_types])
+        ids_not_found = id_list - returned_parameter_type_ids
+        raise EntityDoesNotExistError(
+            "plugin task parameter types",
+            num_expected=num_ids,
+            num_found=len(parameter_types),
+            ids_not_found=sorted(list(ids_not_found)),
+        )
+
+    return {x.resource_id: x for x in parameter_types}
```
jkglasbrenner, line 593: We should try and integrate this with the repository class that's been created for plugin parameter types. I'll need to check a little more on how to do it, so leaving this as a note for now.

I'm noticing this isn't used in the service class itself, so it's a utility being called somewhere else. It's basically a read-only view, but I see that it's used in a couple different places where the `uow` class needed to access the repositories won't be available yet. So we might not be able to fully get rid of this, but I don't think this file is the right spot for this as a temporary location.

jkglasbrenner (changes requested):
Did a thorough reivew of the changes to the Dioptra client, the ORM models, and the alembic migration script. I did some spotchecking on the rest of the changes outside of the service layer, since everyone else that has been looking at this have been exercising that behavior.

Some comments are minor, several are about making sure we're following consistent patterns in naming and in writing docstrings, and a couple are asking questions about a decision that was made. The biggest TODO from this is further cleanup and work on the alembic migration file, which I'll take care of.

src/dioptra/sdk/tasks/artifact.py
hbooth: removed enum restriction and switched to just string

src/dioptra/sdk/tasks/artifact.py
hbooth: added "_artifact" suffix to all of the examples

src/dioptra/client/artifacts.py
hbooth: I didn't realize this was a potential problem. I went through and removed almost all of the "from __future__ import annotations" statements and fixed any issues that popped up(all of them were in relation to the use of types).

src/dioptra/client/artifacts.py
hbooth: as stated above, this has been updated

src/dioptra/client/artifacts.py
hbooth: added/updated documentation

src/dioptra/client/artifacts.py
hbooth: added/updated documentation

src/dioptra/client/artifacts.py
hbooth: added/updated documentation

src/dioptra/client/entrypoints.py
hbooth: added/updated documentation

src/dioptra/client/entrypoints.py
hbooth: added/updated documentation

src/dioptra/client/entrypoints.py
hbooth: yes. added/updated documentation

src/dioptra/client/experiments.py
hbooth: added/updated documentation

src/dioptra/restapi/v1/workflows/lib/run_dioptra_job.py.tmpl
```diff
@@ -1,194 +0,0 @@
...
-
-    try:
-        dioptra_client.jobs.set_mlflow_run_id(
-            job_id=JOB_ID, mlflow_run_id=active_run.info.run_id
-        )
-        mlflow.set_tag(DIOPTRA_JOB_ID, JOB_ID)
-        mlflow.log_dict(job_yaml, Path(JOB_YAML_PATH).name)
-        mlflow.log_params(job_parameters)
-
-        with sys_path_dirs(dirs=(str(plugins_dir),)), env_vars({"__JOB_ID": str(JOB_ID)}):
```
jtsextonMITRE, line 146: This __JOB_ID is no longer set (or maybe it's just moved?) and would be useful to have in plugins

src/dioptra/client/plugins.py
hbooth: Looping in @jtsextonMITRE as he may have had a hand in this too. Looking through things, it doesn't affect anything at this level. The issue comes in with entrypoint which has a check to see if there is an overlap between the artifact plugins and the function plugins. We don't want the same plugin used in the same job due to name collisions. The recommendation helps users to avoid this situation with maximum flexibility for the use of their plugin tasks, but they can avoid it by being careful and not registering the same plugin in both the artifact plugins and the function plugins for an entrypoint.

src/dioptra/client/plugins.py
hbooth: added/updated documentation

src/dioptra/restapi/db/models/plugins.py
hbooth: yep, fixed --> plugin_task

src/dioptra/client/plugins.py
jtsextonMITRE: I can't figure out how to link it, but I think there is a check for this on the RESTAPI side in `_add_plugin_tasks` of `src/dioptra/restapi/v1/plugins/service.py`. Maybe it's worth checking on the client as well?

src/dioptra/restapi/db/models/plugins.py
hbooth: yes, that was the intent of the constraint. this aligns with restrictions in the python language itself.

src/dioptra/restapi/db/models/plugins.py
hbooth: made changes as requested

src/dioptra/restapi/db/models/plugins.py
hbooth: made changes as requested

src/dioptra/restapi/db/models/plugins.py
hbooth: sorry, my first reply was not quite correct. you are asking why I am using this attribute at all. I assume there is never a case where you don't want both the actual inherited type and the base type together. which is the case for plugin tasks. I can see how a different decision for `ResourceSnapshot` would make sense.

src/dioptra/restapi/db/models/plugins.py
hbooth: I did check and they aren't used in the code the moment. Coming new I was confused why they were here. If you don't have the relations a query would likely be able to do the same thing.

src/dioptra/restapi/db/models/plugins.py
hbooth: the `name` property had 'primary_key=True', but I missed that both `plugin_task_id` and `parameter_number` were also labeled as primary keys (through `intpk`). Resolution for this was to remove the `primary_key` from `name` property and add the unique constraint as suggested.

src/dioptra/restapi/db/models/plugins.py
hbooth: update property as requested

src/dioptra/restapi/db/models/plugins.py
hbooth: updated property name as requested -- this also resulted in other changes which I will send to you separately

src/dioptra/restapi/v1/workflows/lib/run_dioptra_job.py.tmpl
jtsextonMITRE:
```
        with env_vars({"__JOB_ID": str(job_id)}):
            run_experiment(
                experiment_desc=job_yaml,
                global_parameters=job_parameters,
                artifact_parameters=artifact_parameters,
                artifact_tasks=artifact_tasks,
                artifacts_dir=context.artifacts_dir,
                plugins_dir=context.plugins_dir,
                serialize_dir=context.serialize_dir,
                deserialize_dir=context.deserialize_dir,
            )
```

This change seems to work in `sdk/run_dioptra_job.py`

hbooth: just so I understand, why would a plugin need a job id?

jtsextonMITRE: @hbooth Currently used for uploading metrics to the current job via RESTAPI

src/dioptra/restapi/db/models/entry_points.py
hbooth: despite my early misgivings with such a long name, given the questions surrounding this I renamed `EntryPointArtifact` to `EntryPointArtifactParameter` and consistently applied this naming convention. This should resolve the inconsistencies noted above.

src/dioptra/restapi/db/models/artifacts.py
hbooth: switched.

src/dioptra/restapi/v1/shared/job_run_store.py
hbooth: per our discussion, I updated to use this style to be consistent with other usage of `Protocol` within the project. I also added some additional documentation

src/dioptra/restapi/db/models/entry_points.py
hbooth: ended up making this change

src/dioptra/restapi/v1/workflows/lib/run_dioptra_job.py.tmpl
hbooth: added this

src/dioptra/restapi/db/alembic/versions/fa39987673e0_modifications_to_support_artifact_.py
jkglasbrenner: Finished. I was able to get the migrations working in full on Postgres, both without data and with data. SQLite unfortunately does not work if you have existing data, it's probably due to the use of `op.batch_alter_table` and the way it's handled for SQLite databases. It seems like you can only change the schema of existing tables _or_ insert data with SQLite, but not both (creating tables and inserting data seems okay based on past scripts working without issue).

We made the decision that this breakage is okay since SQLite is only supported to make development easier. Anything that's going to be deployed and shared by multiple users needs to use Postgres.

src/dioptra/restapi/db/alembic/versions/fa39987673e0_modifications_to_support_artifact_.py
jkglasbrenner: Fixed as part of the alembic migration script update. Postgres works without complaints now.

src/dioptra/restapi/db/alembic/versions/fa39987673e0_modifications_to_support_artifact_.py
jkglasbrenner: Fixed as part of the alembic migration script update.

src/dioptra/restapi/db/models/plugins.py
jkglasbrenner: Now that I see what the `artifact_tasks` and `function_tasks` tables look like in the database (they're just a single column), I see what you mean here and I agree.

src/dioptra/restapi/db/models/plugins.py
jkglasbrenner: Fair enough. I did take a maximalist approach to the SQLAlchemy relationships when I set up the first instantiation of the v1 ORM since I didn't know what we were and weren't going to need/use and I wanted to try and keep the level of change being done to the ORM models to a minimum. We have a better idea of how things are working now, and when we finish the rest of the repo pattern, maybe we should go through and trim out relationships that aren't currently being used (or remove some of the back populations at the very least).

src/dioptra/restapi/v1/plugin_parameter_types/service.py
jkglasbrenner: We can punt on where this function should live for now and leave it where it is. We can re-evaluate after the repo pattern is done.

PR #909, opened 2025-07-29
fix: some resources are unable to search by tag (#896)

### LINKED ISSUE #896: [BUG] Some resources are not searchable by tag
Can't search models, plugins, plugin files, entrypoints by tag.

need to add something like `"tag": lambda x: models.Job.tags.any(models.Tag.name.like(x, escape="/")),` to `SEARCHABLE_FIELDS`

hbooth:
closes #896
In addition to fixing #896 

- moved search parser grammar tests and added additional tests to expand coverage
- updated pyproject.toml so that it will include the tests source code 
      - flagging this change for @jkglasbrenner
      - adding the tests source code directory makes searching and "find all references" behave better in vscode as vscode uses the settings in pyproject.toml to determine what directories will be used. I am unsure of whether there are other side effects, but everything seems to be ok.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -282,7 +282,7 @@ honor_noqa = true
 [tool.mypy]
 python_version = "3.11"
 platform = "linux"
-mypy_path = "src,task-plugins"
+mypy_path = "src,tests"
 explicit_package_bases = true
 namespace_packages = true
 show_column_numbers = true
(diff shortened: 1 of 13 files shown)


jkglasbrenner (approved):
Changes LGTM.

> adding the tests source code directory makes searching and "find all references" behave better in vscode as vscode uses the settings in pyproject.toml to determine what directories will be used. I am unsure of whether there are other side effects, but everything seems to be ok.

This shouldn't cause any issues. The change to `mypy_path` affects what gets scanned if you run `mypy` without any arguments. The `mypy` test in `tox.ini` explicitly specifies `src/dioptra`, so it doesn't change what that test checks (not sure it makes sense to make the mypy check fail due to typing issues in `tests/`).

Since we don't use pyright in any of our tests, modifying `tool.pyright` only impacts pylance in VSCode, so that's also safe.

PR #912, opened 2025-07-30
Entrypoint form with artifacts

henrychoy:
This is the entrypoint form updated with artifacts.  Mainly wanted to focus on the functionality first, and then we can make visual improvements.


src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -27,108 +27,115 @@
...
-                />
-                Queues are not yet avaiable in Entrypoint snapshots
-              </div>
+              <q-icon
+                name="sym_o_info"
+                size="2em"
+                color="grey"
+                class="q-mr-sm"
+              />
+              Queues are not yet avaiable in Entrypoint snapshots
```
hbooth: avaiable -> available

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -27,108 +27,115 @@
...
+              <label :class="`field-label`">Description:</label>
+            </template>
+          </q-input>
+        </div>
+      </div>
+    </q-form>
+  </fieldset>
+
+  <fieldset class="q-px-lg q-mt-lg q-pt-lg">
+    <legend>Function Info</legend>
```
hbooth: Task Graph Info

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -27,108 +27,115 @@
...
+    <div class="row">
+      <div :class="`${isMobile ? 'col-12' : 'col-6'} q-mr-xl column`">
+        <h2>Function Task Graph</h2>
```
hbooth: I think I prefer "Task Graph", particularly in light of my other comments

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -140,223 +147,450 @@
...
-              dense 
-              v-model.trim="parameter.defaultValue"
-              class="q-mb-sm"
-              label="Enter Default Value"
-            />
-            <q-card-actions align="right">
+      </div>
+      
+      <div :class="`${isMobile ? 'q-mt-xl' : ''} col`">
+        <h2>Function Parameters</h2>
```
hbooth: Parameters or "Entrypoint Parameters"
-----
this section should be outside of the "Task Info" section -- perhaps create a "Parameters" section and both the "Entrypoint Parameters" and the "Artifact Parameters" could go there

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -140,223 +147,450 @@
...
+                  <span class="sr-only">Add Parameter</span>
+                  <q-tooltip>
+                    Add Parameter
+                  </q-tooltip>
+                </q-btn>
+              </q-card-actions>
+            </q-form>
+          </q-card>
+
+           <h2 class="q-mt-xl">Function Plugins</h2>
```
hbooth: this section belongs inside "Task Info"
Function Plugins --> "Task Plugins"

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -140,223 +147,450 @@
...
                 color="secondary"
                 text-color="white"
+              />
+            </template>
+          </q-select>
+          <div 
+            class="row items-center q-mt-md" 
+            v-if="route.params.id !== 'new'"
+          >
+            <label class="field-label">Plugins:</label>
```
hbooth: Task Plugins

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -140,223 +147,450 @@
...
+                    Sync to latest version of plugin
+                  </q-tooltip>
+                </q-btn>
+              </div>
             </div>
           </div>
+          <TableComponent
+            :rows="tasks"
+            :columns="taskColumns"
+            title="Plugin Tasks"
```
hbooth: Function Tasks

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -140,223 +147,450 @@
...
         </div>
       </div>
+    </div>
+  </fieldset>
+
+  <fieldset class="q-px-lg q-mt-lg q-pt-lg">
+    <legend>Artifact Info</legend>
+    <div class="row">
+      <div :class="`${isMobile ? 'col-12' : 'col-6'} q-mr-xl column`">
+        <h2>Artifact Task Graph</h2>
```
hbooth: Artifact Task Graph -> Artifact Output Graph

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -140,223 +147,450 @@
...
+        <p class="text-caption q-mb-none text-grey-8 q-pl-xs">
+          Use "Add to Task Graph" button in Plugin Tasks table to insert YAML, and
```
hbooth: Use "Add to Task Graph" button in Plugin Tasks table to insert YAML
-->
Use "Add to Artifact Output Graph" button in Artifact Tasks table to insert YAML

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -140,223 +147,450 @@
...
-        <template #body-cell-outputParams="props">
-          <div v-for="(param, i) in props.row.outputParams" :key="i">
-            <q-chip
-              color="purple"
-              text-color="white"
-              dense
-              clickable
-              class="q-mr-none"
+      <div :class="`${isMobile ? 'q-mt-xl' : ''} col`">
+        <h2>Artifact Parameters</h2>
```
hbooth: This should go into a "Parameters" Section along with the "Entrypoint Parameters"

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -572,6 +849,11 @@
...
   ]
 
+  const artifactColumns = [
+    { name: 'name', label: 'Name', align: 'left', field: 'name', sortable: true, },
+    { name: 'outputParams', label: 'Output Params', align: 'right', field: 'outputParams', sortable: false, classes: 'vertical-top' },
+  ]
+
   const taskColumns = [
     { name: 'pluginName', label: 'Plugin', align: 'left', field: 'pluginName', sortable: true, },
     { name: 'taskName', label: 'Task', align: 'left', field: 'name', sortable: true, },
```
hbooth: Task -> Function Task for "Task Info" section
Task -> Artifact Task for "Artifact Info" section


for "Add to Task Graph" below you need to switch it so that it says "Add to Artifact Output Graph" for "Artifact Info" section

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -140,223 +147,450 @@
...
+              <q-btn 
+                icon="sym_o_delete"
+                round size="md"
+                color="negative"
+                flat
+                @click="selectedTaskProps = props; selectedTaskProps.taskType = 'artifacts'; showDeleteDialog = true"
+              />
+            </template>
+          </TableComponent>
+          <ArtifactParamForm
```
hbooth: sent a message in slack about this

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -140,223 +147,450 @@
...
+            v-if="!history"
+            @submit="addArtifactParam"
+          />
+
+          <h2 class="q-mt-xl">Artifact Plugins</h2>
```
hbooth: Artifact Task Plugins

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -140,223 +147,450 @@
...
-        <template #body-cell-add="props">
-          <q-btn icon="add" round size="xs" color="grey-5" text-color="black" @click="addToTaskGraph(props.row)" />
-        </template>
-      </TableComponent>
-    </fieldset>
-  </div>
+          <TableComponent
+            :rows="artifactTasks"
+            :columns="taskColumns.filter(col => col.name !== 'inputParams')"
+            title="Plugin Tasks"
```
hbooth: Artifact Tasks

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -140,223 +147,450 @@
...
+            :hideToggleDraft="true"
+            :hideSearch="true"
+            :disableSelect="true"
+            :hideOpenBtn="true"
+            :hideDeleteBtn="true"
+            :hideCreateBtn=true
+            :style="history ? 'opacity: 0.5; pointer-events: none;' : ''"
+          >
+            <template #body-cell-inputParams="props">
```
hbooth: this section should be deleted -- no input parameters

keithmanville (commented):
For the `+` buttons to add artifact tasks to the artifact graph, the yaml format needs to be fixed. definition of an artifact task  is different than a function task. See example here:

```
output:
  contents: $goodbye_step.greeting
  task:
    args:
      type: txt
    name: StringArtifactTask
```

`output` is the equivalent of step name like with function tasks.
`task` is a keyword that should always be there
`args` is a keyword that should always be there
`type` is the name of an argument specific to that task. If the args are returned int he artifact task response, these can be filled with placeholders (<value>) for the values
`name` is a keyword that should always be there, and it's value can be filled from the artifact task name

---

I noticed that the following showed up in the artifact yaml box:

```
null
...
```

To replicate, import resources from dev, then look at "Hello Artifact." removing the text and saving then properly shows nothing in the box. Maybe the difference between an empty string and None being for the artifact graph?

---

Autocomplete on when typing `$` isn't working for artifact paramters.

---

There is no way to remove an artifact param in the table.

keithmanville:
While reviewing this PR, I found #921

keithmanville:
Taking another pass after your changes. Referring to the 4 parts of this comment:
https://github.com/usnistgov/dioptra/pull/912#pullrequestreview-3100784076

1. You addressed this for the most part. I think the lingering issue is that it appears to always add `type:` under `args`. `type` is an arg specific to the artifact task I was using as an example. If we don't have the args for each artifact task already in an existing response, we can just populate with `args:` and nothing after it. If we do have the list of args, we can list them with placeholders for the values.
2. not resolved
3. resolved
4. resolved

PR #913, opened 2025-07-30
Migrate project build toolchain to uv

jkglasbrenner:


jtsextonMITRE (approved):
The docker containers looked fine to me!

Some notes for Windows dev environment:

* source and make do not work out of the box on Windows, so for getting flask running, `.venv/Scripts/activate`, `uv sync`, `$Env:DIOPTRA_RESTAPI_DEV_DATABASE_URI = "sqlite:///C:/<path-to-deployments>/dioptra-dev.db" `, `dioptra-db autoupgrade` and `flask run` should be run manually
* the dev-kb local-setup README does not include instructions to clone the branch prior to starting, but this might be out of scope for that page.

JustKuzya:
- I think that we should add the instructions for the **sqlite expected location** to the `dev-kb local-setup README` guide as an addition I dislike deeply when the documents refer to each other and you have an anticipation of the infinite time-horizon, instead of the feasible one in one document. 
- The `dioptra-db autoupgrade` and `flask run` should be either working from the script, or alternatively the script would need to be modified or duplicated for Windows, vs MacOS/Linux, if necessary.
-  I just got PC setup yesterday by iTAC, not yet setup for Dioptra, but I will be able to test both MacOS and Windows usage using both machines if necessary.

jkglasbrenner:
> The docker containers looked fine to me!
> 
> Some notes for Windows dev environment:
> 
> * source and make do not work out of the box on Windows, so for getting flask running, `.venv/Scripts/activate`, `uv sync`, `$Env:DIOPTRA_RESTAPI_DEV_DATABASE_URI = "sqlite:///C:/<path-to-deployments>/dioptra-dev.db" `, `dioptra-db autoupgrade` and `flask run` should be run manually

I pushed updates for `dev-kb/local-setup/MANUAL.md` to include Windows-specific instructions to address this comment. I also added a notes to make it more clear what can and can't be done in a Windows native environment, and that Windows users will need to use WSL2 if they need to run all components of Dioptra.

> * the dev-kb local-setup README does not include instructions to clone the branch prior to starting, but this might be out of scope for that page.

I added this to `dev-kb/local-setup/README.md` so that it's explicit.

PR #916, opened 2025-07-31
refactor: remove name method from ArtifactTaskInterface

hbooth:
- Remove name() from ArtifactTaskInterface
- Updated code to use the name of the class in place of the result from name() method
- Updated implementations in extra/plugins/artifacts/tasks.py
- Updated Examples


keithmanville (approved):
Approved
I fixed a small issue with the artifact task names in sklearn-classification.yaml not being updated

PR #922, opened 2025-08-11
examples: update MNIST example for artifacts

### LINKED ISSUE #596: Refactor the MNIST classifier example using the v1 Python client
Refactor the MNIST classifier example with the v1 client

This includes pulling together the required plugins, refactoring the entrypoint yaml, and updating the notebook to use the new client.

Blocked by:
- #530

### Definition of done
- [ ] The MNIST classifier example is refactored and the notebook is updated
- [ ] The example jobs run successfully
- [ ] The example is merged into dev

jtsextonMITRE:
Closes #596

### DIFF
--- a/examples/mnist_demo.ipynb
+++ b/examples/mnist_demo.ipynb
@@ -31,7 +31,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 73,
+   "execution_count": 1,
    "metadata": {
     "tags": []
    },
@@ -49,7 +49,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 74,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -187,7 +187,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 75,
+   "execution_count": 3,
    "metadata": {
     "tags": []
    },
@@ -206,7 +206,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 89,
+   "execution_count": 13,
    "metadata": {
     "scrolled": true
    },
@@ -217,7 +217,7 @@
        "{'username': 'pluginuser2', 'status': 'Login successful'}"
       ]
      },
-     "execution_count": 89,
+     "execution_count": 13,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -247,7 +247,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 90,
+   "execution_count": 14,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -273,7 +273,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 91,
+   "execution_count": 15,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -282,7 +282,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 92,
+   "execution_count": 16,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -314,7 +314,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 93,
+   "execution_count": 17,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -346,7 +346,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 97,
+   "execution_count": 18,
    "metadata": {},
    "outputs": [
     {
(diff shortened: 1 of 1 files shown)


extra/plugins/mnist_demo/artifacts_utils.py
```diff
@@ -0,0 +1,43 @@
...
+
+import structlog
+from structlog.stdlib import BoundLogger
+
+from dioptra import pyplugs
+
+LOGGER: BoundLogger = structlog.stdlib.get_logger()
+
+@pyplugs.register
+def make_directories(dirs: list[str | Path]) -> None:
```
keithmanville: We don't use this as a plugin task and I feel like it is a remnant of the "old" pattern for designing entrypoints. It's only used to create a single output directory for fgsm and patch attack, so I would just call Path.mkdir directly in those functions. Then this file can be removed.

extra/plugins/mnist_demo/plugins.py
```diff
@@ -0,0 +1,350 @@
...
+    fit_kwargs = {} if fit_kwargs is None else fit_kwargs
+    callbacks_list = [] if callbacks_list is None else callbacks_list
+
+    callbacks = get_model_callbacks(callbacks_list)
+    fit_kwargs['callbacks'] = callbacks
+    fit_tensorflow(estimator=estimator, x=x, y=y, fit_kwargs=fit_kwargs)
+    return estimator
+
+@pyplugs.register    
+def save_models(
```
keithmanville: not used in any entrypoint, so should be removed. I think all of the log_tensorflow and add_model_to_registry stuff is legacy as well and should be removed too.

extra/plugins/mnist_demo/tracking_mlflow.py
```diff
@@ -0,0 +1,98 @@
...
+        post_metrics(metric_name=metric_name, metric_value=metric_value)
+        LOGGER.info(
+            "Logging metric",
+            metric_name=metric_name,
+            metric_value=metric_value,
+        )
+
+
+@pyplugs.register
+def log_parameters(parameters: dict[str, float]) -> None:
```
keithmanville: legacy mlflow stuff, remove.

extra/plugins/mnist_demo/tracking_mlflow.py
```diff
@@ -0,0 +1,98 @@
...
+        LOGGER.info(
+            "Log parameter to MLFlow Tracking server",
+            parameter_name=parameter_name,
+            parameter_value=parameter_value,
+        )
+
+
+@pyplugs.register
+@require_package("tensorflow", exc_type=TensorflowDependencyError)
+def log_tensorflow_keras_estimator(estimator: Sequential, model_dir: str) -> None:
```
keithmanville: legacy mlflow stuff, remove.

extra/plugins/mnist_demo/tracking_mlflow.py
```diff
@@ -0,0 +1,98 @@
...
+
+except ImportError:  # pragma: nocover
+    LOGGER.warn(
+        "Unable to import one or more optional packages, functionality may be reduced",
+        package="tensorflow",
+    )
+
+
+@pyplugs.register
+def log_metrics(metrics: dict[str, float]) -> None:
```
keithmanville: move this to the same file as post_metrics. The other functions in this file can be removed and we are separating out from mlflow at this point since the user is now only interacting with our API.

extra/plugins/mnist_demo/tracking_mlflow.py
```diff
@@ -0,0 +1,98 @@
...
+    """Logs metrics to the MLFlow Tracking service for the current run.
```
keithmanville: remove mlflow references

extra/plugins/mnist_demo/tensorflow.py
```diff
@@ -0,0 +1,148 @@
+# This Software (Dioptra) is being made available as a public service by the
```
keithmanville, line 1: It seems most/all of this is tf support functionality, and not plugin tasks meant to be used directly in an entrypoint, so they probably shouldn't be registered with pyplugs.

extra/plugins/mnist_demo/import_keras.py
```diff
@@ -0,0 +1,64 @@
+# This Software (Dioptra) is being made available as a public service by the
```
keithmanville, line 1: Why do we need this whole layer of indirection? Can we just keras metrics/callbacks/optimizers directly where needed? Seems like this was some kind of compatibility thing we had at some point? Do we still need it?

extra/plugins/mnist_demo/registry_mlflow.py
```diff
@@ -0,0 +1,97 @@
+# This Software (Dioptra) is being made available as a public service by the
```
keithmanville, line 1: I think this is all legacy mlflow stuff and can be removed.

extra/plugins/mnist_demo/attacks_fgm.py
```diff
@@ -0,0 +1,284 @@
...
+            distance_metrics_=distance_metrics_,
+            clean_batch=x.numpy(),
+            adv_batch=adv_batch,
+            distance_metrics_list=distance_metrics_list,
+        )
+
+    LOGGER.info("Adversarial image generation complete", attack="fgm")
+    _log_distance_metrics(distance_metrics_)
+
+    return pd.DataFrame(distance_metrics_)
```
keithmanville: return batch directory in addition to distance metric for artifact capture.

extra/plugins/mnist_demo/attacks_fgm.py
```diff
@@ -0,0 +1,284 @@
...
+        package="tensorflow",
+    )
+
+
+@pyplugs.register
+@require_package("art", exc_type=ARTDependencyError)
+@require_package("tensorflow", exc_type=TensorflowDependencyError)
+def fgm(
+    data_flow: Dataset,
+    adv_data_dir: str | Path,
```
keithmanville, line 84: now that we have artifacts, I think we should try to avoid passing around file paths. Unless there's something I'm missing this just adds additional things the user needs to think about. This task should produce an output that can be saved as an artifact. And it's the output (which can be passed directly to another task) or the artifact (which the user can decide to capture) that user should care about.

extra/plugins/mnist_demo/attacks_fgm.py
```diff
@@ -0,0 +1,284 @@
...
+        adv_batch: A generated batch of adversarial images.
+        adv_data_dir: The directory to use when saving the generated adversarial images.
+        y: An array containing the target labels of the original images.
+        clean_filenames: A list containing the filenames of the original images.
+    """ 
+    for batch_image_num, adv_image in enumerate(adv_batch):
+        adv_image_path = (
+            adv_data_dir
+            / f"{y[batch_image_num]}"
+            / f"adv_{clean_filenames[batch_image_num].name}"
```
keithmanville, line 223: `save_img` will save images according to file extension by default. We shouldn't rely on whatever the input file type is because that could lead to use saving our adversarial dataset using lossy compression (e.g. you could be inadvertently applying jpeg compression which can be seen as a pre-processing defense just because the source images were jpgs).

extra/plugins/mnist_demo/mlflow.py
```diff
@@ -0,0 +1,82 @@
+# This Software (Dioptra) is being made available as a public service by the
```
keithmanville: legacy mlflow stuff, can be removed.

extra/plugins/mnist_demo/restapi.py
```diff
@@ -0,0 +1,157 @@
...
+
+    model_version = dioptra_client.models.versions.create(
+        model_id=model_id,
+        artifact_id=artifact["id"],
+        description=f"{name} model version",
+    )
+    LOGGER.info("model created", response=model_version)
+
+
+def upload_artifact_to_restapi(
```
keithmanville: no longer needed with artifacts handling feature implemented.

extra/plugins/mnist_demo/restapi.py
```diff
@@ -0,0 +1,157 @@
...
+    url = os.environ["DIOPTRA_API"]
+    dioptra_client = connect_json_dioptra_client(url)
+    dioptra_client.auth.login(
+        username=os.environ["DIOPTRA_WORKER_USERNAME"],
+        password=os.environ["DIOPTRA_WORKER_PASSWORD"],
+    )
+    return dioptra_client
+
+
+def upload_model_to_restapi(
```
keithmanville: no longer needed with artifacts handling feature implemented.

extra/plugins/mnist_demo/restapi.py
```diff
@@ -0,0 +1,157 @@
...
+
+def post_metrics(
+    metric_name: str,
+    metric_value: float,
+    job_id: str|int|None = None,
+    metric_step: int|None = None
+) -> dict[str, Any]:
+    import math
+    dioptra_client = get_logged_in_session()
+    if math.isnan(metric_value):
```
keithmanville, line 68: is this still needed once the nan fix PR is merged?

extra/plugins/mnist_demo/restapi.py
```diff
@@ -0,0 +1,157 @@
...
+
+def get_uris_for_job(
+    job_id: str | int
+) -> list[str]:
+    dioptra_client = get_logged_in_session()
+    job = dioptra_client.jobs.get_by_id(job_id)
+    return [artifact["artifactUri"] for artifact in job["artifacts"]]
+
+
+def get_uris_for_artifacts(
```
keithmanville: not used

extra/plugins/mnist_demo/restapi.py
```diff
@@ -0,0 +1,157 @@
...
+    
+    uri = (
+        artifact["artifactUri"]
+        if selected_model is not None
+        else None
+    )
+    return uri
+
+
+def get_uris_for_job(
```
keithmanville: not used

extra/plugins/mnist_demo/plugins.py
```diff
@@ -0,0 +1,350 @@
...
+    n_classes: int = -1,
+) -> tuple[np.ndarray|None, np.ndarray|None]:
+    y_true, y_pred = None, None
+    if dataframe is not None:
+        y_true, y_pred = df_to_predictions(dataframe, dataset, n_classes)
+    return y_true, y_pred
+
+
+@pyplugs.register
+def write_model(model, name, suffix = "keras", **kwargs) -> Path:
```
keithmanville, line 316: make a artifact task

extra/plugins/mnist_demo/plugins.py
```diff
@@ -0,0 +1,350 @@
...
+    result = Path(name).with_suffix('.' + suffix)
+    model.save(result, kwargs)
+    return result
+
+@pyplugs.register
+def read_model(
```
keithmanville, line 322: make a artifact task

extra/plugins/mnist_demo/plugins.py
```diff
@@ -0,0 +1,350 @@
...
+    location: Path, 
+    imagenet_preprocessing: bool = False,
+    art: bool = False,
+    image_size: tuple[int,int,int] | None = None,
+    classifier_kwargs: dict[str, Any] | None = None
+) -> Model:
```
keithmanville, line 328: Ideally any art functions handling the model wrapping. That way load_model always returns the same thing.

We should also make sure any preprocessing is handled consistently. ideally these would be preprocessing layers in the model so it can't accidentally be separated out. Currently preprocessing isn't being applied if `art=false`.

extra/plugins/mnist_demo/plugins.py
```diff
@@ -0,0 +1,350 @@
...
+    validation_split: float | None = 0.2,
+    rescale = 1/255.,
+    batch_size: int = 32,
+    label_mode: str = "categorical",
+    shuffle: bool = True
+) -> tuple[Dataset, Dataset, Dataset]:
+    seed, rng = init_rng(ep_seed)
+    global_seed = draw_random_integer(rng)
+    dataset_seed = draw_random_integer(rng)
+    init_tensorflow(global_seed)
```
keithmanville, line 65: may want to make setting the seed plugin task to give the user more fine grained control over when it is called.

extra/plugins/mnist_demo/restapi.py
```diff
@@ -0,0 +1,157 @@
...
+from structlog.stdlib import BoundLogger
+
+from dioptra import pyplugs
+from dioptra.client import connect_json_dioptra_client, DioptraClient
+
+LOGGER: BoundLogger = structlog.stdlib.get_logger()
+
+
+@pyplugs.register
+def get_uri_for_model(
```
keithmanville, line 30: won't be needed once we make loading/saving the model an artifact task

extra/plugins/mnist_demo/restapi.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def get_uris_for_artifacts(
+    artifact_ids: list[int]
+) -> list[str]:
+    dioptra_client = get_logged_in_session()
+    return [dioptra_client.artifacts.get_by_id(artifact_id=aid)['uri'] for aid in artifact_ids]
+
+
+def post_metrics(
```
keithmanville, line 60: I think this is the only case where we actually need to call the rest api from this plugin. Makes me wonder if we should have some mechanism for logging metrics that doesn't require the user to use the client directly.

extra/plugins/mnist_demo/plugins.py
```diff
@@ -0,0 +1,350 @@
...
+from .registry_art import load_wrapped_tensorflow_keras_classifier
+from .registry_mlflow import load_tensorflow_keras_classifier
+from .restapi import get_uri_for_model
+from .tensorflow import get_optimizer, get_model_callbacks, get_performance_metrics, evaluate_metrics_tensorflow, init_tensorflow, fit_tensorflow, predict_tensorflow
+from .tracking_mlflow import log_parameters, log_tensorflow_keras_estimator, log_metrics
+
+LOGGER: BoundLogger = structlog.stdlib.get_logger()
+
+@pyplugs.register
+def load_dataset(
```
keithmanville, line 50: consider https://keras.io/api/datasets/mnist/

extra/plugins/mnist_demo/artifacts_utils.py
jtsextonMITRE: Agreed

extra/plugins/mnist_demo/plugins.py
jtsextonMITRE: I agree it's no longer used but maybe we should leave the branch in existence with deleting this since I expect whatever mlflow model integration may want to reference this

extra/plugins/mnist_demo/import_keras.py
jtsextonMITRE: @jkglasbrenner can correct me but I think this is so we can specify optimizers/metrics/callbacks by name in entrypoints without really any coding required. Like this in train.yml:

```      
      callbacks_list:
        - name: EarlyStopping
          parameters:
            monitor: val_loss
            min_delta: .01
            patience: 5
            restore_best_weights: true
```

I'm inclined to want to keep these since they make the model more customizable from the entrypoint level

extra/plugins/mnist_demo/attacks_fgm.py
jtsextonMITRE: I will replace this with a randomly generated directory name then so that calling fgm() twice doesn't have weird side effects

PR #926, opened 2025-08-18
Forward worker job logs to api

### LINKED ISSUE #925: Forward worker job logs to API
Implement a custom Python logging handler that forwards captured job logs to the Dioptra API for storage and retrieval using an authenticated `DioptraClient`.

jkglasbrenner:
This update implements real-time log forwarding from the Dioptra workers to the API, enabling jobs to stream their logs for storage and retrieval. The implementation uses a context manager that attaches a custom logging handler during job execution, which buffers log records and periodically sends them to the API's job logging endpoint. The API client is authenticated using the worker credentials that are provided in environment variables and ensures logs are flushed even if jobs terminate unexpectedly.

The forwarding mechanism includes filters to prevent recursive logging loops when the client itself emits logs while sending data to the API, and to exclude noise from third-party libraries. The worker's logging configuration has been updated to use structured logging with proper formatting for both console output and API transmission.

A new unit test verifies that the messages emitted by multiple named loggers are forwarded to the API and recorded in the database while the context manager is active, and that out-of-context messages aren't captured.

Closes #925

---

Note: Logging Handlers in the Python stdlib use camelCase for naming methods instead of snake_case. The `DioptraJobLoggingHandler` class follows this convention since it is adapted from the source code for the `logging.handlers.MemoryHandler` class.

### DIFF
--- a/src/dioptra/rq/tasks/run_v1_dioptra_job.py
+++ b/src/dioptra/rq/tasks/run_v1_dioptra_job.py
@@ -20,6 +20,7 @@ import structlog
 from structlog.stdlib import BoundLogger
 
 import dioptra.sdk.utilities.run_dioptra_job as run_dioptra_job
+from dioptra.sdk.utilities.logging import forward_job_logs_to_api
 from dioptra.sdk.utilities.paths import set_cwd
 
 LOGGER: BoundLogger = structlog.stdlib.get_logger()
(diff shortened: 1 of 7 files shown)


jkglasbrenner:
#935 should be merged before this.

keithmanville (approved):

PR #929, opened 2025-08-19
Jk i651 entity types cleanup

JustKuzya:
Added the consistent uniform way to handle entity-types in the code. Rebased to the dev as of Aug 19, resolved all merge conflicts.


src/dioptra/restapi/v1/entity_types.py
```diff
@@ -0,0 +1,268 @@
...
+        """
+         EntityTypes.get_from_string ('entity_type')
+        Args:
+            value (_type_): String entity name to get matching Enum, .NONE, or .UNDEFINED
+
+        Returns:
+            _type_: The EntityTypes Enum
+        """
+        if not resource_type_name:
+            # print(f"1. {resource_type_name=} ")
```
keithmanville: remove debug prints before merging

src/dioptra/restapi/v1/entity_types.py
```diff
@@ -0,0 +1,268 @@
...
+            name = EntityTypes.to_snake_case(resource_type_name).upper()
+            # print(f"pre-2. {name=} ")
+            for current_value in EntityTypes.list_names():
+                # for entry in
+                # print(f"2. {current_value=} vs. {name=} ")
+                if name == current_value:
+                    return EntityTypes[name]
+        # print(f"{'!' * 80}\n\t{resource_type_name=} Was Not Found in ENUM {'!' * 80}")
+        return EntityTypes.UNDEFINED
+        # ----------------------------------------------------------------------
```
keithmanville: remove these "separator" comments. Not used anywhere else in codebase.

src/dioptra/restapi/v1/entity_types.py
```diff
@@ -0,0 +1,268 @@
...
+        Returns:
+            _type_: The EntityTypes Enum
+        """
+        if not resource_type_name:
+            # print(f"1. {resource_type_name=} ")
+            return EntityTypes.NONE
+        else:
+            name = EntityTypes.to_snake_case(resource_type_name).upper()
+            # print(f"pre-2. {name=} ")
+            for current_value in EntityTypes.list_names():
```
keithmanville: instead of looping and checking current value, would it be better to do something like this?

```python
try:
    return EntityTypes[name]
except KeyError as e:
    return EntityTypes.UNDEFINED
```

src/dioptra/restapi/v1/entity_types.py
```diff
@@ -0,0 +1,268 @@
...
+        return text_to_snake.replace(" ", "_").lower()
+        # ------------------------------------------------------------------
+
+    def get_db_schema_name(self) -> str:
+        """Returns current instance name/key
+
+        Returns:
+            str: returns inst-ENUM.all-lower-case - snake
+        """
+        if self.db_schema_name == EntityTypes.to_snake_case(self.db_schema_name):
```
keithmanville: why not just
```python
return EntityTypes.to_snake_case(self.db_schema_name)
```

src/dioptra/restapi/v1/entity_types.py
```diff
@@ -0,0 +1,268 @@
...
+            readable_name (str): The name to be read by user in UI/UX/Logs/Errors
+
+        Returns:
+            _type_: The created const-style immutable tuple-member the enum
+        """
+        new_instance = object.__new__(cls)
+        new_instance._value_ = (
+            original_name  # Set default value to make sure the Enum base works
+        )
+        new_instance.db_schema_name = original_name
```
keithmanville, line 133: variable names are confusing. We set `db_schema_name` to `original_name` and then the `get_original_name()` returns the `db_schema_name`. Are we trying to make some distinction between what we call this internally/externally from the enum?

src/dioptra/restapi/v1/entity_types.py
```diff
@@ -0,0 +1,268 @@
...
+        new_instance.ui_print_name = readable_name
+        return new_instance
+        # ----------------------------------------------------------------------
+
+    NONE = "none", "Entity of Type NONE"
```
keithmanville, line 137: If there are only two variants of a name we are trying to preserve, why do we need this whole "named tuple" kind of enum? Couldn't it just be a standard enum with name being lower snake case and value being the display string?

src/dioptra/restapi/v1/shared/tags/service.py
```diff
@@ -260,6 +260,9 @@ def get(self, resource_id: int, **kwargs) -> models.Resource:
         resource = db.session.scalar(stmt)
 
         if resource is None:
-            raise EntityDoesNotExistError(self._resource_type, resource_id=resource_id)
+            raise EntityDoesNotExistError(
+                EntityTypes.get_from_string(self._resource_type),
```
keithmanville, line 264: I think in some cases were are translating to and from string representations when it's not needed. In this case, `_resource_type` ultimately comes from the calls to `generate_resource_tags_endpoint` where we pass `RESOURCE_TYPE.get_db_schema_name()`.

We could just store the enum object directly and not need to go back and forth from string representations.

src/dioptra/restapi/v1/entity_types.py
```diff
@@ -0,0 +1,268 @@
...
+    METRIC = "metric", "Metric"
+    RESOURCE = "resource", "Resource"
+    WORKFLOW = "workflow", "Workflow"
+    EXPERIMENT = "experiment", "Experiment"
+
+    ML_FLOW_RUN = "ml_flow_run", "M.L. Flow Run"
+    ML_MODEL = "ml_model", "M.L. Model"
+    ML_MODEL_VERSION = "ml_model_version", "M.L. Model Version"
+    # --- Artifact-Related Entities ---
+    HAS_DRAFT = "has_draft", "Has Draft"
```
keithmanville, line 156: I'm seeing several entries that aren't Resources. I thought the intent was to capture only Resources.

src/dioptra/restapi/v1/artifacts/service.py
```diff
@@ -46,7 +47,7 @@
 
 LOGGER: BoundLogger = structlog.stdlib.get_logger()
 
-RESOURCE_TYPE: Final[str] = "artifact"
+ARTIFACT_RESOURCE_TYPE: Final[EntityTypes] = EntityTypes.ARTIFACT
```
keithmanville, line 50: I think wherever we use `EntityTypes` now it's cleaner to
```python
from dioptra.restapi.v1.entity_types import EntityTypes
```
and then access the EntityTypes needed as `EntityTypes.ARTIFACT`.
Basically we don't need the extra `ARTIFACT_RESOURCE_TYPE` constants floating around anymore.

src/dioptra/restapi/v1/entity_types.py
```diff
@@ -0,0 +1,268 @@
...
+        Returns:
+            str: returns inst-ENUM.all-lower-case - snake
+        """
+        if self.db_schema_name == EntityTypes.to_snake_case(self.db_schema_name):
+            return self.db_schema_name
+        else:
+            return EntityTypes.to_snake_case(self.db_schema_name)
+        # ---------------------------------------------------------------------
+
+    def get_original_name(self) -> str:
```
keithmanville, line 212: remove getters and access properties directly.

src/dioptra/restapi/v1/entity_types.py
```diff
@@ -0,0 +1,268 @@
...
+    else:
+        # Dispatch bask to the original mechanism of json-dumping for non-Enum objects
+        return _saved_default
+
+
+# Glue-up the new default method with the patch-back,
+# and tell MyPy not to wrinkle it's nose at the "monkey patch"
+JSONEncoder.default = _new_default  # type: ignore
+# End of the monkey-patching the serialization fot he Enums
+# ======================================================================
```
jkglasbrenner, line 55: This kind of monkeypatch seems brittle and unpredictable, as it only is applied once this file is imported. Depending on how code is imported and called, it's very possible for `json.dump` to use different versions of the `JSONEncoder` default during the lifecycle of the application. The original default also is never restored, so this is a permanent override.

The usual way to do this kind of thing is to inherit from the `JSONEncoder` class, override the default method, and then pass it to `json.dump` or `json.dumps` explicitly:

```python
import json
from enum import Enum


class EnumJSONEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, Enum):
            return obj.name

        # Dispatch bask to the original mechanism of json-dumping for non-Enum objects
        return super().default(obj)


class Color(Enum):
    RED = 1
    GREEN = 2
    BLUE = 3


print(json.dumps({"favorite_color": Color.BLUE, "name": "First", "id": 1}, cls=EnumJSONEncoder))
```

which prints

```json
{"favorite_color": "BLUE", "name": "First", "id": 1}
```

Stepping back, where is `json.dump` or `json.dumps` being used that requires this approach? Is it an invocation of `json.dump` or `json.dumps` that we control in the codebase? Or is it being invoked by one of our dependent libraries, and there's no way to provide the custom encoder?

_If_ we truly need to override this for the entire application like this, then we should consider doing something similar to `dioptra.restapi.patches.monkey_patch_flask_restx`, which is called within `dioptra.restapi.app.create_app`. This would better control where in the application lifecycle the behavior gets changed.

src/dioptra/restapi/v1/entity_types.py
```diff
@@ -0,0 +1,268 @@
...
+        # Insert underscore before uppercase letters in camelCase/PascalCase
+        # Handles cases like:
+        #   "PascalCase" -> "Pascal_Case" and
+        #   "varInCamelCase" -> "var_In_Camel_Case"
+        text_to_snake = re.sub(r"(?<!^)(?=[A-Z])", " ", text_to_snake)
+        # Clean multiple spaces by collapsing them, if they happen
+        text_to_snake = re.sub(r"\s+", " ", text_to_snake).strip()
+        # Replace spaces with underscores
+        return text_to_snake.replace(" ", "_").lower()
+        # ------------------------------------------------------------------
```
jkglasbrenner: I consider this to be a general utility function instead of a method that is specific to the details of this class. I recommend extracting this into `dioptra.restapi.utils`, it would fit nicely right below the `slugify` function.

Also, after moving the function, I would move the regular expressions into module-level level constants and compile them, since in principle this function could be called many times.

```python
CAMEL_PASCAL_SEP_REGEX = re.compile(r"(?<!^)(?=[A-Z])")
MULTI_SPACE_REGEX = re.compile(r"\s+")

# Used like
CAMEL_PASCAL_SEP_REGEX.sub(" ", text_to_snake)
```

src/dioptra/restapi/v1/entity_types.py
jkglasbrenner: To that end, if you want to make these properties "read-only" (well, as much as you can do that in Python), prefer this pattern

```python
    def __new__(
        cls,
        original_name: str,
        readable_name: str,
    ):
        """Default constructor for stuffing EntityTypes Enum

        Args:
            original_name (str): DB-schema-safe entity type name
            readable_name (str): The name to be read by user in UI/UX/Logs/Errors

        Returns:
            _type_: The created const-style immutable tuple-member the enum
        """
        new_instance = object.__new__(cls)
        new_instance._value_ = (
            original_name  # Set default value to make sure the Enum base works
        )
        new_instance._db_schema_name = original_name
        new_instance._ui_print_name = readable_name
        return new_instance

    @property
    def db_schema_name(self) -> str:
        return EntityTypes.to_snake_case(self._db_schema_name)

    @property
    def ui_print_name(self) -> str:
        return EntityTypes.to_snake_case(self._ui_print_name)
```

You would also need to **remove** these from the class definition to make room for the above:

```python
    db_schema_name: str  # DB-compatible name representation
    ui_print_name: str  # Yser-friendly name representation
```

src/dioptra/restapi/v1/entity_types.py
jkglasbrenner: This is also what I thought. There's a mixture of REST API entities, database tables that contain attributes about a resource type that is a backend implementation detail (the artifact parameters, for example), request parameters like `PASSWORD`, and `HAS_DRAFT` which is a derived attribute that we attach to resources in the app's responses.

src/dioptra/restapi/v1/entity_types.py
```diff
@@ -0,0 +1,268 @@
...
+    ARTIFACT_TASK_INPUT_PARAMETER = (
+        "artifact_task_input_parameter",
+        "Artifact Task Input Parameter",
+    )
+    # --- Entry-Point Entities ---
+    ENTRYPOINT = "entry_point", "Entry Point"
+    ENTRYPOINT_PLUGIN = "entry_point_plugin", "Entry Point Plugin"
+    # --- Entry-Point Entities ---
+    ENTRY_POINT = "entry_point", "Entry Point"
+    ENTRY_POINT_PLUGIN = "entry_point_plugin", "Entry Point Plugin"
```
jkglasbrenner, line 175: Why not just use one of these everywhere instead of aliasing the two variants? I know there's inconsistency in naming in different sections of the code base (entrypoint is found in the Flask code, while entry_point is used in the SQLAlchemy ORM code), but will that impact which of these get used?

src/dioptra/restapi/v1/entity_types.py
jkglasbrenner: I agree with this, it's easier to understand at a glance. As a bonus, it might be more efficient (although I haven't done any timing tests to check).

src/dioptra/restapi/errors.py
```diff
@@ -316,7 +326,7 @@ def __init__(self, type: str, constraint: str, **kwargs):
 class QueryParameterNotUniqueError(QueryParameterValidationError):
     """Query Parameters failed unique validatation check."""
 
-    def __init__(self, type: str, **kwargs):
+    def __init__(self, type: EntityTypes, **kwargs):
         super().__init__(type, "unique", **kwargs)
```
jkglasbrenner, line 330: I don't think I agree with changing this error to use an enum like this. Query parameters aren't entities, they're input data that's attached to a request sent to the endpoint.

src/dioptra/restapi/errors.py
```diff
@@ -271,8 +279,10 @@ def __init__(
 class SortParameterValidationError(DioptraError):
     """The sort parameters are not valid."""
 
-    def __init__(self, type: str, column: str, **kwargs):
-        super().__init__(f"The sort parameter, {column}, for {type} is not sortable.")
+    def __init__(self, type: EntityTypes, column: str, **kwargs):
+        super().__init__(
+            f"The sort parameter, {column}, for {type.get_print_name()} is not sortable."
+        )
```
jkglasbrenner, line 285: Another place I'm not sure I would agree we should use `EntityTypes`. The sorting parameters map to attributes/properties on an Entity (or resource), they are not entities themselves.

src/dioptra/restapi/errors.py
```diff
@@ -298,11 +308,11 @@ def __init__(self, plugins: str, **kwargs):
 class QueryParameterValidationError(DioptraError):
     """Input parameters failed validation."""
 
-    def __init__(self, type: str, constraint: str, **kwargs):
+    def __init__(self, type: EntityTypes, constraint: str, **kwargs):
```
jkglasbrenner, line 311: I don't agree that `EntityTypes` belongs here, see my other comment about query parameters.

src/dioptra/restapi/v1/entity_types.py
jkglasbrenner: One argument for this is that Enum attributes tend to be expressed in the upper-cased snake case style to indicate they function as a kind of constant.

Something that wouldn't work if we just set the value to the display name is aliasing. For example:

```python
class Color(Enum):
    BLUE = 1
    GREEN = 1

print(f"This shouldn't be True, but it is: {Color.BLUE == Color.GREEN}")
print(f"Apparently green is blue: {Color.GREEN}")
```

The above isn't what we would expect based on how the `EntityTypes` class is working.

src/dioptra/restapi/v1/shared/tags/service.py
jkglasbrenner: I agree, and perhaps the way to do this is to do something like this in the `ResourceIdService`:

```python
    @inject
    def __init__(self, resource_type: EntityTypes):
        self._resource_type = resource_type
```

And then the `ResourceTagsIdService` would get this modification:

```python
    @inject
    def __init__(
        self,
        resource_type: EntityTypes,
        resource_id_service: ClassAssistedBuilder["ResourceIdService"],
        tag_id_service: TagIdService,
    ) -> None:
        self._resource_id_service = resource_id_service.build(
            resource_type=resource_type
        )
        self._tag_id_service = tag_id_service
```

And so on.

Of course, you'll also need to find where the `resource_type` argument is actually passed in each service layer definition and change that to the appropriate enum.

jkglasbrenner (changes requested):
My review focused on the scope of where `EntityTypes` should (and shouldn't) be used, some implementation details, and code organization. I also left comments on the threads started by @keithmanville when I had something to add.

src/dioptra/restapi/v1/entity_types.py
JustKuzya: I like the instance switch suggested, but I don't remember what was breaking without the patch. I agree that probably a uniform way of doing it once and for all enums is better way to handle the Enums serialization.

src/dioptra/restapi/v1/entity_types.py
JustKuzya: Agree that one index call is better than looping - did not feel natural though, even though it the more Pythonic way.

src/dioptra/restapi/v1/entity_types.py
JustKuzya: Until the end I was not sure if the third or fourth member would be needed for Errors and/or Logs. Neither was needed, but tuples would be easier to expand if I missed some use-cases later on.

src/dioptra/restapi/v1/entity_types.py
JustKuzya: The DB names seem to be set in stone (or rather "Philosophers Stone") of SQLAlchemy. Meanwhile the Textual representation is supposed to be human readable

src/dioptra/restapi/v1/entity_types.py
JustKuzya: Done. Really good thought. Except for one thing: I always prefer to use a common prefix, in this care REGEX_... for similar purpose constants - this way the code assist becomes more helpful as all similar use constants have same prefix and pop up as a suggestion, developer only needs to read the suggestions carefully and choose. I know that most of the developers are "writers", not "readers" ☺ , but still it is more helpful than not for the most cases.

src/dioptra/restapi/v1/entity_types.py
JustKuzya: Missed from the initial iteration.

src/dioptra/restapi/errors.py
hbooth: The type was passed in as a string value to provide context. I am not sure we should have two ways of providing this context. Do you have a suggestion how we can provide the proper context to indicate the properties that are (attempted) for sorting?

hbooth: after thinking about this a little bit, perhaps the better way to think about this, as not so much mapping to an entity but to a resource, with resource being the actual REST API Path we are mapping to. Therefore the mapping is not to an entity, but to the resource path.

src/dioptra/restapi/errors.py
hbooth: Instead of EntityType, this should be the REST resource on which this query parameter exists.

src/dioptra/restapi/errors.py
hbooth: agree with this. The context is the REST resource, not an entity resource.

src/dioptra/restapi/v1/artifacts/service.py
hbooth: agree with this.

src/dioptra/restapi/v1/entity_types.py
hbooth: I think we need to make a strong distinction between REST Resources and internal Entity Types which generally map to a base object. If there are other ways or object that were referred to as an entity type, we need to resolve that. @JustKuzya let me know if you see others and let's discuss.

src/dioptra/restapi/v1/entity_types.py
hbooth: the whole idea of the entity was to allow a "translation" between the entity type as it is in the DB, the names it references, and the human readable description. Having an Enum also will likely prove useful for the repository pattern allowing a Repository instance to have a reference to the Entity that it is providing.

src/dioptra/restapi/v1/shared/tags/service.py
hbooth: agree, with storing the enum reference as much as possible and only output to string when needed.

keithmanville:
OBE by #1240

PR #938, opened 2025-08-20
fix: capture EP param type mismatch

### LINKED ISSUE #891: [BUG] type mismatch in entrypoint validation causes uncaught ValueError
To reproduce, see the following Entrypoint graph yaml:

```
hello_step:
  hello:
    name: $name
goodbye_step:
  greet:
    greeting: Goodbye
    name: $name
shout_step:
  shout:
    greeting: $goodbye_step.greeting
```

where the `name` plugin task has the argument name registered as a `string`.

If you set the EntryPoint parameter to an `int` you will get a 500 error:

```
  File "/Users/kmanville/dev/dioptra/src/dioptra/restapi/v1/workflows/lib/type_coercions.py", line 70, in to_integer_type
    return int(x)
           ^^^^^^
ValueError: invalid literal for int() with base 10: 'World'
```

This should be caught and reported to the user.

jtsextonMITRE:
Closes #891 and #959

### DIFF
--- a/src/dioptra/restapi/errors.py
+++ b/src/dioptra/restapi/errors.py
@@ -320,6 +320,52 @@ class QueryParameterNotUniqueError(QueryParameterValidationError):
         super().__init__(type, "unique", **kwargs)
 
 
+class EntrypointParameterTypeMismatchError(DioptraError):
+    """Could not cast to the specified type"""
+
+    values: list[str | None]
+    parameter_names: list[str]
+    correct_types: list[str]
+
+    def __init__(
+        self,
+        values: list[str | None],
+        parameter_names: list[str],
+        correct_types: list[str],
+    ):
+        super().__init__(
+            f'Could not cast value(s) "{values}" of parameter(s) "{parameter_names}" to declared '
+            f'entrypoint parameter type "{correct_types}".'
+        )
+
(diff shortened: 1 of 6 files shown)


jtsextonMITRE:
Note:  Although we could have captured the error once in `coerce_to_type` I figured it would be more useful to the user to be able to see the parameter name as well, so instead I captured it twice in contexts where the parameter name is also available.

keithmanville (changes requested):
We should catch this earlier, before the job is executed. We can leave the the checks where you have them, but add the same check to the in the Job create service.

jtsextonMITRE:
> We should catch this earlier, before the job is executed. We can leave the the checks where you have them, but add the same check to the in the Job create service.
> 
> edit: per Harold a very similar check is happening for artifact parameters, so that might be a good point of reference.

I ended up doing this check a little differently because the artifact parameters don't have default values, and instead have declared types. Throwing an error for regular entrypoint parameters requires attempting to cast the variable, so I ended up making a specific function to do that. One instance of this check has a slightly different structure, so this function only gets 2/3 of the cases. Not sure whether it's better to just have it duplicated 3 times or to keep the `entrypointTypeMismatchError()` function.

PR #939, opened 2025-08-21
Job form artifacts

henrychoy:
Allow user to select artifact snapshot when creating job

### DIFF
--- a/src/frontend/src/views/CreateJob.vue
+++ b/src/frontend/src/views/CreateJob.vue
@@ -176,50 +176,136 @@
         </div>
       </fieldset>
     </div>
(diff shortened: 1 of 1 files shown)


keithmanville (changes requested):
This is working as expected for me.

As to the design of the snapshot selection: I think most of the time the user will want the latest snapshot, so it might be confusing / unintuitive, or just verbose to always show the snapshot drop down. What do you think about having a checkbox or similar element next to the resource drop down that when selected enables the selection of a specific snapshot?

We could use an icon to indicate browsing the history of the artifact with a tool tip explaining what selecting the checkbox does. Something like https://coreui.io/icons/linear/history/

It might also be helpful to be explicit about the latest snapshot by putting "(latest)" after the most recent snapshot ID in the dropdown. e.g. "Snapshot: 220 (latest)"

Issue #871, opened 2025-06-09
[BUG] 500 Internal Server Error when duplicate parameters exist in TOML


elyplant:
**Describe the bug**
An Internal Server Error with status code 500 is received when duplicate parameters exist under the same entrypoint in the TOML file. Should this error have be caught and reinterpreted?

**To Reproduce**
In the TOML file, under an entrypoint,
```
params = [
  { name = "image_size", type = "list", default_value = "[28, 28, 1]" },
  { name = "image_size", type = "list", default_value = "[28, 28, 1]" }
]
```
then attempt to upload

PR #941, opened 2025-08-25
feat(restapi): capture duplicate names in TOML file

### LINKED ISSUE #885: [BUG] Duplicate entrypoint names allowed in TOML file
**Describe the bug**
In the TOML file, when you create multiple entrypoints with the same name, uploading to Dioptra passes the TOML parser without a warning. This may potentially be caught later if `resolve_name_conflicts_strategy="overwrite"` is not set while importing resources but no warning is relayed when it is set.

**To Reproduce**
```
[[entrypoints]]
path = "entrypoints/hello-world.yaml"
name = "Hello World"
description = "A simple example using the hello_world plugin."
params = [
  { name = "name", type = "string", default_value = "World" }
]
plugins = [ "hello_world" ]

[[entrypoints]]
path = "entrypoints/hello-world.yaml"
name = "Hello World"
description = "A simple example using the hello_world plugin."
params = [
  { name = "name", type = "string", default_value = "World2" }
]
plugins = [ "hello_world" ]
```
then import the TOML

**Expected behavior**
The restapi should throw an error or maybe a warning that one of the entrypoints will not be considered when parsing the TOML.

jtsextonMITRE:
Closes #885 and #871

### DIFF
--- a/src/dioptra/restapi/v1/workflows/service.py
+++ b/src/dioptra/restapi/v1/workflows/service.py
@@ -478,6 +478,9 @@ class ResourceImportService(object):
         Returns:
             A dictionary mapping newly registered Plugin names to the ORM objects
         """
+        ResourceImportService._check_for_duplicate_names(
+            named_objects=plugins_config, named_object_type="plugin", key="path"
+        )
 
         param_types = param_types.copy()
         builtin_param_types = self._builtin_plugin_parameter_type_service.get(
@@ -578,6 +581,9 @@ class ResourceImportService(object):
         Returns:
             A dictionary mapping newly registered Entrypoint names to ORM object
         """
+        ResourceImportService._check_for_duplicate_names(
+            named_objects=entrypoints_config, named_object_type="entrypoint"
+        )
 
         param_types = param_types.copy()
         builtin_param_types = self._builtin_plugin_parameter_type_service.get(
@@ -588,6 +594,7 @@ class ResourceImportService(object):
         )
 
         entrypoints = {}
+
         for entrypoint in entrypoints_config:
             if overwrite:
                 existing = self._entrypoint_name_service.get(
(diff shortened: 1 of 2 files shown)


src/dioptra/restapi/v1/workflows/service.py
```diff
@@ -787,6 +815,21 @@ def _build_artifacts_params_list(
                 "is not defined in provided toml config.",
             ) from e
 
+    @staticmethod
+    def _duplicate_name_prevention(
+        resources: list[dict[str, Any]],
+        resource_type: str,
```
keithmanville: param names are confusing here, because (with the exception of entrypoint [edit: and param type]) everything else being passed here is not a Dioptra Resource.

src/dioptra/restapi/v1/workflows/service.py
```diff
@@ -588,6 +588,11 @@ def _register_entrypoints(
         )
 
         entrypoints = {}
+
+        ResourceImportService._duplicate_name_prevention(
```
keithmanville: Duplicate resources should already be caught when attempting to insert into the db with a 409 error.

However, it's probably best to catch duplicate resource names in the config separately like you're doing. This catches things earlier and differentiates between a clash with an existing resource and a clash within the import.

We should add a check for plugin.

src/dioptra/restapi/v1/workflows/service.py
```diff
@@ -787,6 +815,21 @@ def _build_artifacts_params_list(
...
+    @staticmethod
+    def _duplicate_name_prevention(
+        resources: list[dict[str, Any]],
+        resource_type: str,
+    ):
+        encountered_names = set()
+        for resource in resources:
+            if resource["name"] not in encountered_names:
+                encountered_names.add(resource["name"])
+            else:
```
keithmanville, line 830: instead of immediately raising an error, keep track of all duplicates and raise error with message containing full list.

keithmanville (changes requested):
Add new unit test to capture the cases this fix catches. I'm going to commit a starting point for you to expand on.

Issue #930, opened 2025-08-19
Replace mlflow metrics with a new Dioptra db table


keithmanville:
To remove dependence on mlflow, we should create a new database table to capture metrics associated with a job.

A job can have multiple metrics with the following properties:
- name (str)
- value (float)
- step (int)

The service layer should be updated to store this information int he new table instead of using the mlflow client.

PR #946, opened 2025-08-27
feat(restapi): replace mlflow metrics with database table

### LINKED ISSUE #810: [BUG] Metrics endpoint and client should be able to accept NaN

jtsextonMITRE:
Closes #810, #930

Supersedes PR #834


src/dioptra/restapi/db/alembic/versions/ad4f89b2288d_add_metrics_table.py
```diff
@@ -0,0 +1,31 @@
...
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = 'ad4f89b2288d'
+down_revision = '06e8176155bf'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
keithmanville, line 20: Since this replaces data previously stored in mlflow, we don't have a real migration path for that data. Is that something we are okay with? Do we need to document somewhere? Or is there some kind of reasonable solution here?

jtsextonMITRE: that's a good point, I don't have a good answer for this.. If we eventually remove mlflow as a dependency entirely, an upgrade script which relies on accessing data from it would break too, right?

keithmanville (commented):
Hopefully not expanding/changing the scope too much here: previously we were limited to how mlflow stored metrics, but now we are free to handle it however we want. Is there anything we should change?

- better handling of NaN (possibly resolving or making it easier to resolve #810)
- Any additional information we would want to capture with a metric?
- Does it make sense to associate metrics with other resources in addition to or instead of jobs? I'm thinking of artifacts or models here.

src/dioptra/restapi/db/alembic/versions/ad4f89b2288d_add_metrics_table.py
keithmanville: After discussion with James and Harold, we are going to leave this as a breaking change. If we get requests for a migration path for metrics data, we will prioritize the development a script that pulls data from mlflow and populates the new db table.

keithmanville:
> Hopefully not expanding/changing the scope too much here: previously we were limited to how mlflow stored metrics, but now we are free to handle it however we want. Is there anything we should change?
> 
>     * better handling of NaN (possibly resolving or making it easier to resolve #810)
> 
>     * Any additional information we would want to capture with a metric?
> 
>     * Does it make sense to associate metrics with other resources in addition to or instead of jobs? I'm thinking of artifacts or models here.

After discussion with Harold and James, we are leaving the scope of this issue as is.

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -677,15 +671,47 @@ def get(
...
+        job_metrics_dict: dict[str, models.JobMetric] = {}
+
+        # select metrics with the biggest timestamp for each metric name + step
+        for job_metric in job_metrics:
+            kept = job_metrics_dict.get(job_metric.name, None)
+
+            if kept is None or kept.timestamp < job_metric.timestamp:
+                job_metrics_dict[job_metric.name] = job_metric
 
-        return self._job_run_store.get_metrics(run_id)
```
keithmanville, line 686: Can we remove the now unused code related to metrics in `MlFlowJobRunStore`?

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -677,15 +671,47 @@ def get(
...
+                ),
+            )
+        )
+
+        # select metrics with the highest step value
+        job_metrics = list(db.session.scalars(job_metrics_stmt).unique().all())
+
+        job_metrics_dict: dict[str, models.JobMetric] = {}
+
+        # select metrics with the biggest timestamp for each metric name + step
```
keithmanville: To make sure I understand: a metric can have multiple values posted with the same step, so we need to return only the most recent metric at that latest step?

What benefit is there allowing multiple values for a metric at the same step? Should we instead overwrite the previous value?

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -701,14 +727,22 @@ def update(
...
         log.debug("Update job metrics by id", job_id=job_id)
 
-        run_id: str | None = self._job_id_mlflowrun_service.get(job_id=job_id, log=log)
-
-        if run_id is None:
-            raise JobMlflowRunNotSetError
+        job_dict = self._job_id_service.get(job_id, error_if_not_found=True)
+        # can't be None since error_if_not_found is True: the .get() call would
+        # error instead of returning None.
+        assert job_dict is not None
```
keithmanville: given the comment, can we get rid of this assert?

src/dioptra/restapi/v1/jobs/service.py
keithmanville: And could this be handled as part of the subquery with a sortby date?

keithmanville (changes requested):

src/dioptra/restapi/v1/jobs/service.py
jtsextonMITRE: > To make sure I understand: a metric can have multiple values posted with the same step, so we need to return only the most recent metric at that latest step?

Yes, correct, step is an optional parameter currently, so logging multiple things without specifying a step would send them all to step 0.

> What benefit is there allowing multiple values for a metric at the same step? Should we instead overwrite the previous value?

I lean towards preserving all logged metrics since not doing so is a downgrade from mlflow (and I don't want to force users to keep track of step value to preserve metrics unless they explicitly decide to). 

> And could this be handled as part of the subquery with a sortby date?

I was struggling to create a query with sqlalchemy that entirely avoided the need to iterate over the result. It didn't seem like I was able to add additional fields to the `select` (or `order by` after attempting it) which were not part of the `group_by` clause, and (I think) adding them to the `group by` clause messes with the results cause we want the max step for each name, not the max for each name+timestamp combo. I will play with this a bit more and see if there's a way to make a query to do a max for the timestamp after doing a max for the step (maybe it needs two separate subqueries?)

jtsextonMITRE: Seems like two subqueries let me do it properly!

src/dioptra/restapi/db/alembic/versions/ad4f89b2288d_add_metrics_table.py
```diff
@@ -0,0 +1,31 @@
...
+    op.create_table('job_metrics',
+        sa.Column('job_resource_id', sa.BigInteger().with_variant(sa.Integer(), 'sqlite'), nullable=False),
+        sa.Column('name', sa.Text(), nullable=False),
+        sa.Column('value', sa.Float(), nullable=True),
```
jkglasbrenner: We should use `sa.Double()` to make sure this is double precision on the databases that support it.

src/dioptra/restapi/db/db.py
```diff
@@ -59,6 +69,7 @@
 str255 = Annotated[str, mapped_column(String(255))]
 text_ = Annotated[str, mapped_column(Text())]
 bool_ = Annotated[bool, mapped_column(Boolean())]
+float_ = Annotated[Optional[float], mapped_column(Float(), nullable=True)]
```
jkglasbrenner: I would import and use `Double` rather than `Float`. Also, use the `optional` prefix to indicate the type is nullable. The Python-side type remains as `float`.

```python
# In imports
from sqlalchemy import Double

# In types list
optionaldouble_ = Annotated[Optional[float], mapped_column(Double(), nullable=True)]
```

src/dioptra/restapi/db/models/jobs.py
```diff
@@ -256,3 +257,30 @@ class JobLog(db.Model):  # type: ignore[name-defined]
...
+
+    # Database fields
+
+    job_resource_id: Mapped[bigint] = mapped_column(
+        ForeignKey("resources.resource_id"), init=False
+    )
+    name: Mapped[text_]
+    value: Mapped[float_]
+    step: Mapped[bigint]
+    timestamp: Mapped[bigint] = mapped_column(init=False, nullable=False)
```
jkglasbrenner: I think this is a carry-over from how MLflow handles timestamps for the metrics. I'd prefer us to use the `datetimetz` type instead of using integers, which is in line with what we are using on our other ORM models. This would require a change in the `MetricsSchema` and anything else that assumes this is an integer.

src/dioptra/restapi/db/models/jobs.py
```diff
@@ -256,3 +257,30 @@ class JobLog(db.Model):  # type: ignore[name-defined]
...
+
+    # Relationships
+    job_resource: Mapped["Resource"] = relationship()
+
+    # Additional settings
+    __table_args__ = (PrimaryKeyConstraint("job_resource_id", "name", "timestamp"),)
```
jkglasbrenner: I think the primary key should be
```suggestion
    __table_args__ = (PrimaryKeyConstraint("job_resource_id", "name", "step"),)
```

The issue with timestamps in this specific case is that their resolution can vary between platforms (milliseconds vs microseconds vs nanoseconds). Also, if we allow users to set the timestamp, then there's also time zones to worry about. The ordering may actually be more ambiguous this way. The steps, on the other hand, should correspond to some sort of loop or iteration taking place within a job, so the meaning of the step integer should be unambiguous.

src/dioptra/restapi/db/models/jobs.py
```diff
@@ -256,3 +257,30 @@ class JobLog(db.Model):  # type: ignore[name-defined]
...
+class JobMetric(db.Model):  # type: ignore[name-defined]
+    __tablename__ = "job_metrics"
+
+    # Database fields
+
+    job_resource_id: Mapped[bigint] = mapped_column(
+        ForeignKey("resources.resource_id"), init=False
+    )
+    name: Mapped[text_]
+    value: Mapped[float_]
```
jkglasbrenner: Replace with `optionaldouble_`, see my other comment on the `src/dioptra/restapi/db/db.py` file.

src/dioptra/restapi/db/alembic/versions/ad4f89b2288d_add_metrics_table.py
```diff
@@ -0,0 +1,31 @@
...
+        sa.Column('step', sa.BigInteger().with_variant(sa.Integer(), 'sqlite'), nullable=False),
+        sa.Column('timestamp', sa.BigInteger().with_variant(sa.Integer(), 'sqlite'), nullable=False),
+        sa.ForeignKeyConstraint(['job_resource_id'], ['resources.resource_id'], name=op.f('fk_job_metrics_job_resource_id_resources')),
+        sa.PrimaryKeyConstraint('job_resource_id', 'name', 'timestamp', name=op.f('pk_job_metrics'))
```
jkglasbrenner: Replace with

```suggestion
        sa.PrimaryKeyConstraint('job_resource_id', 'name', 'step', name=op.f('pk_job_metrics'))
```

See my comment on the `src/dioptra/restapi/db/models/jobs.py` file.

src/dioptra/restapi/db/models/jobs.py
```diff
@@ -256,3 +257,30 @@ class JobLog(db.Model):  # type: ignore[name-defined]
...
+
+    # Additional settings
+    __table_args__ = (PrimaryKeyConstraint("job_resource_id", "name", "timestamp"),)
+
+    # Initialize default values using dataclass __post_init__ method
+    # https://docs.python.org/3/library/dataclasses.html#dataclasses.__post_init__
+    def __post_init__(self) -> None:
+        import time
+
+        self.timestamp = int(time.time() * 1000.0)
```
jkglasbrenner: We should add this to the ORM definition to make it easy to check whether a `JobMetric` is the "latest" value. This should simplify the logic the in service layer too.

```python
    # Self-referential derived fields (read-only)
    @classmethod
    def __declare_last__(cls) -> None:
        job_metric_alias = aliased(cls)
        cls.is_latest = column_property(
            select(func.max(job_metric_alias.step))
            .where(
                job_metric_alias.job_resource_id == cls.job_resource_id,
                job_metric_alias.name == cls.name,
            )
            .correlate_except(job_metric_alias)
            .scalar_subquery()
            == cls.step
        )
```

You'll need to add these imports at the top of the file:

```python
from sqlalchemy import func
from sqlalchemy.orm import aliased
```

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -677,15 +671,65 @@ def get(
...
+                    models.JobMetric.name == job_metrics_max_timestamp_sub_query.c.name,
+                    models.JobMetric.step == job_metrics_max_timestamp_sub_query.c.step,
+                    models.JobMetric.timestamp
+                    == job_metrics_max_timestamp_sub_query.c.mtimestamp,
+                ),
+            )
+        )
+
+        # select metrics with the highest step value and highest timestamp for that step
+        job_metrics = list(db.session.scalars(job_metrics_stmt).unique().all())
```
jkglasbrenner: If you make the changes I suggest for the ORM file, then this can be simplified to the following:

```python
stmt = (
    select(models.JobMetric)
    .where(models.JobMetric.is_latest, models.JobMetric.job_resource_id == job_id)
)
job_metrics = list(db.session.scalars(stmt).all())
```

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -755,16 +805,26 @@ def get(
...
-            run_id=run_id,
-            name=metric_name,
-            page_index=page_index,
-            page_length=page_length,
+        job_metrics_stmt = select(models.JobMetric).where(
+            models.JobMetric.job_resource_id == job_id,
+            models.JobMetric.name == metric_name,
         )
 
+        history = list(db.session.scalars(job_metrics_stmt).unique().all())
```
jkglasbrenner, line 808: Do we need the `unique()`? Why not just

```suggestion
        history = list(db.session.scalars(job_metrics_stmt).all())
```

src/dioptra/restapi/db/models/jobs.py
jkglasbrenner: Whether we use `init=False` in tandem with the `__post_init__` method depends on whether we allow the user to optionally set the timestamp client-side, or if this will always be set by the API. MLflow's approach allowed users to set the timestamp if they wanted, which might be something we want to do as well (this would allow the option for metrics to track when they were produced locally, not when they arrived at the API). If they don't provide it, then we can use the server's time to set it. This would also require a change to the `MetricsSchema`, the service layer to handle it, as well as the client.

src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -677,15 +671,65 @@ def get(
...
+            }
+            for job_metric in job_metrics
+        ]
 
     def update(
         self,
         job_id: int,
         metric_name: str,
-        metric_value: float,
+        metric_value: float | None,
```
jkglasbrenner: This is specifically a comment about the `update` method as a whole, not this particular line.

In a side conversation with @keithmanville, I had mentioned that the primary key should be `(job_resource_id, name, step)` and that the `timestamp` field should not be included in the definition. He mentioned that since the name and step are chosen by the user, in an automated logging process it could be quite easy to POST to the same primary key more than once, and there are 3 ways to handle it:

> 1. reject the request to post the metric (probably bad because now the user has to handle this)
> 2. update the value for that timestep
> 3. auto increment the step and post the new value

If the `step` is one of the parameters that must be set, then 3 is surprising. I agree that 1 could break things unexpectedly and be frustrating as a user. Not going with 1 however breaks our typical API patterns (POST always means create something in the database, not edit it). But, at the same time, `JobMetric` is _not_ a `Resource`, so if we deviate from our patterns here (which is probably the correct choice), that's the justification. This leaves us with 2 as the most sensible option.

To incorporate 2, we would need to update the logic for this, as right now the method only creates new `JobMetric` entities and does not check if one already exists and uses it if it does.

src/dioptra/restapi/db/alembic/versions/ad4f89b2288d_add_metrics_table.py
```diff
@@ -0,0 +1,31 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.create_table('job_metrics',
+        sa.Column('job_resource_id', sa.BigInteger().with_variant(sa.Integer(), 'sqlite'), nullable=False),
+        sa.Column('name', sa.Text(), nullable=False),
+        sa.Column('value', sa.Float(), nullable=True),
+        sa.Column('step', sa.BigInteger().with_variant(sa.Integer(), 'sqlite'), nullable=False),
+        sa.Column('timestamp', sa.BigInteger().with_variant(sa.Integer(), 'sqlite'), nullable=False),
```
jkglasbrenner: This would need to be changed when changing this over to a datetime type rather than an integer type, see my comment in file `src/dioptra/restapi/db/models/jobs.py`

jkglasbrenner (changes requested):
I gave this a thorough review and left a number of comments. @keithmanville  will probably want to take a look at them at least.

It's not in the scope of this PR, but I'm also thinking that, if we're going to make this level of a breaking change, then we should use this as an opportunity to streamline the API URLs for the metrics. Having API URLs like `"/<int:id>/metrics/<string:name>/snapshots"` resembles what we do for resources, but since the metrics are not resources, this feels verbose and complicated. The metrics history is not what I'd call snapshots. I think we might be able to get away with just GET and POST methods on `"/<int:id>/metrics"` if we add some additional query parameters to the GET method (which, because it's not a resource, does not need to follow the same patterns).

src/dioptra/restapi/db/models/jobs.py
jkglasbrenner: I know this touches on the conversation that started here: https://github.com/usnistgov/dioptra/pull/946#discussion_r2319775759

It's true that MLflow is less strict with their metrics requirements, and that they decided to make this the primary key:

    (key, value, timestamp, run_uuid, step, is_nan)

where `run_uuid` is basically equivalent to our `job_resource_id` and we are supporting `nan` by allowing `value` to be nullable and for use, NULL = nan, so we drop `is_nan`. Even then, I'm not sure I like their choice. This allows for several weird things to happen in the database:

1. You can log multiple metrics to the same `(run_uuid, key, step)` tuple under different timestamps.
2. Even stranger, you can log different values to the same `(run_uuid, key, step, timestamp)` tuple.
3. Even stranger than that, you could log two versions of the `(run_uuid, key, step, timestamp, value)` tuple, one where `is_nan = False` and one where `is_nan = True`.

My guess is they chose this for maximum flexibility, and the user can decide what any of this means. But that also means that, if someone fails to document or capture how to interpret the above scenarios, then what any of it means is lost.

With Dioptra, we're trying to provide guardrails that avoid these situations because it hurts auditing and review, which is why I am arguing we make things much stricter and go with just `(job_resource_id, name, step)` as the primary key.

jkglasbrenner: Copying a couple other comments I made to @keithmanville in a side conversation that I think are relevant to this topic:

> Metrics are point estimates, so if you've subdivided a loop into subloops, and each subloop collects data, then some kind of aggregate/reduce operation should happen at the end to produce the metric before it is logged. Metrics should be insensitive to parallelization choices (assuming the math is deterministic, I should get the same results regardless of whether I use 1, 2, 4, 8 cores or use a GPU, within machine precision).

Also about why we want to be strict about what "step" means.

> If there's a heuristic rule that's kind of emerging as I type this out, it's that the definition of a "step" should be defined at the job level, so that step = 1, 2, 3 for metrics A, B, and C all correspond to values that were collected at the same internal "step" of the entrypoint (here I mean the same internal loop, such as a training loop, not necessarily a very specific "step" in the entrypoint YAML). This indicates a way to determine separation boundaries for entrypoints, if you feel like you need two different step counters in your job to capture everything, maybe because you're measuring stuff in two different types of loops, then that probably means you're doing too much in a single entrypoint and should break them apart.

jkglasbrenner:
Capturing this here because it took a while to track it down:

This is how MLflow sanitizes `nan`, `-Inf`, and `+Inf` before sending it to a SQLAlchemy backend: https://github.com/mlflow/mlflow/blob/a04c61ce588be4ffe3d5389cbfb1f85a39380ebb/mlflow/store/tracking/sqlalchemy_store.py#L805-L820

They handle `nan` by storing the value as 0 and also storing an `is_nan=True` boolean in the database. Alternatively, if they have either `-Inf` or `+Inf` as the value, they store the value as the smallest or largest possible signed 64-bit float.

src/dioptra/restapi/db/models/jobs.py
jkglasbrenner: To close this out, timestamp can now be set by the user.

jkglasbrenner:
Also, once we merge this PR, we can close #834 since this solution also addresses #810 and hence that PR is no longer needed.

PR #954, opened 2025-09-04
et-594: Build and sign container images

### LINKED ISSUE #594: Implement our own GitHub runners to enable signing of docker images using cosign

etrapnell-nist:
This PR re-enables automatic building of non-GPU container images. These images are signed and registered with Cosign/Rekor for signature verification. Closes #594


docs/source/how-to/build-container-images.rst
```diff
@@ -15,10 +15,10 @@
...
 
-.. _getting-started-building-the-containers:
+.. _how-to-build-container-images:
 
-Building the Containers
-=======================
+Building the Container Images
+=============================
 
 .. include:: /_glossary_note.rst
```
keithmanville: Can we add something here that makes it clear downloading the pre-built containers is the easiest path? This is only needed if you want to customize something, need the gpu containers, or have some other requirement to build from scratch. I think it still reads as if building yourself is the only way. I know it says this in the guide where this how-to is linked, but would be good to repeat here since you could find the how-to from the navigation bar.

keithmanville (dismissed):

docs/source/how-to/build-container-images.rst
etrapnell-nist: Good idea. I added some language to explain that downloading is easier, and why one might want to build the images.

jkglasbrenner (commented):
As a status update on where things stand with this branch and PR, I've gone through and refactored the `.github/workflows/docker-images.yml` file by encapsulating certain behaviors into custom GitHub Actions. I extracted the more complicated, script-like steps from the workflow into 3 distinct custom actions:

- `install-rekor-cli` - This generalizes the logic for downloading and installing the `rekor-cli` binary by following the same patterns used in the `sigstore/cosign-installer@v3.9.2` action. Installation now works on a wider variety of operating systems and CPU architectures.
- `sign-payload-fips` - This encapsulates the FIPS-compliant logic used to sign the cosign payloads. The action is set to spin up and prepare the `registry.access.redhat.com/ubi8/ubi:8.6` Docker image automatically using a `.github/actions/sign-payload-fips/Dockerfile`. The logic of the action itself was migrated into the bash script `.github/actions/sign-payload-fips/entrypoint.sh`, which now follows the same patterns and practices used in our other bash scripts. The main changes are that the payload and keys are passed to the script as script arguments rather than through files (the secrets are obscured in the logs).
- `create-rekor-entry` - This encapsulates the logic used to register signed cosign payloads to the Rekor transparency log. The action is set to spin up and prepare the `registry.access.redhat.com/ubi8/ubi:8.6` Docker image automatically using a `.github/actions/create-rekor-entry/Dockerfile`. The logic of the action itself was migrated into the bash script `.github/actions/create-rekor-entry/entrypoint.sh`, which now follows the same patterns and practices used in our other bash scripts. The main changes are that the payload, base64-encoded signature, and public key are passed to the script as script arguments rather than through files (the secrets are obscured in the logs).

As I was extracting the logic into these modules, I noticed the following things:

The variable `IMAGE=ghcr.io/${{ steps.extract_tag.outputs.image_tag }}` was used as the Docker image tag in all the steps in the docker-images workflow related to cosign, openssl, and rekor. When expanded, this becomes `ghcr.io/ghcr.io/usnistgov/dioptra/nginx:pr-954`, so the `ghcr.io` prefix appears twice, which doesn't match how you would pull and use the containers in the GitHub Package Registry (see https://github.com/usnistgov/dioptra/pkgs/container/dioptra%2Fnginx/533123945?tag=pr-954). Is there documentation somewhere that says you ought to do this? If not, the fact that it seemed to work, meaning a signature was being generated and attached, seems to be an accident.

It isn't clear where the logic for posting the signed payload to the Rekor server comes from. It's using `curl` and assembling the payload manually instead of using `rekor-cli` as the documentation recommends: https://docs.sigstore.dev/logging/cli/. Also, I wasn't able to find any documentation that says this is what payloads `POST`ed to <https://rekor.sigstore.dev/api/v1/log/entries> should look like: 

```json
{
  "apiVersion": "0.0.1",
  "kind": "hashedrekord",
  "spec": {
      "data": {
          "hash": {
              "algorithm": "sha256",
              "value": "${payload_checksum}"
          }
      },
      "signature": {
          "content": "${_arg_signature}",
          "publicKey": {
              "content": "${public_key_b64}"
          }
      }
  }
}
```

Note that the above JSON is using variables found in `.github/actions/create-rekor-entry/entrypoint.sh`, but it should be equivalent to what was being done directly in `docker-images.yml` before the refactor. Where did the above JSON come from? And, as part of that, why does the logic have the public key stored in `${{ secrets.DIOPTRA_PUBLIC_KEY }}` being base64 encoded? Isn't this base64 encoding the key twice? I thought it was already encoded (the `--armor` export option) since you have to provide plain text when you declare GitHub secrets variables.

After finishing the refactor, which included removing the additional `ghcr.io` prefix, the "cosign verify" step started to fail with the following message:

```
Error: no matching signatures: searching log query: [POST /api/v1/log/entries/retrieve][400] searchLogQueryBadRequest {"code":400,"message":"verifying signature: crypto/rsa: verification error"}

error during command execution: no matching signatures: searching log query: [POST /api/v1/log/entries/retrieve][400] searchLogQueryBadRequest {"code":400,"message":"verifying signature: crypto/rsa: verification error"}
```

So something in the validation chain is failing related to the signature, but it isn't clear what. The OpenSSL verification passes, so it's either something that gets messed up when passing data through the `$GITHUB_OUTPUT` mechanism, it's something that was being obscured by having a second `ghcr.io/` prefix, or maybe it's the possible double-base64 encoding I mentioned? At the moment it isn't clear to me which it is.

PR #964, opened 2025-09-12
fix: handle temporary import issue

### LINKED ISSUE #963: [BUG] importing tensorflow in an artifact plugin and task plugin crashes the job
Importing tensorflow in an artifact plugin and a task plugin crashes the job. This potentially affects other libraries with bindings to other languages.

```
/Users/kmanville/dev/dioptra/.venv/lib/python3.11/site-packages/tensorflow/python/framework/ops.py:29: UserWarning: The NumPy module was reloaded (imported a second time). This can in some cases result in small but subtle issues and is discouraged.
  import numpy as np
2025-09-12 15:10:06.533503: F tensorflow/python/lib/core/py_exception_registry.cc:28] Check failed: singleton_ == nullptr PyExceptionRegistry::Init() already called
[1]    18692 abort      ipython
```

Here is a minimal example to reproduce the error:
```python
import sys
original_modules = sys.modules.copy()
import tensorflow
for name in list(sys.modules.keys()):
    if name not in original_modules:
        del sys.modules[name]
import tensorflow
```

The source of the issue is that we need to import specific versions of artifact tasks, which may have name clashes with other versions of the task. To solve this, we import the artifact task module and then attempt to clean it up. If a dependency of that plugin task has some state that is not handled properly by our attempts to unload the module, it can cause problems if the same package is loaded in a task plugin.

See:
https://github.com/usnistgov/dioptra/blob/70af33728ffc3e0777c0c6e1647c3b9d9d8e8962/src/dioptra/task_engine/task_engine.py#L543
and:
https://github.com/usnistgov/dioptra/blob/70af33728ffc3e0777c0c6e1647c3b9d9d8e8962/src/dioptra/sdk/utilities/contexts/_import_temp.py#L23

Proposed solution:
Instead of attempting to cleanup all modules that weren't loaded before the artifact task was imported, only cleanup the artifact task module itself.

keithmanville:
closes #963

### DIFF
--- a/src/dioptra/sdk/utilities/contexts/_import_temp.py
+++ b/src/dioptra/sdk/utilities/contexts/_import_temp.py
@@ -25,8 +25,6 @@ def import_temp(package_name: str, path: Path):
     """
     Context manager to temporarily add a module
     """
-    original_modules = sys.modules.copy()
-
     sys.path.insert(0, path.resolve().as_posix())
     module = importlib.import_module(package_name)
     try:
@@ -34,5 +32,5 @@ def import_temp(package_name: str, path: Path):
     finally:
         sys.path.pop(0)
         for name in list(sys.modules.keys()):
-            if name not in original_modules:
+            if name.startswith(f"{package_name}."):
                 del sys.modules[name]


src/dioptra/sdk/utilities/contexts/_import_temp.py
```diff
@@ -25,14 +25,12 @@ def import_temp(package_name: str, path: Path):
...
-
     sys.path.insert(0, path.resolve().as_posix())
     module = importlib.import_module(package_name)
     try:
         yield module
     finally:
         sys.path.pop(0)
         for name in list(sys.modules.keys()):
-            if name not in original_modules:
+            if name.startswith(package_name + "."):
```
jkglasbrenner: Very minor, prefer f-strings to addition operator:

```suggestion
            if name.startswith(f"{package_name}."):
```

Issue #957, opened 2025-09-05
Update GET /artifacts endpoint to support querying by output parms


keithmanville:
We want to be able to search for all artifacts that are a valid parameter to an entrypoint.

To do this we need to update the `GET /artifacts` endpoint with a new `outputParams` query parameter. The value of the new parameter should be a list of plugin param type IDs.

The service layer needs to be updated to include this as part of the query. The output type of the artifact must match the IDs exactly.

This should be treated as an AND with the other query parameters.

PR #976, opened 2025-09-22
feat: added ability to search for artifacts based on output parameters

hbooth:
Closes Issue #957
Adds ability to search for artifacts based on the output parameters of the artifact task associated with the Artifact.
The list is assumed to be ordered. The new query parameter is called outputParams and is a comma separated list of parameter type ids.

### DIFF
--- a/src/dioptra/restapi/v1/artifacts/schema.py
+++ b/src/dioptra/restapi/v1/artifacts/schema.py
@@ -16,7 +16,12 @@
 # https://creativecommons.org/licenses/by/4.0/legalcode
 """The schemas for serializing/deserializing Artifact resources."""
 
+from typing import List
+
+import structlog
 from marshmallow import Schema, fields
+from marshmallow.exceptions import ValidationError
+from structlog.stdlib import BoundLogger
 
 from dioptra.restapi.utils import validate_artifact_url
 from dioptra.restapi.v1.file_types import FileTypes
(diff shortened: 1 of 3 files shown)


src/dioptra/restapi/v1/artifacts/schema.py
```diff
@@ -206,10 +214,42 @@ class ArtifactContentsGetQueryParameters(Schema):
...
+
 class ArtifactGetQueryParameters(
     PagingQueryParametersSchema,
     GroupIdQueryParametersSchema,
     SearchQueryParametersSchema,
     SortByGetQueryParametersSchema,
 ):
     """The query parameters for the GET method of the /artifacts endpoint."""
+
+    outputParams = DelimitedIntegerList(
```
keithmanville, line 246: Wasn't sure if there is a standard way of handling multiple values for a query param. According to the link below there isn't, so I'm fine with the comma delimited approach, especially since it is limited to a list of ints.

https://stackoverflow.com/questions/24059773/correct-way-to-pass-multiple-values-for-same-parameter-name-in-get-request

hbooth: I chose this way because I wanted to insure the order. The other way, supplying the same parameter name multiple times with different values, would have potentially consisted of the same parameter being supplied twice with no guarantees on ordering.

Issue #971, opened 2025-09-16
UI Changes


keithmanville:
- [ ] Job Create Page (artifact param snapshot dropdown:
- add a small down-right arrow as a visual cue that the snapshot selection is related to the artifact resource.
- add the snapshot create_on date to the top right.

- [ ] Artifacts Page Changes
- remove artifactUri from UI
- remove artifactUri from response (separate issue)
- make artifact task, output params, and plugin one row
  - choose plugin first, choose (ANY) task, then display output params
- sort artifact table by job ID descending
- update to show all dates
  - resource created on
  - resource last modified on
  - snapshot timestamp

- [ ] Job Create Page Changes:
- Params -> Parameters
- Check this throughout the UI

- [ ] Job Dashboard Changes:
- Parameters -> Entrypoint Parameters
- Artifacts -> Artifact Parameters
- Artifacts Created -> Artifact Outputs
- fix Job dashboard

- [ ] Entrypoint Create / Edit Page:
- make adding entrypoint and artifact params work as adding a new row to the tables.
- make entrypoint and artifact params tables side-by-side
- no paging in param tables

- [ ] PluginFile Edit Page:
- for function and artifact tasks: add param by adding row in table

- [ ] Plugin Edit page:
- clicking a row goes to plugin edit screen
- add details table for metadatab (name and description) that is editable (with pencil icon)
- add save button by metadata table inside the same fieldset

- [ ] Experiment Edit page
- follow same pattern as plugins
- remove jobs link
- metadata table at top that is saveable
- Save button always inside of fieldset with metadata table

- [ ] Models:
- Hide the whole page

- [ ] Home Page:
- Add overview dashboard (@JustKuzya  to propose idea)

- [ ] Nav Bar:
- Reorder: 
  - Experiments
  - Jobs
  - Artifacts
  - Entrypoints
  - Plugins
  - Plugins Parameters
  - Queues
  - Tags

- [ ] All Resources table pages:
- remove expand from tables
- always go to full edit page (no dialogs)
- Always make the name a blue link

- [ ] All edit pages:
- make save button active when changes

PR #977, opened 2025-09-22
feat: clean up job and artifact pages

henrychoy:
Closes first 4 checkboxes of #971.  

Job and artifact page cleanup


keithmanville (changes requested):
Job Create Page looks good

Artifacts Page:
- for "make artifact task, output params, and plugin one row," can you make it so the plugin chip is part of the plugin selection drop down? This would make it so it's no longer select *other* plugin but the selected plugin is part of the dropdown itself. This would be more consistent with how the task dropdown works (and I think how dropdowns in general work).
- I think you missed "update to show all dates"

For the params -> parameters change, I see "output params" in the table header of the Artifact Parameters table on the Entrypoint page

Job Dashboard:
- I know we said the tab should be "Artifact Outputs" but I think "Output Artifacts" actually makes more sense.
- I forgot what "fix Job dashboard" meant. Can you remind me?

PR #978, opened 2025-09-23
docs: add how-to's for creating users/queues/plugins

jtsextonMITRE:
Addresses #1037, #1058, #1038, #1043, #1053, #1041

### DIFF
--- a/docs/source/how-to/essential-workflows/creating-a-queue.rst
+++ b/docs/source/how-to/essential-workflows/creating-a-queue.rst
@@ -57,7 +57,7 @@ Register a queue for a specific group, with a name and a description.
 
    .. group-tab:: GUI
 
-      In the Dioptra GUI, navigate to the **Queues tab**. Click **Create**. Enter a *name* and *description*, select a *group* for the queue, then click **Confirm**.
+      In the Dioptra GUI, navigate to the **Queues** tab. Click **Create**. Enter a *name* and *description*, select a *group* for the queue, then click **Confirm**.
 
    .. group-tab:: Python Client
 
--- a/docs/source/how-to/essential-workflows/creating-param-types.rst
+++ b/docs/source/how-to/essential-workflows/creating-param-types.rst
@@ -56,7 +56,7 @@ Register a parameter type for use in dioptra.
 
    .. group-tab:: GUI
 
-      In the Dioptra GUI, navigate to the **Plugin Parameters tab**. Click **Create**. 
+      In the Dioptra GUI, navigate to the **Plugin Parameters** tab. Click **Create**. 
       
       Enter a *name*, *description*, select a *group* for the type, enter a *structure* if applicable, then click **Confirm**.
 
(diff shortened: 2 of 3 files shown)


docs/source/how-to/code/pyplug_annotation.py
```diff
@@ -0,0 +1 @@
+@pyplugs.register
```
keithmanville, line 1: Should the python code for this how-to simply be included inline? I know we discussed storing code separately for the tutorial, but I don't know that we want to do the same thing here.

I think the reasoning for the tutorial was to both display the code (in docs or in jupyter notebook) as well as make it useable (via import functionality) without having to maintain the code in two places. But in this how-to most of these files aren't actually stand-alone working pieces that we expect the user to make use of.

docs/source/how-to/creating-a-user.rst
```diff
@@ -0,0 +1,47 @@
...
+
+      Dioptra user login page
+
+2. On the presented login screen, click **Signup**.
+3. Enter a username, email, and password in the appropriate fields.
+
+   .. figure:: /images/creating-a-user-registration.png
+      :alt: Figure depicting the Dioptra user registration page
+       
+      Dioptra user registration page
```
keithmanville: I would probably avoid using "administrator" as a username, since we don't actually have the concept of admin accounts. Maybe just `dioptra-user` ?

docs/source/how-to/creating-a-plugin.rst
```diff
@@ -0,0 +1,178 @@
...
+1. Click the **Plugin Parameters** tab. 
+
+   .. figure:: /images/plugin-param.png
+      :alt: Figure depicting the Dioptra Plugin Parameters page
+    
+      Dioptra Plugin Parameters page
+
+2. Click **Create**.
+3. Enter a name, select the group that should have access to the type, and enter a description.
+4. If your type is a structure consisting of mappings, lists, and you would like to represent that,
```
keithmanville: link to structured types section of task engine reference doc.

docs/source/how-to/creating-a-queue.rst
```diff
@@ -0,0 +1,55 @@
...
+
+
+.. _how_to_create_a_queue:
+
+How-To: Create a Queue
+======================
+
+In Dioptra, jobs are submitted to queues, which represent a computing environment. This
+guide assumes that Dioptra is already installed and is accessible through the web browser,
+and that the user is logged in.
```
keithmanville: also assumes a worker container is running / available as part of the deployment.

docs/source/how-to/creating-a-queue.rst
```diff
@@ -0,0 +1,55 @@
...
+
+   .. figure:: /images/queues-page.png
+      :alt: Figure depicting the Dioptra queues page
+       
+      Dioptra queues page
+
+2. On the Queues page, click **Create**.
+3. Enter a name for the queue, select the group that should have access to the queue,
+   and a description in the appropriate fields. *Note: the name should match the worker 
+   the queue is associated with.*
```
keithmanville: If we have some other documentation (maybe another how-to) on creating a worker, link to it here.

PR #983, opened 2025-09-26
MNIST example

keithmanville:


colton-lapp:
One thing that I think would make the MNIST notebook really interesting is if you found a couple changes to the experiment that yielded different conclusions, and let the user switch a couple hardcoded constants up top, re-run the notebook from top to bottom and observe the changes. 

For example, it looks like the defense mechanisms don't help much against FGM and the patch attacks. What if the result was different for a different model architecture? Or a different dataset? Setting:
```python
DATASET='mnist' # options: mnist, fruits, etc
MODEL = 'resnet' # options: resnet, lenet, etc
```

at the top of the notebook and having those changes be propagated throughout the entry points would be pretty neat in my opinion and shows the power of running things in the Jupyter notebook as opposed to the UI - a kind of "meta workflow" that can be parameterized.

examples/mnist_demo.ipynb
```diff
@@ -0,0 +1,776 @@
...
+    "\n",
+    "def find_artifact_by_name(artifact_name: str, job: dict[str, Any]) -> dict[str, Any]:\n",
+    "    for job_artifact in job[\"artifacts\"]:\n",
+    "        artifact = client.artifacts.get_by_id(job_artifact[\"id\"])\n",
+    "        if artifact_name in artifact[\"description\"]:\n",
+    "            return {\n",
+    "                \"id\": artifact[\"id\"],\n",
+    "                \"snapshotId\": artifact[\"snapshot\"],\n",
+    "            }\n",
+    "    raise Exception(\"Could not retrieve artifact\")\n",
```
colton-lapp, line 144: These helper functions are nice because they make the ensuing code blocks short and clean. They do, however, make it a little harder to understand what the client is capable of and how to use it. When these helper functions are called throughout the notebook, I found myself scrolling up to try and understand "what is actually happening here and how much of this is automatically completed by the client". 

One potential change could be to introduce these helper functions throughout the notebook when they are called for the first time. It's less organized because they're all not in one place, but it would help people understand exactly what  additional code needs to be provided to interact with the client. Feel free to reject this idea - I think it depends on how much you want people to understand how to use the Python client from this and how much you just want it to be a clean demonstration of Dioptra.

examples/mnist_demo.ipynb
```diff
@@ -0,0 +1,776 @@
...
+    "metrics[\"fgm_jpeg\"] = get_metrics(jobs[\"fgm_jpeg\"])\n",
+    "jobs[\"fgm_gaussian\"] = run_defense_job(artifacts[\"fgm_data\"], artifacts[\"model\"], \"gaussian_augmentation\")\n",
+    "metrics[\"fgm_gaussian\"] = get_metrics(jobs[\"fgm_gaussian\"])\n",
+    "jobs[\"patch_smoothing\"] = run_defense_job(artifacts[\"patch_data\"], artifacts[\"model\"], \"spatial_smoothing\")\n",
+    "metrics[\"patch_smoothing\"] = get_metrics(jobs[\"patch_smoothing\"])\n",
+    "jobs[\"patch_jpg\"] = run_defense_job(artifacts[\"patch_data\"], artifacts[\"model\"], \"jpeg_compression\")\n",
+    "metrics[\"patch_jpg\"] = get_metrics(jobs[\"patch_jpg\"])\n",
+    "jobs[\"patch_gaussian\"] = run_defense_job(artifacts[\"patch_data\"], artifacts[\"model\"], \"gaussian_augmentation\")\n",
+    "metrics[\"patch_gaussian\"] = get_metrics(jobs[\"patch_gaussian\"])"
+   ]
```
colton-lapp, line 722: Up until this code block, we are used to seeing an invocation of the client as a multi-line Python block. With the introduction of the helper function above, you are now able to create a bunch of jobs with single line commands. 

This confused me a bit at first, and I was wondering "where in the notebook do they try out the various combinations of defenses against attacks?" Adding some comments / breaking up this cell could make that a little more clear while still using the helper function to avoid repeating the same kind of client calls.

colton-lapp:
I was a little confused when I first saw the image of the adversarial patch. If it had a title and/or was smaller, that would help me understand it's relationship to the MNIST images.

examples/mnist_demo.ipynb
jtsextonMITRE: I'm a bit torn on this, I agree it does teach the client better if we introduce the helper functions one at a time, but at the same time, in other projects we've often started with this notebook and removed everything after the helper functions. So having them in one place lets you sort of quickly create new examples based on MNIST and keep the useful functionality in them.

jkglasbrenner: My thought is that whether or not we motivate and explain the utility functions depends on the purpose of the notebook. If this is already a tutorial that is focused on a realistic end-to-end example, then also introducing and motivating the utility functions a priori can potentially distract from the overall message. In this case, I would recommend that we instead capture the idea in a separate "How To" that is just focused on building wrapper utilities for orchestrating Python client calls, that way users can have more digestable documentation rather than huge, multi-purpose notebooks/documents.

Potential "How To": How To Perform a Parameter Sweep in Dioptra

keithmanville:
closing draft request in favor of #997

PR #986, opened 2025-10-03
feat: updated experiment page

henrychoy:
Closes experiment edit page of #971


jkglasbrenner (approved):
Just checked the fix for routing the user to the correct page after creating a job from the experiments page. It now works as I would expect.

LGTM

PR #997, opened 2025-10-22
Dioptra OPTIC plugin

keithmanville:


extra/plugins/dioptra_optic/data.py
```diff
@@ -0,0 +1,218 @@
...
+            ds = ds.cache()
+            ds = ds.shuffle(
+                builder.info.splits[name].num_examples // 10,
+                seed=seed,
+                reshuffle_each_iteration=True,
+            )
+            ds = ds.batch(batch_size)
+        else:
+            ds = ds.batch(batch_size)
+            ds = ds.cache()
```
jkglasbrenner, line 210: Dataset should be cached before batching:

```suggestion
            ds = ds.cache()
            ds = ds.batch(batch_size)
```

extra/plugins/dioptra_optic/models.py
```diff
@@ -0,0 +1,465 @@
...
+        - :py:mod:`keras.metrics`
+        - :py:class:`keras.Model`
+    """
+    model_options = model_options or {}
+    loss_options = loss_options or {}
+    optimizer_options = optimizer_options or {"learning_rate": learning_rate}
+    metrics = metrics or []
+
+    input_shape = next(iter(dataset))[0].shape[-3:]
+    batch_size, num_classes = next(iter(dataset))[1].shape
```
jkglasbrenner: These lines are why you see warnings in the logs that look like this:

```text
tensorflow/core/kernels/data/cache_dataset_ops.cc:858] The calling iterator did not fully read the dataset being cached. In order to avoid unexpected truncation of the dataset, the partially cached contents of the dataset  will be discarded. This can happen if you have an input pipeline similar to `dataset.cache().take(k).repeat()`. You should use `dataset.take(k).cache().repeat()` instead.
```

The prefetched dataset has specs attached that let you get this information without needing to consume anything from the dataset iterator:

```suggestion
    input_shape = [int(x) for x in dataset.element_spec[0].shape.as_list()[1:]]
    num_classes = int(dataset.element_spec[1].shape.as_list()[1])
```

examples/scripts/download_data.py
```diff
@@ -1,1130 +1,51 @@
...
+@click.argument("datasets", nargs=-1)
+@click.option("--data-dir", envvar=["DIOPTRA_DATASETS_DIR", "TFDS_DATA_DIR"], help="datasets_dir")
+@click.option("--dryrun", is_flag=True, help="datasets_dir")
+def download(datasets: tuple[str, ...], data_dir: str | None, dryrun: bool):
+    print(f"saving dataset(s) to {data_dir or '[tfds default location]'}")
+    for dataset in datasets:
+        print(f"  -- downloading and preparing {dataset}")
+        if not dryrun:
+            builder = tfds.builder(dataset)
+            builder.download_and_prepare(download_dir=data_dir)
```
jkglasbrenner: The `download_dir` seems to be only be for intermediate files, setting `data_dir` on the `tfds.builder` looks like it controls where the final outputs of the download go (and are what you need to load in the plugins).

```suggestion
            builder = tfds.builder(dataset, data_dir=data_dir)
            builder.download_and_prepare()
```

extra/plugins/dioptra_optic/models.py
```diff
@@ -0,0 +1,465 @@
...
+
+    Args:
+        module: The Python module that has a `preprocess_input` function and `model_name` function.
+        model_name: The model name matching the function name in `module` that initalizes the model.
+    Returns:
+        A Callable whose signature matches matches ModelFn(input_shape, classes, **model_kwargs).
+
+    """
+    model_fn = getattr(module, model_name)
+    preprocess_fn = getattr(module, "preprocess_input")
```
jkglasbrenner: Ruff seems to prefer this:

```suggestion
    preprocess_fn = module.preprocess_input
```

extra/plugins/dioptra_optic/data.py
```diff
@@ -0,0 +1,218 @@
...
+    for name, ds in dataset.items():
+        ds = ds.map(one_hot, num_parallel_calls=tf.data.AUTOTUNE)
+        if normalize_val is not None:
+            ds = ds.map(normalize, num_parallel_calls=tf.data.AUTOTUNE)
+        if image_size is not None:
+            ds = ds.map(resize, num_parallel_calls=tf.data.AUTOTUNE)
+        if name == "train":
+            ds = ds.cache()
+            ds = ds.shuffle(
+                builder.info.splits[name].num_examples // 10,
```
jkglasbrenner: I'm not sure I understand what this does. What is `num_examples`, why are you dividing it by 10, and what does it have to do with dataset shuffling?

extra/plugins/dioptra_optic/data.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+@pyplugs.register
+def load_dataset_from_tfds(
+    name: str,
+    split: dict[Literal["train", "val", "test"], str] = {
+        "train": "train[:80%]",
+        "val": "train[80%:]",
+        "test": "test",
+    },
```
jkglasbrenner: Mutable data structures shouldn't be used as default arguments, see <https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments> for why this can lead to unexpected behavior.

Instead, we would want something like:

```suggestion
    split: dict[Literal["train", "val", "test"], str] | None = None,
```

And in the body of the function:

```python
if split is None:
    split = {
        "train": "train[:80%]",
        "val": "train[80%:]",
        "test": "test",
    }
```

extra/plugins/dioptra_optic/data.py
```diff
@@ -0,0 +1,218 @@
...
+
+    Returns:
+        A tuple of train, val, test tf.data.Dataset objects.
+    """
+    # get a dictionary of dataset splits from the builder
+    dataset: dict = builder.as_dataset(split=split, as_supervised=True)
+
+    def normalize(x, y):
+        """Normalizes images: `uint8` -> `float32`."""
+        return tf.cast(x, tf.float32) / normalize_val, y
```
jkglasbrenner: Explicitly cast all non-Tensor values to desired datatype before running division op:

```suggestion
        return tf.cast(x, tf.float32) / tf.cast(normalize_val, tf.float32), y
```

extra/plugins/dioptra_optic/data.py
```diff
@@ -0,0 +1,218 @@
...
+    return (
+        dataset.get("train", None),
+        dataset.get("val", None),
+        dataset.get("test", None),
+    )
+
+
+def create_transformed_dataset(
+    dataset: tf.data.Dataset,
+    fn: Callable,
```
jkglasbrenner: Kind of vague, what's the input/output that's allowed for a mapping function? Is it something like `Callable[[tf.Tensor, tf.Tensor], tf.Tensor]`?

extra/plugins/dioptra_optic/models.py
```diff
@@ -0,0 +1,465 @@
...
+
+from .utils import DioptraMetricsLoggingCallback
+
+LOGGER = structlog.get_logger()
+
+
+@pyplugs.register
+def create_model(
+    model_name: str,
+    dataset: tf.data.Dataset,
```
jkglasbrenner: Design-wise it's surprising that initializing a model architecture requires having a tf.data.Dataset object available. It's a heavy dependency for a function like this when all you actually need here are the _specifications_ that set the dimensions for the input and output layers.

An idea is to modify the `data.load_dataset_from_tfds` and `load_dataset_from_image_directory` plugin tasks and have them return a dataclass that contains a metadata spec for the dataset, for classification that's the input_shape and the number of classes. Then you could pass this metadata spec object to this function instead of the full dataset. This would be also be nice for testing purposes.

extra/plugins/dioptra_optic/models.py
```diff
@@ -0,0 +1,465 @@
...
+    fit_options: dict[str, Any] | None = None,
+) -> keras.Model:
+    """Trains a keras model using model.fit.
+    Registers metrics to the Dioptra Job.
+
+    Args:
+        model: The keras model to be trained.
+        train_dataset:
+        val_dataset:
+        epochs:
```
jkglasbrenner: Missing docstrings

extra/plugins/dioptra_optic/attacks.py
```diff
@@ -0,0 +1,241 @@
...
+        eps_step: The step size of the input variation for minimal perturbation
+            computation. The default is `0.1`.
+        minimal: If `True`, compute the minimal perturbation, and use `eps_step` for the
+            step size and `eps` for the maximum perturbation. The default is `False`.
+        save_dataset: Whether to save the dataset to disk.
+
+    Returns:
+        The transformed tf.data.Dataset.
+    """
+    batch_size, num_classes = next(iter(dataset))[1].shape
```
jkglasbrenner: Prefer to use the dataset specs rather than consuming from the iterator:

```suggestion
    batch_size, num_classes = [int(x) for x in dataset.element_spec[1].shape.as_list()]
```

Or, even better, we could keep this information in the metadata spec object for the dataset that I suggested in my comment on the `models.create_model` plugin task.

extra/plugins/dioptra_optic/attacks.py
```diff
@@ -0,0 +1,241 @@
...
+    clip_values = (0.0, 255.0) if ops.max(next(iter(dataset))[0]) > 1.0 else (0.0, 1.0)
```
jkglasbrenner: Isn't this something that we can infer from knowing the `normalize_val` that's used when loading the dataset? Also, are these the only two clip_values that make sense? What if normalization produces an image matrix with values between -1 and 1? Or -127.5 and 127.5?

The min, max values for the image matrix elements post normalization are something that could be included in the metadata spec object for the dataset that I suggested in my comment on the `models.create_model` plugin task.

extra/plugins/dioptra_optic/attacks.py
```diff
@@ -0,0 +1,241 @@
...
+
+    y_target = (
+        None
+        if target is None
+        else ops.repeat([ops.one_hot(target, num_classes)], batch_size, axis=0)
+    )
+
+    @tf.numpy_function(Tout=(tf.float32, tf.float32))
+    def attack_fn(x, y):
+        return attack.generate(x, y=y_target), y
```
jkglasbrenner, line 90: Maybe I'm misunderstanding what's needed for generate, but why not this?

```python
    @tf.numpy_function(Tout=(tf.float32, tf.float32))
    def attack_fn(x, y):
        return attack.generate(x, y=y), y
```

Maybe you have to include the one hot transform on the `y=y` part in the attack_fn, but that should be easy enough to add.

Why does `y_target` have to be captured in a closure like this with a repeating op?

extra/plugins/dioptra_optic/attacks.py
```diff
@@ -0,0 +1,241 @@
...
```
jkglasbrenner: The same comments I have on the FGM plugin task also apply here.

extra/plugins/dioptra_optic/attacks.py
```diff
@@ -0,0 +1,241 @@
...
+        scale_min: The min scale of the patch as a fraction of image size.
+        scale_max: The max scale of the patch as a fraction of image size.
+        learning_rate: The learning rate used in optimization.
+        max_iter: The maximum iterations in the optimization.
+
+    Returns:
+        The trained patch and its mask.
+    """
+    batch_size, num_classes = next(iter(dataset))[1].shape
+    clip_values = (0.0, 255.0) if ops.max(next(iter(dataset))[0]) > 1.0 else (0.0, 1.0)
```
jkglasbrenner: The same comment I have on the FGM plugin task for these metadata specs also apply here.

extra/plugins/dioptra_optic/attacks.py
```diff
@@ -0,0 +1,241 @@
...
+        patch: The adversarial patch to be applied to the dataset.
+        rotation_max: The max angle to rotate the patch by.
+        scale: The scale of the patch as a fraction of image size.
+        save_dataset: Whether to save the dataset to disk.
+
+    Returns:
+        The transformed tf.data.Dataset
+    """
+    batch_size, num_classes = next(iter(dataset))[1].shape
+    clip_values = (0.0, 255.0) if ops.max(next(iter(dataset))[0]) > 1.0 else (0.0, 1.0)
```
jkglasbrenner: The same comments I have on the FGM plugin task also apply here.

extra/plugins/dioptra_optic/attacks.py
```diff
@@ -0,0 +1,241 @@
...
+        rotation_max=rotation_max,
+        scale_min=scale_min,
+        scale_max=scale_max,
+        learning_rate=learning_rate,
+        max_iter=max_iter,
+        verbose=False,
+    )
+
+    x, _ = next(iter(dataset))
+    y_target = ops.repeat([ops.one_hot(target, num_classes)], batch_size, axis=0)
```
jkglasbrenner: Isn't this just grabbing the first image return by the tf.data.Dataset object? Is that actually what we would want?

extra/plugins/dioptra_optic/defenses.py
```diff
@@ -0,0 +1,72 @@
...
+        dataset: The tf.data.Dataset to apply the defense to.
+        defense_name: The name of the defense.
+        defense_options: Options passed to the defense as kwargs.
+
+    Returns:
+        The defended dataset.
+    """
+    defense_options = defense_options or {}
+
+    clip_values = (0.0, 255.0) if ops.max(next(iter(dataset))[0]) > 1.0 else (0.0, 1.0)
```
jkglasbrenner: The same comment I have on the FGM plugin task (`attacks.fast_gradient_method`) also apply here.

extra/plugins/dioptra_optic/utils.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class DioptraMetricsLoggingCallback(Callback):
+    """
+    A Keras Callback for logging metrics to the current Dioptra job.
+    """
+
+    def __init__(self, logger):
+        self._logger = logger
+        self._dioptra_client = get_authenticated_worker_client(LOGGER, "json")
```
jkglasbrenner, line 69: The logging capture enabled on the Dioptra worker should be enough so that you don't have to embed the client like this anymore. Probably what we need to do is update the logs that are captured/filtered based on the logger's name, and then just say something like "if your logger's name begins with `dioptra_plugin_task.` then it will be captured".

src/dioptra/sdk/utilities/auth_client.py
```diff
@@ -0,0 +1,65 @@
...
+
+ENV_DIOPTRA_API: Final[str] = "DIOPTRA_API"
+ENV_DIOPTRA_WORKER_USERNAME: Final[str] = "DIOPTRA_WORKER_USERNAME"
+ENV_DIOPTRA_WORKER_PASSWORD: Final[str] = "DIOPTRA_WORKER_PASSWORD"
+
+
+def get_authenticated_worker_client(
+    log: logging.Logger | BoundLogger,
+    client_type: Literal["json", "response"] = "response",
+) -> DioptraClient:
```
jkglasbrenner: Looks like the return type annotation(s) were dropped.

```suggestion
) -> DioptraClient[DioptraResponseProtocol] | DioptraClient[dict[str, Any]]:
```

pyproject.toml
```diff
@@ -85,8 +85,10 @@ dependencies = [
     "simplejson>=3.10.0",
     "SQLAlchemy>=2",
     "structlog>=20.2.0",
+    "tensorflow-datasets>=4.9.9",
```
jkglasbrenner, line 89: Isn't this something that only gets used in the worker? (the download_data.py script can be executed with `uv run`, which is the other place I believe it's used). Putting this here will mean it gets packaged with the restapi container. I'd recommend moving it down to the worker optional dependency.

pyproject.toml
```diff
@@ -85,8 +85,10 @@ dependencies = [
...
     "tomli>=2.2.0",
     "typing-extensions>=3.7.4.3",
+    "universal-pathlib>=0.2.6",
```
jkglasbrenner, line 92: I agree with moving this from the "examples" optional dependency up here, since `UPath` gets used in an edit to the Artifact handler to make sure artifact URIs come out correctly.

Given that we've completely replaced the old `download_data.py` script (and also now will suggest using `uv run` to execute it), remove the entirety of the `examples` optional dependency, since that was only there for that script.

When you regenerate `uv.lock`, it's important to use `uv lock --upgrade` and not just run `uv lock` by itself. Something weird happens with the version constraints otherwise (it seems to be a consequence of us pulling from the pytorch wheel index) that can break cross-platform compatibility.

pyproject.toml
jkglasbrenner, line 1: I'm not able to figure out why, but the tensorflow-metal dependency led to crashing behavior on my Apple Silicon Mac when I tested out the `model_train.yml` entrypoint. Removing line 214 from pyproject.toml,

https://github.com/usnistgov/dioptra/blob/9904efc91d65b683ad1ff7316c8697ff7bd50602/pyproject.toml#L214

and regenerating the lockfile using `uv lock --upgrade` resolved the issue.

Since tensorflow-metal enables usage of the Mac's onboard GPU, this makes me worry that something about the plugin task code is unstable on GPUs in general. If at all possible, we should validate if the Tensorflow-based dioptra-optic plugin tasks work when used on an Nvidia GPU.

I _believe_ the old version of the Mnist classifier example didn't cause crashes like this. I can try to diagnose further, but if we can't figure it out, we're going to have to remove this dependency since the crashes seem to be unpredictable.

examples/hello_world.ipynb
jkglasbrenner, line 1: The changes to the hello_world notebook look like changes that were done for development purposes, should these be reversed?

jtsextonMITRE: I think these were intentional actually - the goal was to keep mnist_demo and hello_world in sync, and to follow the assumption that the user is using the container setup, so we changed most of the front matter of these to match each other and also use one of the container's default queues

jkglasbrenner: That part makes sense. The change I am noticing is `resolve_name_conflicts_strategy="update"` in the plugin import, but that isn't available in this branch (it's a different pull request that hasn't been merged yet I think).

The `suppress(StatusCodeError)` is different, but I can see the reasoning behind it.

Maybe in `DIOPTRA_REST_API_ADDRESS` we want to explicitly include the port 80 in the URL, `localhost:80`? Just to make it clear how to adjust it if someone needs to change the port docker forwards to?

jtsextonMITRE: > That part makes sense. The change I am noticing is resolve_name_conflicts_strategy="update" in the plugin import, but that isn't available in this branch (it's a different pull request that hasn't been merged yet I think).

Ah yes I agree overwrite makes more sense. 

> Maybe in DIOPTRA_REST_API_ADDRESS we want to explicitly include the port 80 in the URL, localhost:80? Just to make it clear how to adjust it if someone needs to change the port docker forwards to?

This makes sense to me.

pyproject.toml
keithmanville: I had the same issue with `tensorflow-metal`. I seem to remember diagnosing this as an issue with my setup, because I tried a super minimal example and it still crashed. But I can try to test on a gpu.

examples/hello_world.ipynb
keithmanville: Yeah, as Julian said this was to make setup in both notebooks consistent. I think I caught the update strategy in the mnist one but not here.

pyproject.toml
keithmanville: yes, worker only. That's a mistake on my part. I re-did the dependencies when I cleaned up and squashed commits and forgot to set the --extra flag for this.

extra/plugins/dioptra_optic/data.py
keithmanville: it's setting the buffer used for shuffle to 1/10th of the dataset. I can make this clear with a comment (and likely breaking this out into its own statement). I'll also think about a better heuristic for setting the buffer size.

examples/scripts/download_data.py
keithmanville: good catch!

extra/plugins/dioptra_optic/data.py
keithmanville: I think I was working off of this guidance where it says cache should come after batch for eval pipelines.
https://www.tensorflow.org/datasets/keras_example#build_an_evaluation_pipeline

I have it cache then batch for the train subset.

extra/plugins/dioptra_optic/models.py
keithmanville: I can fix this. I think at one point I wasn't sure if the preprocess function always had the same name, so I was passing that in as a variable. But then I removed it and hardcoded "preprocess_input"

extra/plugins/dioptra_optic/data.py
keithmanville: I can fix this. I think it should be two tensors for both input and output.

extra/plugins/dioptra_optic/models.py
keithmanville: I think at some iteration of this I had a similar approach to what you're suggesting. But I had it as a second return type so that it could be captured independently by the task engine and artifact handler.

Would you be okay with a return type that's something like (tf.data.Dataset, DatasetMetaDataClass)?

extra/plugins/dioptra_optic/attacks.py
keithmanville: wasn't aware of element_spec. But if we want to avoid passing datasets just for there metadata anyway, I can pass along a metadata object instead.

extra/plugins/dioptra_optic/attacks.py
keithmanville: Yes, I think this goes along with your feedback of having a Dataset Metadata object we pass around. I was trying to do everything with passing only tf.data.Dataset.

extra/plugins/dioptra_optic/attacks.py
keithmanville: Not sure I'm totally following this comment.

y_target is only set if a target label is provided.

I kept y_target outside of attack_fn so that operation doesn't happen every time. The repeat is to match the batch size of x.

extra/plugins/dioptra_optic/attacks.py
keithmanville: dataset should be batched, so it would be a batch of images, but yes.

Here x contains the background images the patch is applied to while training the patch. We might want to provide a larger subset of the data here.

extra/plugins/dioptra_optic/utils.py
keithmanville: But I need the client to post metrics to the job. See line 85.

extra/plugins/dioptra_optic/data.py
jkglasbrenner: Ah okay, that makes sense then. I think the advice I found when I googled about it was assuming you had random shuffling in the pipeline, so the batches would be different each epoch. This matches what you have in the two sections.

Maybe just add a comment, including a link to the URL you provided so it's clear?

extra/plugins/dioptra_optic/data.py
jkglasbrenner: Got it. The concern I have is this could become extremely large on big datasets (maybe not an issue with anything offered via Tensorflow datasets?). Conversely, on very small datasets (few hundred examples) the shuffling would be poor.

I suggest we can just go with a default integer like 10,000 as a default so it doesn't scale with the size of the dataset, and then allow users to change it via the entrypoint's global parameters, just in case they are working on a resource constrained machine. We can also add logic to check the number of examples of the data, and if it's less than the buffer size (10,000 for default), then it switches to the number of examples instead (e.g. dataset has 8000 examples, 8000 < 10000 so buffer = 8000).

extra/plugins/dioptra_optic/models.py
jkglasbrenner: Yeah, that's what I would suggest. Based on the return signature, it'd be something like:

```python
tuple[tf.data.Dataset | None, tf.data.Dataset | None, tf.data.Dataset | None, DatasetMetadataClass]
```

And I agree, this should be a separate plugin parameter type in the task engine.

extra/plugins/dioptra_optic/attacks.py
jkglasbrenner: My mistake, I missed that `target` was an argument passed to the plugin task itself. I think I had gotten too into the weeds with looking at other code and lost the context, and for some reason thought the attack target should match the true label for each image (which now that I think about it again, would mean its not a useful attack).

Yes, if this is a user-supplied input, then the closure you've created here makes sense.

extra/plugins/dioptra_optic/attacks.py
jkglasbrenner: Okay got it. Yeah, it feels like we'd want the number of images needed for training the patch to be independent of the batch size, otherwise if I reduce the batch size to something like 8, I'll probably end up with a poor patch and not really understand why.

Maybe we should have users specify a percentage of the total data to use for the patch? The default could be 0.1% (0.001), which for the MNIST training set would be 60 randomly selected images.

Not sure how much the label distribution of the data you use to make the patch matters for optimal patch generation. Ideally I'd think you'd want a proportional sample of all labels when making the patch, although I don't know how easy/hard it would be to do a stratified shuffle split with a `tf.data.Dataset` object.

extra/plugins/dioptra_optic/utils.py
jkglasbrenner: Ah, you're right, my mistake, ignore my comment. I was assuming it was just for the logging endpoint and forgot about the metrics endpoint and got confused.

I do think there's a way we can create a metrics logger that leverages the Python logging machinery and can be routed to the correct endpoint so you don't have to embed the authenticated client here. However, that would be a separate feature that's outside the scope of this work, so this works for now.

pyproject.toml
jkglasbrenner: I used this as a quick minimal example, and it runs when I have `tensorflow-metal==1.1.0` installed with Tensorflow 2.16:

```python
import numpy as np
import tensorflow as tf
from tensorflow import keras


# Load the Fashion MNIST dataset
fashion_mnist = keras.datasets.fashion_mnist
(train_images, train_labels), (test_images, test_labels) = fashion_mnist.load_data()

# Define the class names to make labels understandable
class_names = ['T-shirt/top', 'Trouser', 'Pullover', 'Dress', 'Coat',
               'Sandal', 'Shirt', 'Sneaker', 'Bag', 'Ankle boot']

# Normalize the pixel values to be between 0 and 1
train_images = train_images / 255.0
test_images = test_images / 255.0

# Define the model architecture
model = keras.Sequential([
    keras.layers.Input(shape=(28, 28)),
    keras.layers.Flatten(),
    keras.layers.Dense(128, activation='relu'),
    keras.layers.Dense(10, activation='softmax')
])

# Compile the model
model.compile(optimizer='adam',
              loss='sparse_categorical_crossentropy',
              metrics=['accuracy'])

# Train the model
model.fit(train_images, train_labels, epochs=2)

# Evaluate the model on the test dataset
test_loss, test_acc = model.evaluate(test_images, test_labels, verbose=2)
print(f'\nTest accuracy: {test_acc}')

# Make predictions on the test set
predictions = model.predict(test_images)

# Get the predicted class for the first image
predicted_label = np.argmax(predictions[0])
print(f'Prediction for the first test image: {class_names[predicted_label]}')

# Get the true label for comparison
true_label = test_labels[0]
print(f'True label for the first test image: {class_names[true_label]}')
```

Not sure what is different about the other example that tensorflow-metal doesn't like.

extra/plugins/dioptra_optic/attacks.py
keithmanville: You can actually call generate multiple times and it will continue optimizing from the current state of the patch. So I think I'm going to implement this as a training loop over the dataset using the existing max_iter parameter.

PR #1001, opened 2025-10-29
Plugin dropdown component

### LINKED ISSUE #989: [BUG] Attaching Plugins to Entrypoints in UI
**Describe the bug**
There are two functionalities in the UI that don't seem ideal w.r.t. attaching plugins to Entrypoints. 

1. You can save an entrypoint without attaching any plugins. This should probably trigger an error or at least a warning, as a task graph won't execute without any attached plugins
2. More importantly, you are not able to attach new plugins to an entrypoint once it's been submitted in the UI. So if you create an entrypoint and attach a plugin, but want to later go back and add a new plugin (say, an artifact task), you aren't able to do this in the UI.

**Expected behavior**
1. If you don't attach a plugin to an entrypoint, a warning/error banner should appear in the UI
2. After opening a previously submitted entrypoint, you should be able to open it up and attach new plugins (right now those selection boxes are greyed out)

 - OS: WSL2
 - Browser Chrome

**Server**
- OS: Ubuntu
- Virtualization: Docker

**Additional context**
Referenced in [this google doc](https://docs.google.com/document/d/1x9QQlG8WTRPWbTb0Z57mE195cM7I4yyqa2w3uPew1Bc/edit?tab=t.ukwxae5948w0#heading=h.52n20dfovfvq)

henrychoy:
closes #989 

- Adding/removing/syncing plugins to entrypoints should be possible in both the entrypoint table page and the entrypoint edit page.
- The plugin dropdown is now a re-usable component.  Adding/removing/syncing should work like before.

### DIFF
--- a/src/frontend/src/components/AssignPluginsDropdown.vue
+++ b/src/frontend/src/components/AssignPluginsDropdown.vue
@@ -1,0 +1,116 @@
+<template>
+  <q-select
+    outlined
+    dense
+    v-model="selectedPlugins"
+    use-input
+    use-chips
+    multiple
+    option-label="name"
+    option-value="id"
+    input-debounce="100"
+    :options="pluginOptions"
+    @filter="getPlugins"
+    @add="(added) => addPlugin(added.value)"
+    @remove="(removed) => removePlugin(removed.value)"
+  >
+    <template v-slot:before>
+      <div class="field-label">Plugins:</div>
+    </template>
+    <template v-slot:selected>
+      <div>
+        <div
+          v-for="(plugin, i) in selectedPlugins"
+          :key="plugin.id"
+          :class="i > 0 ? 'q-mt-xs' : ''"
+          >
+            <q-chip
+              removable
+              color="secondary"
+              text-color="white"
+              class="q-ml-xs "
+              @remove="selectedPlugins.splice(i, 1); removePlugin(plugin)"
+            >
+              {{ plugin.name }}
+              <q-badge
+                v-if="!plugin.latestSnapshot" 
+                color="red" 
+                label="outdated" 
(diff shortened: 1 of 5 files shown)


henrychoy:
Regarding our discussion on the call, I confirmed that when creating a new entrypoint, it uses a single POST call.  And when editing an entrypoint, it uses the PUT call to edit, and the any additional POST/DELETE calls to update the plugins as needed.

Should be good to review again, let me know if any issues.

henrychoy:
> The UI elements look good and work as expected.
> 
> Can you add logic to hit the minimum number of endpoints on submit based on what's changed? e.g. only PUT to `/entrypoints/{id}` if something other than plugin/artifact plugins is modified.
> 
> This will prevent unnecessary requests to the REST API and prevent additional snapshots from being created when only plugins are attached/detached.

@keithmanville Thanks for catching that.  Just updated so that it will not call PUT `/entrypoints/{id}` if only plugins/artifactPlugins are changing.  And the plugins/artifactPlugins calls are currently only made when necessary.  Should be good now.

henrychoy:
> The UI looks good to me as well!
> 
> This is tangentially related, and I've asked it in other places, but should we be able to mix artifact plugins and function plugins? When you try to, you get this error: "Bad Request - The plugin task(s) {7} are in the list of artifact tasks, when they should be separate."
> 
> There are a couple other bugs I raised connected to this that we should coordinate on.

@colton-lapp I had the same questions so it'll be good to discuss on the next call with the team.  The plugins have no type and looks like you can select any plugin as a regular plugin or an artifact plugin in an entrypoint (you just can't select the same plugin as both types in an entrypoint).

PR #1008, opened 2025-11-06
feat(restapi): add swaps base function and tests

### LINKED ISSUE #822: Create a function for rendering a swappable task graph YAML
This function should take a task graph yaml and a list of chosen swaps and return a rendered version.

The output should be a stripped down version of the graph that works with the task engine.

The chosen swaps should be specified via the following structure:
```
[
  { "swap_name": str, "swap_task": str},
  ...
]
```

Each possible swap should have one and only one swap task specified.
Each swap task name must exist in the provided yaml.

Blocked by:
- #820

### Defintion of Done:
- [ ] New tests are added to test the render function
- [ ] All tests pass
- [ ] The feature is merged into dev.

jtsextonMITRE:
Closes #822 based on #832


src/dioptra/sdk/utilities/swaps.py
```diff
@@ -0,0 +1,86 @@
...
+
+    """
+    rendered_graph: dict[str, Any] = {}
+
+    used_swaps = set()
+    not_found_swaps = set()
+    not_found_tasks = set()
+
+    for step, task in graph.items():
+        if step.startswith("_"):
```
keithmanville: remove this. I think this was part of a design where we would have content in the steps position that could be referenced (via anchor / aliases) elsewhere.

src/dioptra/sdk/utilities/swaps.py
```diff
@@ -0,0 +1,86 @@
...
+    Renders a task graph given a graph containing swaps and dictionary
+    specifying the swap choices.
+
+    Args:
+        graph: A dictionary object representing a task graph.
+        swaps: A dictionary mapping swap names to the selected task.
+
+    Returns:
+        The rendered graph using the selected swaps.
+
```
keithmanville: remove extra line

keithmanville (commented):
very minor feedback based on code review. I still need to test things out myself and go through the tests.

How do we want to handle swaps issues? Merge into dev, or do we want to hold until after 1.1? We could have a larger swaps feature branch? The swaps issues depend on eachother, so we need somewhere to merge so we don't get blocked.

keithmanville (commented):
recommend renaming files to `entrypoint_swaps` and `test_entrypoint_swaps` to be more explicit.

keithmanville (commented):
I think we decided to not promote the anchor/alias idea. Recommend removing the `_with_anchor` yaml and replacing any tests.

I think we said we want this to work on a graph with no swaps, right? Can you add that as a test?

PR #1009, opened 2025-11-12
feat(UI): make small UI Tweaks

colton-lapp:
This PR makes two small UI tweaks.

**Fix display for artifact tasks to remove unused "Input Parameters":**
- [x] Remove the "Input Parameters" from the Artifact Plugin Task registration table (you cannot set input parameters for Artifact Plugin Tasks in the UI) while keeping "Input parameters" for Function Plugin Tasks

**Enhance UI by adding loading indicators**:
- [x] Implement loading animation with minimum 300ms loading for all tables i.e. entrypoints, jobs, artifacts, etc (implemented minimum loading time to remove "flicker" effect. We could implement [advanced logic](https://stackoverflow.com/questions/71518887/is-it-ok-to-set-a-minimum-display-time-for-the-spinner-from-the-ux-standpoint) to not show the loading animation for the first 300ms or so instead. 

Potential future changes we could work on after discussing with the team:
- Enable multiple selection and deletion in all tables. Checkboxes on each row would allow you to select multiple, and then a button would let you delete all of them. 
- "Save parameter and add another" feature - when creating an entrypoint parameter, add a button that says "Save parameter and add another" which would save the current parameter but keep the dialogue box open for additional parameters. Small quality of life improvement to increase speed
- Job log display tweaks - adding color for severity, syntax highlighting potentially, filters, etc to log view for job


src/frontend/src/components/TableComponent.vue
```diff
@@ -102,15 +100,15 @@
               v-if="col.name === 'expand'" 
               size="md" flat dense round  
               @click.stop="props.expand = !props.expand" 
-              :icon="props.expand ? 'expand_less' : 'expand_more'" 
+Thinker:                :icon="props.expand ? 'expand_less' : 'expand_more'"
```
henrychoy: Not sure what "Thinker" is here?

src/frontend/src/views/PluginsView.vue
```diff
@@ -1,17 +1,29 @@
...
+      class="q-mb-md"
+    />
+
   <TableComponent
     :rows="plugins"
     :columns="columns"
     title="Plugins"
     v-model:selected="selected"
-    @edit="router.push(`/plugins/${selected[0].id}`)"
+    selection="multiple"     @edit="router.push(`/plugins/${selected[0].id}`)"
```
henrychoy: put on separate lines

src/frontend/src/components/TableComponent.vue
```diff
@@ -18,25 +18,32 @@
     @keydown="keydown"
     :rows-per-page-options="props.showAll ? [0] : [5,10,15,20,25,50,0]"
     :hideBottom="props.hideBottom && rows.length > 0"
+    :loading="props.loading"
   >
     <template v-slot:header="props">
       <q-tr :props="props">
+                <q-th auto-width v-if="selection === 'multiple'">
```
henrychoy: Just a nitpick, but some of the indents on the edits in this file are off

colton-lapp:
> Hi Colton, I think the changes look great. The 300ms minimum loading time for the table looks good.
> 
> The multiselect/delete on the plugins page looks good as well, but it clashes with the "click to open" workflow that the rest of the pages follow. think if the team really wants the ability to multiselect, maybe we can add a small "delete multiple" checkbox somewhere in the header of the table. But I think the rest of the team will have some thoughts as to how we want to approach multi-select in general, will probably be a good topic for next call.

I agree the team should probably weigh in on the multi-select / delete logic. I could clean up the PR to address your edits and delete the text artifact for loading though. I wonder if there is some way to put the "wait 300 ms" logic into a decorator or standalone wrapper function so that it's not repeated everywhere... maybe eventually we could enhance the logic to not show loading if the request takes less than 100ms. I'm still very new to Vue / Quasar, so I don't really know what the best pattern is here.

colton-lapp:
Sounds good! Well I deleted/reverted the code that implemented the "delete multiple" in the PluginsView, which should address all your comments. If it looks good to you, we can merge the loading functionality and the artifact task registration changes into dev

henrychoy (approved):
Looks good to me, thanks!

PR #1019, opened 2025-12-01
Added Dynamic Entrypoint Globals endpoint

### LINKED ISSUE #818: Dynamic entrypoint global parameters
To support entrypoint task swaps, we need to support a dynamic list of global params that is dependent on which tasks are specified for each swap.

This can be implemented before the swaps feature is complete and should work the same way on entrypoints without any swaps.

This will allow us to notify the user of any unused global parameters.

The implementation will parse a entrypoint graph YAML and cross reference declared global params for the entrypoint with the dereferenced variable names in the graph YAML (these are the ones that are not linked to step names).

This assumes the entrypoint is validated.

This will add a new endpoint which information about the validated entrypoint. This information should include:
- The topologically sorted steps/tasks (tells you the active tasks)
- The list of global parameters (those that are necessary to run the list of tasks)
inputs: 
- entrypoint snapshot ID
- list of chosen swaps
outputs: 
```
{
    "entrypoint_params": list of entrypoint params,
    "topological_sort": [task_name, ...],
    "active_plugins": [ PluginSnapshotRefSchema, ...],
}
```

### Definition of Done:
- [ ] The new endpoint is added
- [ ] New tests are added
- [ ] All tests pass

jtsextonMITRE:
Closes #818


tests/unit/restapi/v1/swaps/test_entrypoint_swaps.py
```diff
@@ -0,0 +1,276 @@
...
+    {
+        "swaps": {"step2_choice": "task10", "step3_choice": "task2"},
+        "globals": ["global1", "global6", "global12"],
+        "sort_order": [["step1", "step3", "step4", "step2"]],
+        "active_plugins": ["plugin1", "plugin9", "plugin13"],
+    },
+]
+
+
+@pytest.fixture
```
keithmanville: fixture should be moved to conftest

tests/unit/restapi/v1/swaps/test_entrypoint_swaps.py
```diff
@@ -0,0 +1,276 @@
...
+            filename=plugin_filename,
+            contents=plugin_file_contents,
+            function_tasks=tasks,
+        ).get_json()
+
+        output[plugin_name] = plugin_response
+    return output
+
+
+@pytest.fixture
```
keithmanville: fixture should be moved to conftest

tests/unit/restapi/v1/swaps/test_entrypoint_swaps.py
```diff
@@ -0,0 +1,276 @@
...
+            parameters=parameters,
+            plugin_ids=plugin_ids,
+            queue_ids=queue_ids,
+        ).get_json()
+
+        output[entrypoint["name"]] = entrypoint_response
+    return output
+
+
+def test_entrypoint_swaps_endpoint(
```
keithmanville: break this up into multiple tests. It is testing various cases.

Also rename the new tests to indicate they are testing the dynamic global params endpoint. swaps endpoint is too general.

keithmanville (changes requested):
I made a couple minor changes for you and am recommending others.

In addition to inline comments, can you move the tests into the entrypoint test file since this an entrypoint endpoint now?

PR #1025, opened 2025-12-08
Navbar and button cleanup

henrychoy:
Closes NavBar cleanup and remove floating create button from #971 

- Move NavBar icon buttons into dropdowns
- Remove floating create button
- Move all routing logic to `index.ts` file where it belongs


src/frontend/src/components/NavBar.vue
```diff
@@ -53,74 +53,129 @@
...
+              <q-item-label>GitHub Repository</q-item-label>
+            </q-item-section>
+            <q-item-section side>
+              <q-icon name="fa-brands fa-github" size="sm" />
+            </q-item-section>
+          </q-item>
+          <q-separator />
+          <q-item>
+            <q-item-section>
+              <q-item-label>Dioptra Version v1.0</q-item-label>
```
keithmanville: version should be read from somewhere (env var?) not hardcoded.

src/frontend/src/components/NavBar.vue
```diff
@@ -53,74 +53,129 @@
...
+        </q-list>
+      </q-menu>
+    </q-btn>
+
+    <q-btn round icon="sym_o_info" flat class="q-mr-xs">
+      <q-menu>
+        <q-list class="noselect">
+          <q-item v-ripple href="https://pages.nist.gov/dioptra/" target="_blank" clickable v-close-popup>
+            <q-item-section>
+              <q-item-label>What is Dioptra?</q-item-label>
```
keithmanville: I would call this "Dioptra Documentation." I don't know that there will be a "What is Dioptra?" page in the new docs.

src/frontend/src/components/NavBar.vue
henrychoy: Sure I can get it from an env file, but it'd have to be hardcoded there too.  It'd be nice if I can get it from the backend.  I see that the backend has the info below under /swagger.json => info.  Is this a reliable place to get the version number?  Not sure what dev0 means but I can strip it out when displaying it?

```
{
    "title": "Dioptra REST API",
    "version": "1.0.0dev0"
}
```

Issue #757, opened 2025-02-07
As a user I would like to be able to easily swap Plug-ins within an Entry Point in order to compare different algorithms, attacks, defenses, etc. encapsulated by each plugin


hbooth:
Need to design a new feature enabling that would enable substitution for Task plug-ins within an Entry Point.  Conceptually this could be thought task plugin slots or Entry Point Template. Initial thoughts include that task plug-ins will need to produce the same outputs in order to be substitutable with each other.

Other items to consider would be what are the types of task plugins (i.e. level of abstractions) and what are the types of outputs.

The design should produce artifacts such as:
 - Conceptual user operations
 - Mockups
 - Notional REST API calls and results

PR #1063, opened 2025-12-12
Entrypoint swaps feature

keithmanville:
#757


src/frontend/src/views/CreateJob.vue
jkglasbrenner, line 1: **issue (blocking):** Prevent out-of-order responses from changing the current form

This is a race condition. Changing the entry point or a swap choice sends asynchronous requests for the rendered graph and its required parameters:

```text
GET /api/v1/entrypoints/{id}/snapshots/{snapshotId}/config
GET /api/v1/entrypoints/{id}/snapshots/{snapshotId}/dynamicGlobalParameters
```

The responses are not guaranteed to arrive in the order they were sent, but the current implementation applies them without checking whether the entry point, snapshot, and swap choices that started the request are still selected.

We reproduced the problem by adding controlled latency to one response:

1. Select an entry point and swap choices that require `alpha_value` and `delta_value`. Hold the response from `GET /api/v1/entrypoints/{id}/snapshots/{snapshotId}/dynamicGlobalParameters`.
2. Switch to a no-swap entry point. Allow its `GET /api/v1/entrypoints/{id}/snapshots/{snapshotId}/config` and `GET /api/v1/entrypoints/{id}/snapshots/{snapshotId}/dynamicGlobalParameters` requests to complete. The current graph and Values form now require only `alpha_value`.
3. Release the older `dynamicGlobalParameters` response. Because it arrives last, it adds the previous entry point's `delta_value` field to the current form.

c3-after-release

The controlled delay guarantees that the responses arrive out of order during the reproduction. The same ordering can occur in a real deployment whenever an earlier request experiences more network or server latency than a later one. Asynchronous requests do not guarantee their arrival order.

A failed request can produce a related stale state. The graph updates to the new selection, but Values retains rows from the previous selection without displaying an error or recovery message.

Please ensure that out-of-order responses cannot change the form after their entry point, snapshot, or swap choices are no longer current. A failed current request should also clear or mark the affected Values instead of leaving rows from the previous selection in place. One possible approach is to assign each form refresh a request identifier and apply its responses only while that identifier still belongs to the latest refresh.

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -1158,6 +1370,789 @@ def get(self, name: str, group_id: int, **kwargs) -> models.EntryPoint | None:
...
+    rendered_graph: dict[str, Any],
+) -> tuple[set[str], set[str]]:
+    """Return external graph references and task names used by a graph."""
+    graph_steps = rendered_graph.keys()
+    needed_globals: set[str] = set()
+    used_tasks: set[str] = set()
+
+    for step in rendered_graph.values():
+        for task_name, task_definition in step.items():
+            used_tasks.add(task_name)
```
jkglasbrenner, line 2147: **issue (non-blocking):** Return active plugins for mixed task invocations

For example, consider an entry point associated with one plugin named `scoring`. That plugin defines the task `score`, which is invoked using the supported mixed form:

```yaml
evaluate:
  task: score
  args: [$predictions]
  kwargs:
    threshold: 0.5
```

Calling the following endpoint should identify `predictions` as a required global parameter and `scoring` as the plugin used by the graph:

```text
GET /api/v1/entrypoints/{id}/snapshots/{snapshotId}/dynamicGlobalParameters
```

The relevant part of the response should look like this:

```json
{
  "entrypointParams": [{"name": "predictions"}],
  "activePlugins": [{"name": "scoring"}]
}
```

Instead, the endpoint recognizes the global parameter but reports no active plugin:

```json
{
  "entrypointParams": [{"name": "predictions"}],
  "activePlugins": []
}
```

This happens because `_get_required_globals()` adds the mixed invocation's keys `task`, `args`, and `kwargs` to `used_tasks` instead of reading `score` from the `task` field. The response therefore recognizes the task's inputs while failing to identify the plugin that provides the task.

Please use the existing `util.step_get_plugin_short_name()` helper so short, positional, and mixed invocations all resolve task names consistently and return the corresponding active-plugin metadata.

jkglasbrenner (changes requested):
I am trying something different with how I am organizing and structuring my code review this time around. I am structuring my comments according to the [Conventional Comments](https://conventionalcomments.org/) standard, which follows a structure that is similar to [Conventional Commits](https://www.conventionalcommits.org):

````text
<label> (<blocking|non-blocking>): <subject>

<discussion>
````

I am only using a subset of the proposed labels (I've also tweaked the descriptions to match how I'm trying to use them):

| label | description |
| --- | --- |
| `issue` | Issues highlight specific problems with the subject under review. It is recommended to pair this comment with a `suggestion`. Use a `question` instead if it is unclear whether a problem exists. |
| `question` | Questions are appropriate if you have a potential concern but are not quite sure if it’s relevant or not. |
| `suggestion` | Suggestions propose improvements or implementation approaches to the current subject. They should not carry a blocking/non-blocking decoration. If by themselves and not part of an `issue`, they should be optional and non-blocking. |
| `nitpick` | Nitpicks are trivial preference-based requests. These should be non-blocking by nature. |

I'm also only using `blocking` and `non-blocking` decorations:

| decoration | description |
| --- | --- |
| `blocking` | 	A comment with this decoration **should** prevent the subject under review from being merged until it is resolved. |
| `non-blocking` | A comment with this decoration **should not** prevent the subject under review from being merged. |

Regarding the review itself, generally speaking, I found that the standard swap workflow appears to work when it is exercised with the patterns our team typically uses in Dioptra. The pattern I see across the findings below is that the validation strategy and tests mostly exercise swaps in isolation, with consistent plugin metadata and our usual graph syntax. Less common but supported task-engine forms, interactions between swaps, and equivalent operations through different API and UI paths are not covered consistently. That helps explain why the missing stable output interface and several inconsistencies in validation, rendering, and UI state did not surface earlier. The individual findings below show the concrete failures.

The requested changes I placed in this summary are cross-cutting concerns that touch multiple files or are about broader behavior that can't be isolated to specific lines in one file. I left inline comments where one file or location contains the complete problem and is where the solution should also be implemented.

### issue (blocking): Give each swap a stable output interface

While working through this review, I came across a fundamental design issue. Swaps are effectively our version of the Gang of Four's [Strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern), but they do not have a stable output namespace. Because every choice is supposed to be interchangeable, validating only the output types is not enough. I think this issue is best illustrated with an example. Let's assume that we have registered the following tasks:

```yaml
emit_x:
  plugin: org.example.emit_x
  outputs:
    - x: number
emit_y:
  plugin: org.example.emit_y
  outputs:
    - y: number
constant:
  plugin: org.example.constant
consume:
  plugin: org.example.consume
  inputs:
    - value: number
```

The relevant tasks are `emit_x` and `emit_y`. They both return a single numeric value, but `emit_x` names its output `x` while `emit_y` uses `y`. For simplicity, assume that these outputs are semantically equivalent. Now consider the following entry point graph:

```yaml
producer:
  ?producer-choice:
    x-output:
      emit_x: []
    y-output:
      emit_y: []
consumer:
  ?consumer-choice:
    independent:
      constant: []
    use-x:
      consume:
        value: $producer.x
```

The validation procedure in this PR checks the following combinations: `x-output + independent`, `y-output + independent`, and `x-output + use-x`. It does not check `y-output + use-x`. Running the full validator over that omitted combination rejects it with `In step "consumer": reference "producer.x": unrecognized output: x`. The worker performs the same task-engine validation, so a job that selects this combination will fail before task execution.

This problem is different from the [dependency cycle issue](#issue-blocking-prevent-any-swap-selection-from-forming-a-dependency-cycle) discussed later. That is an implementation issue where an invalid graph can sneak through because the current validation does not cover interactions between nondefault choices. Here, even if validation surfaced the problem before persistence, the core problem would still be one of design. The graph has no single output name that is valid for every choice, even though swap choices are meant to be fully interchangeable.

One solution would be to extend the constraints on tasks used in a swap by requiring their output names, as well as their types, to be identical. I do not like this solution. Plugins and tasks are intended to be reusable building blocks that can be built and registered independently, but this constraint would indirectly couple their naming decisions to entry points. It would also make it much harder to build a swap from tasks provided by multiple third-party plugins. In effect, it would impose a shared output-name vocabulary across otherwise independent plugins, which works against an ecosystem of shareable plugins.

**suggestion:** My recommended solution is to update the design by requiring users to declare the output interface for each swap explicitly.

We could reserve a `?outputs` list inside the swap declaration containing the canonical output names in order. The example below illustrates what I'm proposing:

```yaml
tasks:
  score_model:
    plugin: example.org.score_model
    inputs:
      - metric_name: string
    outputs:
      - score: number
      - class_name: string
  predict_model:
    plugin: example.org.predict_model
    outputs:
      - probability: number
      - label_name: string
  summarize:
    plugin: example.org.summarize
    inputs:
      - prediction: number
      - label: string

graph:
  inference:
    ?strategy:
      ?outputs:
        - prediction
        - label
      scoring:
        score_model: "accuracy"
      prediction:
        predict_model: []
  report:
    summarize:
      prediction: $inference.prediction
      label: $inference.label
```

Here, every reference to `inference` uses the declared interface rather than the underlying task's output names. Once the interface is declared, direct references to the underlying names should be invalid even when the selected task happens to expose them. The length of `?outputs` establishes the required output count, and validation can infer the type at each position from the registered task metadata and require every option to have the same type name at that position. The interface standardizes only the outputs. Each choice can still have a different input signature, so the full validation that saves and dry runs should share still needs to render each choice once and check its inputs against the rest of the graph.

<details>
<summary>Example generated task-engine configuration for a selected swap</summary>

Let's assume the user selects the `scoring` choice for the `?strategy` swap. Under this design, rendering the graph choice alone is not enough. Dioptra also needs to update the generated `tasks` section when it builds the task-engine configuration. The selected task definition needs a normalized local name, its actual plugin coordinate, the output names declared by the swap interface, and the types resolved from its registered metadata. An example of the generated task-engine configuration follows:

```yaml
tasks:
  generated_inference_strategy:
    plugin: example.org.score_model
    inputs:
      - metric_name: string
    outputs:
      - prediction: number
      - label: string
  summarize:
    plugin: example.org.summarize
    inputs:
      - prediction: number
      - label: string

graph:
  inference:
    generated_inference_strategy: "accuracy"
  report:
    summarize:
      prediction: $inference.prediction
      label: $inference.label
```

Here, the input definitions from `score_model` are copied verbatim into `generated_inference_strategy`, while its registered output names `(score, class_name)` are remapped by position to `(prediction, label)`. Selecting the `prediction` choice instead would remove the generated task's `inputs:` field, retain the same canonical output names, and use `predict_model`'s actual plugin coordinate. The `inference` invocation under `graph` would also change to `generated_inference_strategy: []` because that choice takes no inputs. The generated task-definition key is an implementation detail and need not follow the pattern `generated_<swap-step-name>_<swap-name>`.

</details>

I think this is the model we should implement for swaps in this PR. It gives each swap a stable interface without forcing independently registered tasks to adopt entry-point-specific names.

### issue (blocking): Prevent any swap selection from forming a dependency cycle

It is possible for an entry point to pass the full validation that saves and dry runs should share even when certain swap combinations create a dependency cycle, which violates the task graph's directed acyclic graph requirement. The current full validation approach works by varying one swap at a time while every other swap remains fixed on its first option. Our thought experiment on whether we needed to evaluate all combinations did not account for two nondefault choices referencing each other's outputs.

For example, this graph is accepted by `POST /api/v1/entrypoints`:

```yaml
left:
  ?left-choice:
    left-default:
      task: constant_a
    left-cross:
      passthrough:
        value: $right.value
right:
  ?right-choice:
    right-default:
      task: constant_b
    right-cross:
      passthrough:
        value: $left.value
```

The current validation approach does not catch the cycle because each cross-reference is valid while the other swap uses its default option. But if you then submit a job that selects both `left-cross` and `right-cross` through `POST /api/v1/experiments/{experimentId}/jobs`, job submission still succeeds. You receive an HTTP 200 status code, the job and both swap rows are created in the database, and the job is placed on the queue. Once the worker picks up the job and reruns validation, it rejects the rendered graph with:

```text
Step cycle detected: left > right > left
```

This means we need a different approach to validating the correctness of the swap setup that guarantees that every possible swap selection is acyclic.

**suggestion:** Create and validate a graph that represents the union of dependencies across all swap choices

The naive way to do this is to validate the Cartesian product of all combinations, but that is computationally expensive. Thankfully for us, we already have the machinery in place to validate this in a single check. What we need to do is create a graph that represents the union of dependencies across all swap choices and require it to contain no cycles. Let's call this the `may-depend-on` graph. We can express this synthetic graph using the existing `dependencies:` field. For the above example, the `may-depend-on` graph looks like this:

```yaml
left:
  dependencies: [right]
right:
  dependencies: [left]
```

Our existing cycle validation can now see the possible cycle and reject it.

<details>
<summary>Implementation sketch using the existing task-engine helpers</summary>

The dependency union can be constructed directly from the pre-render swap graph. It does not require rendering every choice or writing another YAML parser. [`get_references()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/task_engine/util.py#L199-L229) can inspect each alias definition, [`get_reference_coords()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/task_engine/util.py#L191-L196) identifies the referenced step, and [`get_sorted_steps()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/task_engine/util.py#L253-L290) performs the topological sort and cycle check.

```python
from dioptra.task_engine import util

step_names = set(swap_graph)
may_depend_on = {
    step_name: set()
    for step_name in swap_graph
}

for step_name, step_definition in swap_graph.items():
    explicit_dependencies = step_definition.get("dependencies", [])
    if isinstance(explicit_dependencies, str):
        explicit_dependencies = [explicit_dependencies]

    may_depend_on[step_name].update(explicit_dependencies)

    for key, aliases in step_definition.items():
        if not key.startswith("?"):
            continue

        for alias_definition in aliases.values():
            for reference in util.get_references(alias_definition):
                referenced_step, _ = util.get_reference_coords(reference)
                if referenced_step in step_names:
                    may_depend_on[step_name].add(referenced_step)

union_graph = {
    step_name: {"dependencies": sorted(dependencies)}
    for step_name, dependencies in may_depend_on.items()
}

util.get_sorted_steps(union_graph)
```

For the example above, the generated graph is:

```yaml
left:
  dependencies: [right]
right:
  dependencies: [left]
```

The existing topological sort rejects it with:

```text
StepReferenceCycleError: Step cycle detected: left > right > left
```

This is an implementation sketch rather than a required integration point, but it demonstrates that the existing task-engine utilities can perform the analysis before rendering.

</details>

We need to update this shared full validation to construct the `may-depend-on` graph and ensure that a cyclic selection cannot be persisted as an available entry point or reach job submission.

### issue (blocking): Keep step dependencies when rendering a swap

[`render_swaps_graph()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/sdk/utilities/entrypoint_swaps.py#L48-L69) can discard a step's `dependencies` when that metadata appears before the swap key, allowing a dependent task to run before its prerequisite.

This standalone example defines `consume_step` as dependent on `prepare_step`:

```python
import yaml

from dioptra.sdk.utilities.entrypoint_swaps import render_swaps_graph

graph = yaml.safe_load(
    """
consume_step:
  dependencies: [prepare_step]
  ?implementation:
    selected:
      consume:
prepare_step:
  prepare:
"""
)

print(render_swaps_graph(graph, {"implementation": "selected"}))
```

It prints:

```text
{'consume_step': {'consume': None}, 'prepare_step': {'prepare': None}}
```

The rendered `consume_step` no longer contains `dependencies: [prepare_step]`. In a worker run with this graph, Dioptra ran the consuming task first and it failed because the prerequisite had not created its output.

Please fix this to preserve the containing step's metadata independently of YAML key order so rendering cannot change the declared execution order.

### issue (blocking): Require only parameters used by the selected swap options

Job submission currently requires global parameters from options that were not selected, even though the selected graph does not use them and the Create Job UI gives the user no way to provide them.

For example, create an entry point with two swap groups where each option uses a different required parameter:

```yaml
first:
  ?first-choice:
    kw-alpha:
      alpha:
        value: $alpha_value
    positional-beta:
      beta: [$beta_value]
second:
  ?second-choice:
    mixed-gamma:
      task: gamma
      args: [$gamma_value]
      kwargs:
        suffix: mixed
    kw-delta:
      delta:
        value: $delta_value
        suffix: keyword
```

Declare `alpha_value`, `beta_value`, `gamma_value`, and `delta_value` as required parameters, then select `kw-alpha` and `mixed-gamma`. `GET /api/v1/entrypoints/{id}/snapshots/{snapshotId}/dynamicGlobalParameters?swaps=first-choice:kw-alpha,second-choice:mixed-gamma` returns only `alpha_value` and `gamma_value`, and the Create Job page likewise provides fields for only those two values. However, submitting the job through `POST /api/v1/experiments/{experimentId}/jobs` is rejected with:

```text
The following job Parameters are missing a value: beta_value,delta_value.
```

The Create Job UI cannot supply those extra values. A client can submit this selection only by adding `beta_value` and `delta_value` as a workaround, despite neither parameter being used by the selected options. When swaps narrow the required global parameters, job submission should require only the minimal set used by the selected options, not the full list including parameters associated with options that were not selected.

Please fix this so that a user is able to submit a job without having to provide values for unused parameters without a default value.

### issue (blocking): Restore the existing task-name syntax

This PR narrows the existing task-engine naming rules in a way that should be reversed. Before this change, the keys used to name tasks in `tasks` and steps in `graph` could contain `:` and `,`. In the current PR, `experiment_schema.json` limits the task name used by an ordinary graph-step invocation to letters, numbers, and underscores. Graph step names remain unrestricted, and `swap_graph.json` still allows the broader character set for swap names and aliases.

For example, this configuration passed schema validation in the `dev` branch but fails in this PR because `score:model` no longer matches the task-invocation pattern:

```yaml
tasks:
  "score:model":
    plugin: scoring.score
graph:
  "run,scoring":
    "score:model": []
```

The `tasks` block can still declare that task name, and the graph can still use `run,scoring` as a step name, but the graph is no longer allowed to invoke the declared task. The equivalent restriction is not applied to task names inside swap choices. This leaves the task-engine syntax internally inconsistent and rejects configurations that were valid before this PR.

Please remove the new character restriction so that ordinary and swappable task invocations both preserve the existing task-engine naming rules.

### issue (blocking): Use an unambiguous representation for swap choices

Swap names and aliases permit `:` and `,`, but the current query format for `GET /api/v1/entrypoints/{id}/snapshots/{snapshotId}/config` also uses `:` to separate a swap name from its selected alias and `,` to separate choices. This leads to problems when the schema/controller layer attempts to parse the request parameters into the structure needed for the service layer. This issue is analogous to the parsing problem behind SQL injection, in that a character the user intended as data can instead terminate or alter the surrounding structure.

For example, this entry point saves successfully:

```yaml
selected:
  ?method:
    attack:
      task: constant_a
    "attack:v2":
      task: constant_b
    "attack,v3":
      task: constant_b
```

`GET /api/v1/entrypoints/{id}/snapshots/{snapshotId}/config` with `swaps=method:attack:v2` returned HTTP 200 but rendered `constant_a`, as if `attack` had been selected. Submitting `attack:v2` in the JSON body to `POST /api/v1/experiments/{experimentId}/jobs` persisted the complete alias, rendered and executed `constant_b`, and finished successfully. Asking the query endpoint for `attack,v3` returned HTTP 400 because the comma was interpreted as the separator between choices.

The JSON representation therefore preserves the selected name while the query representation changes or rejects it. Preview, parameter discovery, and job submission need to follow the same naming rules and resolve the same selection.

**suggestion:** Replace the delimiter-based query format used in `src/dioptra/restapi/v1/entrypoints/schema.py` with the OpenAPI 3 [`deepObject`](https://spec.openapis.org/oas/v3.1.1.html#style-values) representation for objects with scalar properties

An example of using this representation for selecting two different swaps is as follows:

```text
?swaps[method]=attack%3Av2&swaps[secondary]=attack%2Cv3
```

The UI and Python client should construct this representation from their existing swap-choice dictionaries and let their URL libraries handle encoding.

Note that sending a request that provides more than one value for the same swap key should be rejected because each swap selects exactly one alias. For example, a request like `?swaps[method]=attack%3Av2&swaps[method]=attack%2Cv3` should be rejected when it hits `GET /api/v1/entrypoints/{id}/snapshots/{snapshotId}/config`.

<details>
<summary>Potential implementation approach</summary>

Flask will parse the example query into an `ImmutableMultiDict` resembling:

```python
ImmutableMultiDict(
    [
        ("swaps[method]", "attack:v2"),
        ("swaps[secondary]", "attack,v3"),
    ]
)
```

The Dioptra request layer should convert that into the structure already expected by the entry-point services:

```python
{
    "swaps": {
        "method": "attack:v2",
        "secondary": "attack,v3",
    }
}
```

In practice, the existing controller decorators can remain unchanged. `SwapChoiceRequestSchema` already owns the `swaps` query field, and `EntrypointConfigRequestSchema` inherits from it. A Marshmallow `pre_load` hook on `SwapChoiceRequestSchema` can therefore normalize the query for both endpoints before the rest of the schema validation runs:

```python
@login_required
@accepts(query_params_schema=EntrypointConfigRequestSchema, api=api)
@responds(schema=EntrypointConfigResponseSchema, api=api)
def get(self, id: int, snapshotId: int):
    ...
```

The implementation could resemble:

```python
import re

from flask import request
from marshmallow import Schema, ValidationError, fields, pre_load


SWAP_QUERY_KEY = re.compile(r"^swaps\[([^\[\]]+)\]$")


def parse_deep_object_swaps(args):
    swaps = {}

    for key in args.keys():
        if key != "swaps" and not key.startswith("swaps["):
            continue

        match = SWAP_QUERY_KEY.fullmatch(key)
        values = args.getlist(key)
        if match is None or len(values) != 1:
            raise ValidationError(
                {"swaps": [f"Invalid or repeated swap selection: {key}"]}
            )

        swaps[match.group(1)] = values[0]

    return swaps


class SwapChoiceRequestSchema(Schema):
    swaps = fields.Dict(
        keys=fields.String(),
        values=fields.String(),
        load_default=dict,
    )

    @pre_load
    def load_deep_object_swaps(self, data, **kwargs):
        normalized = {
            key: value
            for key, value in data.items()
            if key != "swaps" and not key.startswith("swaps[")
        }
        normalized["swaps"] = parse_deep_object_swaps(request.args)
        return normalized
```

When a request arrives, `@accepts` calls the schema before invoking the controller method. The helper reads the original `request.args` `MultiDict`, collects keys matching `swaps[<swap-name>]`, and uses `getlist()` to reject repeated values. The hook removes the bracketed transport keys from the flattened input, adds the normalized `swaps` dictionary, and returns the complete query data to the ordinary schema-loading process. If the hook raises `ValidationError`, `@accepts` returns HTTP 400 and the controller body does not run. Otherwise, `@accepts` assigns the validated result to `request.parsed_query_params` and invokes the controller normally.

For example, this request should return HTTP 400:

```text
?swaps[method]=attack&swaps[method]=attack%3Av2
```

The original `MultiDict` preserves both values:

```python
request.args.getlist("swaps[method]")
# ["attack", "attack:v2"]
```

The duplicate check must use the original `request.args`. The regular dictionary passed into `pre_load` (the `data` argument) has already lost repeated scalar values because `flask-accepts` constructed it by iterating over `request.args` with `MultiDict.items()`.

Update the Python client and frontend to emit the `deepObject` representation from their existing dictionaries. Apply the same query adapter to configuration preview and dynamic-parameter discovery.

As a side note, Flask-RESTX currently generates OpenAPI 2.0 documentation, so it cannot describe `deepObject` directly. In the meantime, the parameter should use a documented example so something meaningful appears in the Swagger docs. Implementing this change should not require patching Flask or forking or modifying the underlying source code of `flask-accepts`.

</details>

### issue (blocking): Make validate-only requests behave like real dry runs

This PR deliberately replaces `POST /api/v1/workflows/validateEntrypoint` with `validateOnly=true` on the ordinary entry-point create and update endpoints. I initially pushed for keeping this path lightweight because I wanted validation to be inexpensive enough to run repeatedly against an entry-point editor. After seeing how `validateOnly` is integrated into the broader API and reviewing the relevant guidance, I have changed my mind. Since I contributed to the ambiguity here, I want to be explicit about the direction I now think we should take: `validateOnly` should mean a true dry run, and we should use that definition consistently going forward.

Google's [`validate_only` guidance](https://google.aip.dev/163) says that a validation-only request should perform the same permission and validation checks as the live request, fail whenever the live request would fail, and return the same response status and body. It may omit output fields that cannot exist without executing the operation, such as an automatically generated ID. The current implementation runs only the lighter checks, so it does not yet meet that contract.

Please make `validateOnly=true` a true dry run of entry-point creation or modification. It should run the same checks as the corresponding save, return the same error when that save would fail, and stop before writing a resource or snapshot to the database. On success, it should return the same entry-point representation while omitting only fields that cannot be known without persistence.

**suggestion:** The lighter checks still have a useful role. We have always intended to use them for low-latency feedback from an entry-point editor, where rendering every swap choice after each change would be unnecessarily expensive. That operation should have its own contract rather than changing the meaning of a dry run. I suggest exposing it as `POST /api/v1/entrypoints:lint` for a proposed entry point and `POST /api/v1/entrypoints/{id}:lint` for an existing one. Google's guidance for [custom methods in resource-oriented APIs](https://google.aip.dev/136) specifically uses a colon before the custom verb for operations that do not fit the standard resource methods.

<details>
<summary>Suggested editor lint interface</summary>

The lightweight lint operation would accept the proposed entry-point content and return a lint report rather than an entry-point resource:

```http
POST /api/v1/entrypoints:lint
POST /api/v1/entrypoints/{id}:lint
```

```json
{
  "valid": false,
  "issues": [
    {
      "path": "graph.inference",
      "message": "..."
    }
  ]
}
```

A completed lint request can return HTTP 200 even when `valid` is false. A malformed request, an authorization failure, or an inability to perform the check should still return the corresponding error status.

The colon is a literal part of the URL, and the current Flask and Flask-RESTX stack accepts and documents it through the usual route declaration. The collection and item routes can be registered without changing the underlying packages:

```python
@api.route(":lint")
class EntrypointLintEndpoint(Resource):
    ...


@api.route("/<int:id>:lint")
class EntrypointIdLintEndpoint(Resource):
    ...
```

</details>

### issue (blocking): Validate imported entry points before committing the batch

This PR (finally) makes it so that we run full validation whenever an entry point is saved through the ordinary create or update endpoints (I thought we had activated that earlier, but it seems we didn't), but resource import still bypasses those checks. For example, I imported an entry point with no declared parameters and this graph:

```yaml
graph:
  selected:
    alpha:
      value: $missing_global
```

A direct `POST /api/v1/entrypoints/` request containing the same undeclared reference returned HTTP 400 and identified both the missing global parameter and the unresolvable reference. `POST /api/v1/workflows/resourceImport` returned HTTP 200 with `successfully imported` and created the plugin and entry point.

The import service calls the entry-point service with `commit=False` so that it can keep the transaction open and commit the complete batch once. The same entry-point service also uses that flag to decide whether to run the full checks required before a save. These are separate concerns, meaning that callers need to be able to request full validation without taking ownership of the commit.

Please update this so that the same full checks applied to an ordinary saved entry point are also applied to the imports before committing.

### issue (blocking): Validate task-plugin changes before creating an entry-point snapshot

Certain sequences of API calls make it possible to store a graph that points to plugin tasks that are no longer associated to the entry point, which should no longer be possible since we have switched to validating entry points when we save them. I found two separate ways to do this, which here I'll refer to as the "removal case" and the "synchronization case." Let's assume that we register two plugins, Plugin A and Plugin B. Plugin A provides the task `task_from_plugin_a` and Plugin B provides the task `task_from_plugin_b`. We set up an entry point that gets associated with both plugins and contains this graph:

```yaml
choose:
  ?implementation:
    use_plugin_a:
      task_from_plugin_a:
        input_param: $input_param
    use_plugin_b:
      task_from_plugin_b:
        input_param: $input_param
```

In the removal case, I called `DELETE /api/v1/entrypoints/{id}/plugins/{pluginId}` for Plugin A. This removes only Plugin A's association from the new entry-point snapshot. It does not delete Plugin A from Dioptra. The request returned HTTP 200 and created a snapshot that retained the graph above but associated only Plugin B. To test whether this new current state could pass the ordinary save path, I sent a `PUT /api/v1/entrypoints/{id}?validateOnly=false` request that resubmitted the entry point's existing graph, parameters, and queues while preserving its current plugin associations. The request attempted to create another snapshot from exactly that state, but full validation returned HTTP 400 because the graph still referred to `task_from_plugin_a` even though Plugin A was no longer associated. No additional snapshot was created. Swap discovery also returned HTTP 400 for the same reason.

In the synchronization case, I started from a separate entry point that was still associated with the original Plugin A and Plugin B snapshots. I updated Plugin A's plugin file through `PUT /api/v1/plugins/{pluginId}/files/{fileId}` with an empty `tasks.functions` list. That created a new Plugin A snapshot whose registered task metadata no longer contained `task_from_plugin_a`. It did not change the entry point yet. I then posted Plugin A's resource ID to `POST /api/v1/entrypoints/{id}/plugins`. For an already associated plugin, this operation preserves the other associations and moves that plugin's association to its latest snapshot. It returned HTTP 200 and created a new entry-point snapshot associated with the taskless Plugin A snapshot and the existing Plugin B snapshot while again retaining the graph above. I then sent that entry point's current data through the same ordinary `PUT` path without changing any inputs. The request attempted to create another snapshot, but full validation returned HTTP 400 because the graph still referred to `task_from_plugin_a` even though the associated Plugin A snapshot no longer supplied that task. No additional snapshot was created, and swap discovery also returned HTTP 400.

In both cases, the preceding historical entry-point snapshot remained valid because it retained its original plugin-snapshot associations. The inconsistency was introduced only in the new current snapshot.

[`EntrypointIdPluginsService.append()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/restapi/v1/entrypoints/service.py#L768-L837) and [`EntrypointIdPluginsIdService.delete()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/restapi/v1/entrypoints/service.py#L892-L962) predate this PR, so I do not think the association inconsistency itself is newly introduced. It matters for this change because the PR enables full validation for a saved entry point, and these are additional paths that save a new entry-point snapshot after changing metadata used to resolve its graph.

Please assemble the prospective plugin-snapshot associations and run the same full validation used by an ordinary entry-point save before committing either removal or synchronization of a plugin association. If validation fails, do not create a new entry-point snapshot or change its current plugin associations.

### issue (blocking): Reject generated configurations with unresolved task references

The configuration endpoint builds the `tasks` section from the plugin snapshots associated with an entry-point snapshot, then renders that snapshot's stored graph using the requested swap choices. These two parts of the generated configuration need to agree: every task invoked by the rendered graph needs a corresponding definition in `tasks`.

Using the same setup as the [issue above](#issue-blocking-validate-task-plugin-changes-before-creating-an-entry-point-snapshot), it is possible to get the API to return an invalid task-engine configuration whose rendered `graph` block invokes tasks without corresponding definitions in the `tasks` block. For completeness, let's repeat the setup process here. Assume that we register two plugins, Plugin A and Plugin B. Plugin A provides the task `task_from_plugin_a`, Plugin B provides the task `task_from_plugin_b`, and both plugins are associated with an entry point containing this graph:

```yaml
choose:
  ?implementation:
    use_plugin_a:
      task_from_plugin_a:
        input_param: $input_param
    use_plugin_b:
      task_from_plugin_b:
        input_param: $input_param
```

After saving the entry point, I called `DELETE /api/v1/entrypoints/{id}/plugins/{pluginId}` to remove Plugin A's association. The request returned HTTP 200 and created a new entry-point snapshot that associated only Plugin B while retaining the graph above. I then requested the complete configuration for that new snapshot from `GET /api/v1/entrypoints/{id}/snapshots/{snapshotId}/config` with `swaps=implementation:use_plugin_a`.

The endpoint returned HTTP 200. The relevant parts of its response were equivalent to:

```yaml
tasks:
  task_from_plugin_b:
    # The remaining task definition from Plugin B.
graph:
  choose:
    task_from_plugin_a:
      input_param: $input_param
```

The `tasks` section reflects the current Plugin B association, but the rendered graph still invokes `task_from_plugin_a` from the removed Plugin A association. If you instead submit the same current entry-point data through the ordinary `PUT /api/v1/entrypoints/{id}?validateOnly=false` update path, full validation returns HTTP 400 because that task cannot be resolved. Swap discovery also returns HTTP 400.

This happens because [`EntrypointConfigService.get_config()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/restapi/v1/entrypoints/service.py#L1408-L1444) builds the task definitions from the associated plugin snapshots, renders the graph separately, and returns the combined result without checking that the rendered task invocations can be resolved from those definitions.

Please make complete rendered-configuration requests reject this inconsistent state instead of returning HTTP 200 by validating the rendered configuration before returning it. Fixing the [issue above](#issue-blocking-validate-task-plugin-changes-before-creating-an-entry-point-snapshot) will prevent association write paths from creating new inconsistent snapshots, but configuration generation should still fail when it encounters an existing snapshot whose rendered graph cannot be resolved from its associated plugin snapshots.

### issue (blocking): Finish removing the retired validation interface

Removing `POST /api/v1/workflows/validateEntrypoint` and `WorkflowsCollectionClient.validate_entrypoint` was deliberate, but the old interface has not been removed completely. The remaining references are:

- The [Python-client reference](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/docs/source/reference/python-client/workflows-client-methods.rst#L65) still directs users to the missing method.
- [`VALIDATE_ENTRYPOINT`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/client/workflows.py#L27) remains as an unused workflow URL constant.
- [`validateEntrypoint()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/frontend/src/services/dataApi.ts#L519-L521) remains as an unused frontend helper for the deleted route.
- The old request and response schemas remain in [`workflows/schema.py`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/restapi/v1/workflows/schema.py#L237-L351), along with another unused validation-issue schema in [`entrypoints/schema.py`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/restapi/v1/entrypoints/schema.py#L518-L525).

Please remove these remnants and document `validateOnly` as the supported dry-run replacement. If the lightweight checks remain available, document the separate lint interface as well.

### issue (blocking): Apply the same validation rule to pending artifact-plugin changes

For an existing entry point, the update request validates against the plugin associations that have already been saved. The [Validate button and its tooltip](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/frontend/src/views/CreateEntryPoint.vue#L563-L572) account for this by checking `taskPluginsChanged` and telling the user to save task-plugin selections first. The component separately tracks [`artifactPluginChangesPending`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/frontend/src/views/CreateEntryPoint.vue#L1412-L1417), but the Validate button does not apply the same rule to it. This allows the page to report that the entry point is valid while checking different artifact plugins from those currently shown in the editor.

Please disable validation whenever either kind of plugin change is pending so the two association types follow the same rule.

### issue (blocking): Make swap-rendering errors and their tests deterministic

[`render_swaps_graph()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/sdk/utilities/entrypoint_swaps.py#L71-L81) handles three expected invalid-input conditions by raising plain `Exception` objects and interpolating sets directly into their messages. [The unused-swaps test](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/tests/unit/sdk/utilities/entrypoint_swaps/test_entrypoint_swaps.py#L100-L114) constructs another set independently and interpolates it into `pytest.raises(match=...)`. Python does not guarantee that sets will render their contents in a deterministic order when you convert them to a string, so this test can fail even when the renderer reports the correct names. Using `Exception` also allows an unrelated failure inside the renderer to satisfy the test or be repackaged by the REST service as an ordinary rendering error.

**suggestion:** Sort the names used in these three messages and raise `ValueError` for the renderer's expected invalid-input cases.

For example:

```python
unused_swaps = sorted(swaps.keys() - used_swaps)

if unused_swaps:
    raise ValueError(f"Swaps {unused_swaps} were provided but not used.")
```

Then, in [the configuration service](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/restapi/v1/entrypoints/service.py#L1436-L1442) and [the dynamic-global-parameters service](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/restapi/v1/entrypoints/service.py#L1477-L1480), catch `ValueError` instead of `Exception` like so:

```python
try:
    rendered = render_swaps_graph(graph, swaps)
except ValueError as error:
    raise EntrypointSwapsRenderError(str(error)) from error
```

Then, update [the unit test](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/tests/unit/sdk/utilities/entrypoint_swaps/test_entrypoint_swaps.py#L100-L114) to check the specific type and canonical message with a small local function:

```python
expected_swap_names = ["extra", "load", "transform_data"]

def is_expected_unused_swaps(error: ValueError) -> bool:
    return str(error) == (
        f"Swaps {expected_swap_names} were provided but not used."
    )

with pytest.raises(ValueError, check=is_expected_unused_swaps):
    render_swaps_graph(...)
```

Note that because [`pytest.raises(check=...)`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.raises) requires pytest 8.4 or later, we'll want to update [the development dependency](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/pyproject.toml#L147) to `pytest>=8.4` and set [pytest's configured minimum version](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/pyproject.toml#L386-L387) to `8.4`. After you make that change, regenerate [`uv.lock`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/uv.lock#L5355-L5356) with [the repository's documented dependency-upgrade command](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/DEVELOPER.md#L132-L137):

```bash
uv lock --upgrade
```

As a general heads up, if we need to rebase on the `dev` branch again before this PR gets merged, and that brings in changes to `uv.lock` and/or `pyproject.toml`, we will have to run `uv lock --upgrade` one last time to ensure the lockfile stays consistent.

### issue (blocking): Let unexpected full-validation failures propagate

[`validate_single_swap_combinations()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/restapi/v1/entrypoints/service.py#L1703-L1734) catches the general `Exception` class, which means it handles every exception raised while rendering a selected graph, constructing its task-engine configuration, running full validation, rewriting issues, and extracting global references. This means that any sort of crash or failure that happens while code runs in that block gets reported as invalid user input. This can obscure the actual root cause of a problem in the logs.

This can be illustrated by replacing `render_swaps_graph()` with a stub that raises `RuntimeError("unexpected validation failure")` and then calling `validate_single_swap_combinations()` directly. Instead of allowing that unexpected exception to propagate, the method returns the following semantic validation issue:

```text
[Swap combination {}] Error rendering graph: unexpected validation failure
```

During a full entry-point save, issues returned by this method are handled as invalid entry-point input and result in HTTP 400, when they should actually result in HTTP 500 and a different error message.

The fix depends on the `ValueError` change requested in the [issue above](#issue-blocking-make-swap-rendering-errors-and-their-tests-deterministic) being in place. Once `render_swaps_graph()` uses that type for raising swap selection failures, please narrow the `try` block so that it wraps only [the call to `render_swaps_graph()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/restapi/v1/entrypoints/service.py#L1704) and catch only `ValueError` there. In that `except ValueError` block, preserve the existing behavior of appending a `ValidationIssue` to `issues_for_swap`, then return `issues_for_swap` and the empty `required_globals` set before attempting the remaining operations.

Configuration construction, task-engine validation, issue processing, and global-reference extraction should remain outside the `try` block. `validate_task_engine_dict()` already reports ordinary user-input problems by returning `ValidationIssue` objects, so those should continue to be collected as they are now. Exceptions raised by these remaining operations indicate unexpected implementation failures and should propagate through the normal HTTP 500 path rather than being caught here.

### suggestion: Make the remaining swap errors deterministic and preserve unexpected failures

There are four related places introduced by this PR where the same deterministic-message and narrow-exception approach suggested in the ["Make swap-rendering errors and their tests deterministic" issue](#issue-blocking-make-swap-rendering-errors-and-their-tests-deterministic) should be applied:

- [`JobService._build_job_swaps()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/restapi/v1/jobs/service.py#L389-L399) calculates the names of required but omitted swaps as a set and includes that set directly in `UnspecifiedSwapsError`. Sort the swap names before formatting the message so the same invalid request always produces the same error message.
- [`SwapsValidationService.validate_swap_outputs()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/restapi/v1/entrypoints/service.py#L1631-L1672) collects each swap's output signatures in a set and includes that set directly in a `ValidationIssue`. Convert the signatures to a stable sorted representation before formatting the message.
- [`SwapsRetrievalService.get_swaps()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/restapi/v1/entrypoints/service.py#L2047-L2104) catches every exception from `yaml.safe_load()` and reports it as `InvalidYamlError`. It also passes an unsorted set of unresolved task names to `TasksNotFoundError`. Catch only `yaml.YAMLError` around the YAML load, and sort the unresolved task names before raising `TasksNotFoundError`.
- [`DelimitedKeyValuePairs._deserialize()` and `DelimitedValues._deserialize()`](https://github.com/usnistgov/dioptra/blob/593dccafc1591a58d159005e996215a315989dd3/src/dioptra/restapi/v1/entrypoints/schema.py#L369-L413) catch every exception raised while parsing the `swaps` and `sections` query parameters and convert it to `ValidationError`. Explicitly verify that both values are strings and that every `swaps` pair contains the expected key/value separator. Report those malformed inputs as `ValidationError` and allow unrelated implementation failures to propagate.

These cases do not need to be additional blockers by themselves. Making these changes would keep their user-visible errors stable and prevent unexpected failures from being misreported as bad input. If we implement these suggestions, I recommend adding unit tests to verify the canonical message order and that unrelated exceptions are not converted to validation errors.

keithmanville:
#1373 
- issue (blocking): Require only parameters used by the selected swap options

#1374
- issue (non-blocking): Return active plugins for mixed task invocations

#1378 
- issue (blocking): Validate imported entry points before committing the batch - https://github.com/usnistgov/dioptra/commit/8852ea6cac6b17106bada325e89307e616f2b9f2
- issue (blocking): Finish removing the retired validation interface - https://github.com/usnistgov/dioptra/commit/b0c4cc6632f6b7c55da28a4592e6e6f9a167e8c6
- issue (blocking): Keep step dependencies when rendering a swap - https://github.com/usnistgov/dioptra/commit/6d0b859ac2404d2bbd035d024a5da5c38a474d55
- issue (blocking): Restore the existing task-name syntax - https://github.com/usnistgov/dioptra/commit/52f9563db2d9b74f58a2d23156a8ea371e53168d
- issue (blocking): Use an unambiguous representation for swap choices - https://github.com/usnistgov/dioptra/commit/796cf1b7da33d1ec17b2e02ab06c4da593e26d7d
- issue (blocking): Apply the same validation rule to pending artifact-plugin changes - https://github.com/usnistgov/dioptra/commit/2e74a65e4f82886118423aa6cba7e71fd4d83f96
- issue (blocking): Make validate-only requests behave like real dry runs - https://github.com/usnistgov/dioptra/commit/a5974b16b25e3b226ee40b66bb4089a0333a225d
- issue (blocking): Validate task-plugin changes before creating an entry-point snapshot  - 55981df8
- issue (blocking): Prevent any swap selection from forming a dependency cycle - 7360bf86
- issue (blocking): Reject generated configurations with unresolved task references - 3f40ad19
- issue (blocking): Let unexpected full-validation failures propagate - 759f94b
- issue (blocking): Give each swap a stable output interface - 396a295f

#1379 
- issue (blocking): Make swap-rendering errors and their tests deterministic

#1380
- issue (blocking): Prevent out-of-order responses from changing the current form

PR #1068, opened 2025-12-16
feat: open new tab

henrychoy:
Closes remaining checkboxes in #971 

- Add ability to open table row in new tab
  - Middle click (mouse wheel)
  - Cmd + click (macOS)
  - Ctrl + click (Windows / Linux)
  - Right click -> open new tab
  - "open new tab" icon button next to id

- When opening in new tab or going directly to a URL, then hitting `Cancel`, it should go to the logical parent page.  Otherwise Cancel should act as a back button.

### DIFF
--- a/src/frontend/src/components/JobArtifactsTable.vue
+++ b/src/frontend/src/components/JobArtifactsTable.vue
@@ -4,7 +4,10 @@
     :columns="columns"
     :title="`Artifacts Created by Job ${route.params.id}`"
     v-model:selected="selected"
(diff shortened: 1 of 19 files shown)


keithmanville (changes requested):
If I open the entrypoint snapshot link from a job detail page in a new tab then hit cancel, nothing happens. Expected behavior is to go to the entrypoint table page.

Can you rebase on dev? I noticed some other recent changes like removing the dialog edits for some resources. There are conflicts that need to be resolved.

PR #1077, opened 2025-12-30
docs: queues and workers

### LINKED ISSUE #1057: Queues & Workers

jtsextonMITRE:
Closes #1057

### DIFF
--- a/docs/source/explanation/components/queues-workers-explanation.rst
+++ b/docs/source/explanation/components/queues-workers-explanation.rst
@@ -20,6 +20,15 @@
 Queues and Workers
 ================
 
+Summary: What is a Queue?
+-------------------------
+
+A **queue** is a resource in Dioptra which represents a logical job queue for workers watching that queue to take from. Entrypoints can
+be assigned a number of queues they are compatible with.
+
+When a job is submitted, a queue is selected for that job to be added to. Any worker listening to that queue can claim the job from the
+queue, and execute the job in its environment.
+
 Summary: What is a Worker?
 --------------------------
 
@@ -41,15 +50,6 @@ Additionally, custom workers can be created which watch queues of other names an
 allowing for additional requirements and packages to be included.
 
 See :ref:`how-to-using-custom-workers` for more information.
-
-Summary: What is a Queue?
--------------------------
-
-A **queue** is a resource in Dioptra which represents a logical job queue for workers watching that queue to take from. Entrypoints can
-be assigned a number of queues they are compatible with.
-
-When a job is submitted, a queue is selected for that job to be added to. Any worker listening to that queue can claim the job from the
-queue, and execute the job in its environment.
 
 
 .. rst-class:: fancy-header header-seealso


docs/source/explanation/components/queues-workers-explanation.rst
```diff
@@ -20,5 +20,56 @@
...
+
+.. code-block::
+    
+    MLFLOW_S3_ENDPOINT_URL=
+    MLFLOW_TRACKING_URI=
+    DIOPTRA_WORKER_USERNAME=
+    DIOPTRA_WORKER_PASSWORD=
+    DIOPTRA_API=
+
+The username and password mentioned here must provide access to a valid account for the worker to interact with the Dioptra RESTAPI.
```
colton-lapp: This seems like reference material to me. If there is no other place to put it (because we don't want a workers reference page, for example) I would put this at the end of this page

docs/source/explanation/components/queues-workers-explanation.rst
```diff
@@ -20,5 +20,56 @@
 Queues and Workers
 ================
 
+Summary: What is a Worker?
+--------------------------
+
+A **worker** is an environment in which a job can be executed. A worker should contain all of the requirements needed for a given
+entrypoint, such as any local files, python packages, or executables needed as part of the job.
```
colton-lapp: For the average user, I would imagine they _won't_ be creating custom workers. Thus, I would think the most important concepts to cover in an explanation document are:

- Workers are the execution environments for jobs
- Workers have a 1:1 relationship with queues. A queue dispatches jobs to a worker. They are currently linked by name (i.e. for a queue to send jobs to a worker, they have to be called the same thing)
- There are default workers you install during setup
- At a high level, a worker must contain certain things / be able to do certain things. 
- Workers are different from each other in specific ways (and thus, you have different workers and attach specific entrypoints to specific workers... maybe your code needs a GPU, so you attach it to a GPU based worker or something)
- There are some default workers for GPU and some for CPU. When would you use each? 
- The mechanics of how queues work - i.e. when a job is submitted, it goes into the queue, the queue pings the worker, things communicate over the RestAPI, etc.

You cover many of these ideas in this document, but not all of them. I also think this explanation leads with more technical details than might be necessary, and a lot of specific details could be put in a reference page or the "create custom workers" guides.

docs/source/explanation/components/queues-workers-explanation.rst
```diff
@@ -20,5 +20,45 @@
...
+    * pytorch_gpu - for systems which do have a GPU present, which contains PyTorch and ART as dependencies
+
+Creating queues with these names, as long as the corresponding worker is enabled, will allow jobs to be run in these worker containers.
+
+Additionally, custom workers can be created which watch queues of other names and provide different environments for the jobs to run in,
+allowing for additional requirements and packages to be included.
+
+See :ref:`how-to-using-custom-workers` for more information.
+
+Summary: What is a Queue?
```
keithmanville: I would move what is a queue above what is a worker.
This would match the explainer title. I also think that Queue flows in to Worker based on how it's written.

PR #1078, opened 2026-01-02
docs: task graph explainer

### LINKED ISSUE #1052: Task Graph

jtsextonMITRE:
Closes #1052

### DIFF
--- a/docs/source/reference/dioptra-components/task-graph-reference.rst
+++ b/docs/source/reference/dioptra-components/task-graph-reference.rst
@@ -34,6 +34,8 @@ In each of these invocation styles, you'll notice some commonalities:
 
     * The ``graph:`` keyword designates this section of the full entrypoint YAML dictionary as the task graph. If creating entrypoints through the UI, this keyword is unnecessary and is provided to the RESTAPI automatically.
 
+Invocation Styles
+-----------------
 
 Positional Style Invocation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~


docs/source/explanation/components/task-graph-explanation.rst
```diff
@@ -18,7 +18,109 @@
...
+      run_metrics:
+         predictions: $predictions
+
+.. figure:: /images/DAG.png
+   :alt: Generated directed acyclic graph based on dependencies within the task graph.
+   :figclass: border-image clickable-image 
+
+   The DAG generated from the above task graph.
+
+See Also
```
keithmanville, line 136: Use Colton's styling for this section.

docs/source/reference/dioptra-components/task-graph-reference.rst
```diff
@@ -0,0 +1,201 @@
+.. This Software (Dioptra) is being made available as a public service by the
```
keithmanville, line 1: add to index page

docs/source/explanation/components/task-graph-explanation.rst
```diff
@@ -18,7 +18,109 @@
...
+each of those tasks. A directed acyclic graph (DAG) representing the dependencies between the 
+various steps is constructed to ensure that the steps are executed in the correct order.
 
-[Todo] Insert an explanation about the task graph.
\ No newline at end of file
+Summary: What is a step?
+------------------------
+
+A **step** is one component of the task graph, which correlates, essentially, to a plugin task invocation
+with a given set of parameters. A single step has several components:
```
keithmanville, line 44: I think we should include an example step (or maybe a couple steps) in a code block up here. This would help with the explanation below where we are referring to variables below.

Maybe we just move the dag example up here and then everything else can refer to it.

docs/source/explanation/components/task-graph-explanation.rst
```diff
@@ -18,7 +18,118 @@
...
+
+
+Summary: What is a step?
+------------------------
+
+A **step** is one component of the task graph, which correlates, essentially, to a plugin task invocation
+with a given set of parameters. A single step has several components:
+    * a **step name**, which functions as a variable storage for the output of the invoked task,
+    * a **task name**, which references the plugin task to invoke by name. At runtime, plugins associated with the entrypoint will be searched for a matching task.
+    * a **list of parameters**, which are passed to the referenced plugin task during invocation.
```
colton-lapp: Consider reducing the indent, unless you were intending to have this be in a shaded call out box (when lists are indented in this way, a shaded box is added in Sphinx)

docs/source/explanation/components/task-graph-explanation.rst
```diff
@@ -18,7 +18,118 @@
...
+
+.. figure:: /images/task-graph-function.png
+   :alt: Task graph purpose.
+   :figclass: border-image clickable-image
```
colton-lapp: To me, this figure makes more sense to include before this section about the Step (after the 'Summary: what is a task graph?')

docs/source/explanation/components/task-graph-explanation.rst
```diff
@@ -18,7 +18,118 @@
...
+   :figclass: border-image clickable-image 
+
+   The primary function of the task graph is to describe the relationship between variables and plugins within an entrypoint.
+   It describes dependencies between the various steps, as well as how the outputs, global variables, and artifacts fit in as 
+   input to the plugins invoked in each step.
+
+
+Below is an example task graph. In the first step, the step name is ``rng``, the task name is ``configure_rng`` and the list of parameters contains
+a single member named ``seed`` which is being passed a value of ``1234``.
+.. code:: yaml
```
colton-lapp, line 55: This code editor isn't being formatted correctly. You might need a newline before the `.. code:: yaml`

docs/source/explanation/components/task-graph-explanation.rst
```diff
@@ -18,7 +18,118 @@
...
+         predictions: $predictions
+
+
+Variables
+~~~~~~~~~
+
+You can reference the output of steps by referencing the step name. This is useful for passing the
+output of one step as a parameter to another. For example: ``$trained_model`` will take the output stored in the
+step with the name ``trained_model``. Alternatively, if the plugin task has named output (perhaps named ``model``,
+it can be accessed with ``$trained_model.model``).
```
colton-lapp: A set of outstanding questions I have is how the output of a task gets mapped onto output names. 

I.e. if you have a plugin task that returns `(model, data)` as a tuple, what determines their names if you want to reference them in the task graph? Is this based on the names given during task registration? If so, is it based on the positional ordering - i.e. `outputs[0]` goes to the first registered output parameter (say, `trained_model.model`), `outputs[1]` goes to the second (`trained_model.data`), etc. What happens if you reference `$trained_model` but the output is an iterable with multiple registered output parameters? Does that pass in the tuple object? What happens if there are multiple registered output parameters (say, `trained_model.model`, `trained_model.data`, `trained_model.config_dict`) but your plugin task only returns a single item (i.e. `return model`)?

The answers to the above questions might belong in a reference document, but they could be addressed at a high level in this explainer.

docs/source/explanation/components/task-graph-explanation.rst
```diff
@@ -18,7 +18,118 @@
...
+
+
+Global variables
+~~~~~~~~~~~~~~~~
+
+Entrypoints are parameterizable, and you can also reference these parameters as variables. So, for example,
+``$num_samples`` could reference an entrypoint parameter, and be used as input to a task.
```
colton-lapp: Potential Typo: "as **an** input to a task"

docs/source/explanation/components/task-graph-explanation.rst
```diff
@@ -18,7 +18,118 @@
...
+You can reference the output of steps by referencing the step name. This is useful for passing the
+output of one step as a parameter to another. For example: ``$trained_model`` will take the output stored in the
+step with the name ``trained_model``. Alternatively, if the plugin task has named output (perhaps named ``model``,
+it can be accessed with ``$trained_model.model``).
+
+
+Global variables
+~~~~~~~~~~~~~~~~
+
+Entrypoints are parameterizable, and you can also reference these parameters as variables. So, for example,
```
colton-lapp: The phrasing "and you can also" is a little confusing to me. This is the only purpose of entrypoint parameters, right? Or is there something else you can use them before outside of variables in the task graph / artifact output graph?

docs/source/explanation/components/task-graph-explanation.rst
```diff
@@ -18,7 +18,118 @@
...
+dependency on the ``trained_model`` step, and the ``predictions`` step has an implicit dependency on the ``defended_model`` step (by using
+the output of those steps).
+
+
+.. figure:: /images/DAG.png
+   :alt: Generated directed acyclic graph based on dependencies within the task graph.
+   :figclass: border-image clickable-image 
+
+   The DAG generated from the above task graph.
+
```
colton-lapp, line 133: Consider adding a paragraph that says something like "Dioptra creates dependencies in the DAG only based off the input/output chaining of plugin tasks. If a user wants to add additional explicit dependencies in the task graph, this can be done. See task-graph-reference for more details on this.

docs/source/explanation/components/task-graph-explanation.rst
jtsextonMITRE: I think I had meant the "also" to be associated with the "variables" not the "parameters", like "parameters are referenced as variables too", referring to the previous section. I agree it's less confusing if the "also" is removed

docs/source/reference/dioptra-components/task-graph-reference.rst
```diff
@@ -0,0 +1,201 @@
...
+..
+.. ACCESS THE FULL CC BY 4.0 LICENSE HERE:
+.. https://creativecommons.org/licenses/by/4.0/legalcode
+
+.. _reference-task-graph:
+
+Task Graph
+==========
+
+The task graph consists of a list of step descriptions. These descriptions can be written using various
```
keithmanville, line 23: I think we should make a top level item called "Invocation Styles" that the next three headers fall under. Then the other sections are also top level items.

Right now everything after Mixed Style gets nested under Mixed because of how it's structured.

PR #1080, opened 2026-01-06
fix: docker frontend errors

henrychoy:
Resolves 2 issues below when running the containers.  See slack thread for details.

- When installing the frontend, we get `You are using Node.js 18.20.8.  Vite requires Node.js version 20.19+ or 22.12+.  Please upgrade your Node.js version.`   In cookiecutter.json I updated the version to `22-bullseye`
- After the node error, we get errors that typescript cant import css files.  I added css module declarations in `env.d.ts` so that typescript treats them as files.  Also I added this file in `init-frontend.sh` so that it's copied over into the build.


jkglasbrenner (approved):
LGTM, merging to dev

PR #1114, opened 2026-01-21
docs: update components reference index page

### LINKED ISSUE #1091: Reference - Dioptra Components - Dioptra Components Glossary
move this into table of contents / landing page with links to subpages.

colton-lapp:
Closes #1091 

Likely needs edits / review on the definitions.

### DIFF
--- a/docs/source/reference/dioptra-components/index.rst
+++ b/docs/source/reference/dioptra-components/index.rst
@@ -50,23 +50,21 @@ Plugins
 ~~~~~~~
 
 - ``Plugin``: A collection of Python files which contain registered ``Function Tasks`` and/or ``Artifact Tasks``.
-- ``Plugin Parameter Type``: Either a built-in or user-defined type, used for Entrypoint task graph ``type validation``. ``Function Task Inputs``, ``Function Task Outputs``, ``Artifact Tasks``, ``Entrypoint Parameters`` and the outputs from ``Artifact Input Parameters`` all have associated ``Plugin Parameter Types``.
+- ``Plugin Parameter Type``: Either a built-in or user-defined type, used for type validation. ``Function Task Inputs``, ``Function Task Outputs``, ``Artifact Task Outputs``, ``Entrypoint Parameters`` and the outputs from ``Entrypoint Artifact Parameters`` all have associated ``Plugin Parameter Types``.
 
-Function Plugins 
+Plugin Function Tasks 
 ~~~~~~~~~~~~~~~~
-- ``Function Plugin``: A Plugin that registers ``Function Tasks`` and contains those task's associated Python files. 
-- ``Function Task``: A Python function within a ``Plugin`` file that defines some computational task, used in ``Entrypoints``.
-- ``Function Task Input``: An input into a ``Function Task`` registered as part of the task definition.
-- ``Function Task Output``: The output of a ``Function Task`` registered as part of the task definition. A ``Function Task Output`` can be fed in as a ``Function Task Input`` via the ``Task Graph``, or saved as an ``Artifact`` via the ``Artifact Output Graph``.
 
-Artifact Plugins and Artifacts
+- ``Function Task``: A registered Python function within a ``Plugin`` file that defines some computational task, used in the ``Task Graph`` in ``Entrypoints``.
+- ``Function Task Input``: An input into a ``Function Task`` registered as part of the task definition.  Has an associated ``Plugin Parameter Type`` for type validation.
+- ``Function Task Output``: The output of a ``Function Task`` registered as part of the task definition.  Has an associated ``Plugin Parameter Type`` for type validation. A ``Function Task Output`` can be fed in as a ``Function Task Input`` via the ``Task Graph``, or saved as an ``Artifact`` via the ``Artifact Output Graph``.
+
+Plugin Artifact Tasks and Artifacts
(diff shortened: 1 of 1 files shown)


docs/source/reference/dioptra-components/index.rst
```diff
@@ -40,3 +42,59 @@ Dioptra Components
...
+
+Plugins
+~~~~~~~
+
+- ``Plugin``: A collection of Python files which contain registered ``Function Tasks`` and/or ``Artifact Tasks``.
+- ``Function Task``: A Python function within a ``Plugin`` file that defines some computational task, used in ``Entrypoints``.
+- ``Artifact Task``: A Python class within a ``Plugin`` file that defines how an ``Artifact`` is serialized and deserialized, used to in ``Entrypoints``.
+- ``Function Task Input``: An input into a ``Function Task`` registered as part of the task definition.
+- ``Function Task Output``: The output of a ``Function Task`` registered as part of the task definition. A ``Function Task Output`` can be fed in as a ``Function Task Input`` via the ``Task Graph``, or saved as an ``Artifact`` via the ``Artifact Output Graph``.
+- ``Plugin Parameter Type``: Either a built-in or user-defined type, used for Entrypoint task graph ``type validation``. ``Function Task Inputs``, ``Function Task Outputs``, and ``Artifact Tasks`` all have associated ``Plugin Parameter Types``.
```
colton-lapp: Entrypoint Parameters and Artifact Input parameters also have types, right? Also, types are used to determine which artifact snapshots are eligible to be loaded in to an artifact input parameter?

docs/source/reference/dioptra-components/index.rst
```diff
@@ -40,3 +42,59 @@ Dioptra Components
...
+
+Artifacts
+~~~~~~~~~
+
+- ``Artifact``: The output of a ``Function Task`` that has been saved to disk via the serialize method of an ``Artifact Task``.
+- ``Artifact Task``: A subclass of ArtifactTaskInterface that defines serialization, deserialization, and validation logic for an ``Artifact``.
```
colton-lapp: Repeated from the Plugins section - maybe that's fine

docs/source/reference/dioptra-components/index.rst
```diff
@@ -40,3 +42,59 @@ Dioptra Components
...
+- ``Artifact Task Output Parameter Type``: The ``Parameter Type`` for the object that is returned by an ``Artifact Task``.
+
+Entrypoints
+~~~~~~~~~~~
+
+- ``Entrypoint``: Define parameterizable workflows, which can be reused across multiple ``Experiments``.
+- ``Entrypoint Parameter``: An input used for the ``Task Graph`` and the ``Artifact Output Graph``. Can be customized during a ``Job`` run and have an optional default value.
+- ``Entrypoint Artifact Parameter``: A kind of ``Entrypoint Parameter``, where the value is read in from a ``Snapshot`` of an ``Artifact`` during ``Job`` execution. Can contain multiple objects.
+- ``Task Graph``: The sequence of ``Function Tasks`` that an ``Entrypoint`` executes, written in YAML. A directed, acyclic graph (DAG).
+- ``Task Graph Step``: A single step in the ``Task Graph``, which must have a name. Can have positional and keyword arguments as well.
```
colton-lapp, line 76: Consider adding to this section: "Step Name" (for the task graph)

docs/source/reference/dioptra-components/index.rst
```diff
@@ -40,3 +42,59 @@ Dioptra Components
...
+- ``Artifact Output Graph``: The logic dictating which ``Function Task Outputs`` are saved to ``Artifacts`` and how (i.e. which ``Artifact Tasks`` are invoked and their inputs). Written in YAML.
+
+Experiments/Jobs
+~~~~~~~~~~~~~~~~
+
+- ``Experiment``: A logical container that holds ``Entrypoints`` and ``Job`` runs.
+- ``Job``: A parameterized run of an ``Entrypoint``.
+- ``Queue``: A queue manager for ``Jobs``. Manages ``Job`` execution for a specific ``worker`` environment.
+- ``Worker``: Contains the resources for executing a ``Job``. Default worker containers are available for GPU and CPU hardware. Custom workers can be developed..
```
colton-lapp: ~~Delete extra period~~ FIXED

docs/source/reference/dioptra-components/index.rst
keithmanville: Should probably keep it in one place. And shouldn't have slightly different definitions.

I think I may have added it to Plugins because we reference Function and Artifact Tasks there. But if we put it there, it probably make sense to keep Artifact Task Output Parameter (and Type) next to it. And then the Artifacts section only has Artifacts...

Not sure what best here. What do you think?

Also, we could potentially drop Artifact Output Parameter Type and just add a mention to Output Parameter that it has a Plugin Param Type.

docs/source/reference/dioptra-components/index.rst
keithmanville: yes, I think it's worth adding a quick mention of types to the existing definitions.

docs/source/reference/dioptra-components/index.rst
keithmanville: You're saying there should be a separate entry for Step Name? I don't think its needed, but happy to be convinced otherwise.

PR #1117, opened 2026-01-22
docs: artifacts explainer

### LINKED ISSUE #1056: Artifacts

jtsextonMITRE:
Closes #1056

### DIFF
--- a/docs/source/explanation/components/artifacts-explanation.rst
+++ b/docs/source/explanation/components/artifacts-explanation.rst
@@ -23,7 +23,7 @@ Artifacts
 Summary: What is an artifact?
 -----------------------------
 
-An **artifact** refers to the stored outputs of jobs. A job can produce multiple artifacts,
+An **artifact** refers to a stored output of a job that is managed by Dioptra. A job can produce multiple artifacts,
 and artifacts produced by a job can be used as inputs to another job. How the artifact is used
 is specified in the entrypoint associated with the job, but the artifact itself is provided at
 runtime as an input to the job. 
@@ -56,7 +56,7 @@ This interface supports three methods:
 * ``deserialize`` - used to read the contents of an artifact with a given path relative to the specified directory
 * ``validation`` - used to validate any keyword arguments passed into ``serialize()``
 
-The functions ``serialize`` and ``deserialize`` both support passing additional keyword arguments. This can be useful
+The ``serialize`` function both supports passing additional keyword arguments. This can be useful
 for selecting file types or configuring other settings when saving an artifact.
 
 Entrypoints can designate the output of a function task as an artifact by referencing an artifact task for serialization. 
@@ -76,5 +76,6 @@ See Also
 
 * :ref:`Entrypoints: explanation <explanation-entrypoints>` - Explanation of Entrypoints, including the Artifact Task Graph.
 * :ref:`Entrypoints: reference <reference-entrypoints>` - Complete YAML syntax guide for entrypoint files and task graphs.
+* :ref:`Artifacts: reference <reference-artifacts>` - Complete reference for Artifacts.
 * :ref:`Artifact Graphs: reference <reference-artifact-graph>` - YAML syntax for artifact graph.
 * :ref:`Plugins: explanation <explanation-plugins>` - Explanation of Plugins, Function Tasks and Artifact Tasks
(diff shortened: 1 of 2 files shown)


docs/source/explanation/components/artifacts-explanation.rst
```diff
@@ -18,6 +18,40 @@
...
 
-[Todo] Insert an explanation about artifacts.
\ No newline at end of file
+Summary: What is an artifact?
+-----------------------------
+
+An **artifact** refers to the stored outputs of jobs. A job can produce multiple artifacts,
+and artifacts produced by a job can be used as inputs to another job. How the artifact is used
+is specified in the entrypoint associated with the job, but the artifact itself is provided at
+runtime as an input to the job.
```
colton-lapp: Maybe consider specifically mentioning how an artifact can be made available to the task graph in the exact same way as an entrypoint parameter. Something like "Artifacts are used in entrypoints through the 'Artifact Input Parameters' portion of an entrypoint. Once an artifact is designated as an input parameter, it can be referenced in the task graph in the exact same way any regular entrypoint parameter or task output can be. The artifact is loaded into memory at the start of the job execution and then is available for any tasks that reference it."

docs/source/explanation/components/artifacts-explanation.rst
```diff
@@ -18,6 +18,40 @@
...
+
+
+Artifact Tasks
+--------------
+
+**Artifact tasks** are a type of plugin task which primarily serve to detail the serialization
```
colton-lapp: Consider deleting the word "primarily" because this is the only thing artifact tasks do? They also do validation, but that's in service of the serialization / deserialization

docs/source/explanation/components/artifacts-explanation.rst
```diff
@@ -18,6 +18,40 @@
...
+and deserialization of a given artifact type. When an output of a function task is designated to be 
+saved as an artifact, it is passed to its corresponding serialization function within the artifact task.
+Similarly, when an artifact is loaded, its deserialization function is used to load it as an object in memory.
+
+Because the serialization input and deserialization output types must be the same, passing artifacts between
+entrypoints is effectively transparent to the entrypoint - the entrypoint can handle those artifacts
+as if they are an object loaded into memory, and ignore the details of reading/writing, as long as artifact
+tasks for that type exist.
```
colton-lapp: This paragraph is a little confusing to me. Not sure what "effectively transparent" means in this context. Is this paragraph trying to say that by separating concerns of computation and reading/writing, the entrypoint task graph can treat artifact input parameters identically to entrypoint parameters and task outputs, because the artifact task handles all the logic needed to get an object written to disk in the proper available state for the task graph? If that is the main point of this paragraph, I'm not sure I understand the first sentence "because the serialization input and deserialization output types must be the same" - that seems to suggest something about the portability of an artifact between entrypoints.

docs/source/explanation/components/artifacts-explanation.rst
```diff
@@ -18,6 +18,40 @@
...
+
+Note that when artifacts are created, they are associated with a snapshot of the artifact task that they were 
+created with. Since the artifact task contains both serialization and deserialization, the same snapshot is used
+for deserialization. 
+
```
colton-lapp, line 68: Consider adding descriptions about the following:
- The pathing of artifacts (i.e. where are they saved locally). Is it in a temporary directory? How are the filenames determined? How does Dioptra remember the filepaths when it comes time to deserialize and artifact (I assume it is saved in Dioptra's backend during serialization)
- The validation method - what is it used for? 
- Why you would pass arguments to the serialize / deserialize methods, maybe with some examples. For example, if you created a pandas dataframe artifact handler, maybe the input parameters specify the file format (.csv, pickle, parquet). 
- Something about type alignment when using an artifact as an input parameter - i.e. "To use an artifact as an entrypoint input parameter, the types from the deserialize methods need to match up with the types from the artifact input parameter. You will not be able to select any saved artifact for the job if these types don't align" or something
- A summary of instructions for creating your own artifact handlers - this can be brief though and could reference the plugin explainer
- Some more details about multiple inputs / outputs for an artifact. For example, can you have a serialize method save an image file as both a .PNG and an underlying Python object, but then have the deserialize method return only the underlying Python object? (this is something I'm trying to do in a tutorial but am not sure how to accomplish / if it's possible)

keithmanville (approved):

Issue #981, opened 2025-09-25
[BUG] In UI, entrypoint params without default value cause internal server error 500


colton-lapp:
**Describe the bug**
In the UI, when users make an entrypoint that has an entrypoint parameter _without_ a default value, this causes the job to fail and the "validate input" button also fails silently. This is true even if the entrypoint parameter is not used in the task graph. 

**To Reproduce**
1. Create an entrypoint
2. Create an entrypoint parameter with no default value
3. Run a job and observe that the logs show "Internal server error 500"

**Expected behavior**
Entrypoints should be able to have parameters with no default, especially if a value is provided during job run (this fails as well). 

**Desktop:**
 - OS: Windows WSL 
 - Browser: Chrome

**Server**
- OS: Ubuntu
- Virtualization: Docker

**Dioptra branch**:
Dev (pulled on 9/24/2025)

**Additional context**
Here are the full logs for the failure:
```bash
2025-09-25T15:11:26.370725Z [error    ] Job failed.                    [dioptra.rq.tasks.run_v1_dioptra_job] experiment_id=12 job_id=28
Traceback (most recent call last):
  File "/opt/venv/lib/python3.11/site-packages/dioptra/sdk/utilities/run_dioptra_job.py", line 156, in main
    job_yaml = dioptra_client.entrypoints.snapshots.get_config(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/dioptra/client/entrypoints.py", line 401, in get_config
    return self._session.get(
           ^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/dioptra/client/sessions.py", line 721, in get
    return convert_response_to_dict(self._get(endpoint, *parts, params=params))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/dioptra/client/sessions.py", line 126, in convert_response_to_dict
    raise StatusCodeError(response.status_code, response.text)
dioptra.client.base.StatusCodeError: Error code returned: 500
Error message returned: 
{
    "message": "Internal Server Error"
}
```

PR #1123, opened 2026-01-23
feat: allow null for entrypoint param value

henrychoy:
Deals with frontend part of #981 

In Entrypoint form, allow users to set null (None) for entrypoint parameter value.  Also allow them to switch back to regular value or empty string.

### DIFF
--- a/src/dioptra/restapi/bootstrap.py
+++ b/src/dioptra/restapi/bootstrap.py
@@ -23,6 +23,7 @@ from typing import Any, Callable, List, Optional
 from boto3.session import Session
 from botocore.client import BaseClient
 from injector import Binder, Module, provider
+from mlflow import MlflowClient
 from passlib.context import CryptContext
 from redis import Redis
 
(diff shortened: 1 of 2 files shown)


src/dioptra/task_engine/type_validation.py
```diff
@@ -831,6 +831,16 @@ def _check_global_parameter_defaults(
         A list of ValidationIssue objects; will be empty if no issues were
             found
     """
+
+    print("=== _check_global_parameter_defaults INPUTS ===")
```
keithmanville: remove debugging print statements.

src/dioptra/task_engine/validation.py
```diff
@@ -265,6 +265,20 @@ def _check_global_parameter_types(
     type_defs = experiment_desc.get("types", {})
     type_names = type_registry.BUILTIN_TYPES.keys() | type_defs.keys()
 
+    print("=== _check_global_parameter_types: global_parameters ===")
```
keithmanville: remove debugging print statements.

Issue #1000, opened 2025-10-27
[BUG] 500 Internal Server Error on resource import, if you use try to use an artifact plugin not declared in the TOML


jtsextonMITRE:
In an entrypoint, if you try to use the `artifacts` plugin like this, but don't have the plugin declared in the TOML somewhere, you get a 500 error.

```
[[entrypoints]]
path = "entrypoints/attack.yml"
name = "Evasion Attack"
description = "Create an adverarial dataset"
params = [
...
]
plugins = [ "assess" ]
artifact_plugins = [ "artifacts" ]
```

```
2025-10-27T17:29:51.744679Z [error    ] Exception on /api/v1/workflows/resourceImport [POST] [dioptra.restapi.app]
Traceback (most recent call last):
  File "/opt/venv/lib/python3.11/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/flask_restx/api.py", line 404, in wrapper
    resp = resource(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/dioptra/restapi/utils.py", line 373, in new_view_func
    return view_obj.dispatch_request(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/flask_restx/resource.py", line 41, in dispatch_request
    resp = meth(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/flask_login/utils.py", line 290, in decorated_view
    return current_app.ensure_sync(func)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1125, opened 2026-01-27
fix(restapi): capture 500 error

jtsextonMITRE:
Address #1000.

In the future, it may be desirable as a feature to have it be possible to use artifacts that are not directly registered (or are declared, but not defined) in the TOML - for example, if an external GitHub repository wants to somehow use some of the artifact plugins in dioptra's extra folder.

### DIFF
--- a/src/dioptra/restapi/v1/workflows/service.py
+++ b/src/dioptra/restapi/v1/workflows/service.py
@@ -725,7 +725,7 @@ class ResourceImportService(object):
                     log=log,
                 )
 
-    def _register_entrypoints(
+    def _register_entrypoints(  # noqa: C901
         self,
         group_id: int,
         entrypoints_config: list[dict[str, Any]],


src/dioptra/restapi/v1/workflows/service.py
```diff
@@ -794,13 +794,32 @@ def _register_entrypoints(
...
+                raise ImportFailedError(
+                    f"Plugins {errors}",
+                    reason="Plugin used but undeclared.",
+                )
+
+            errors = []
+            artifact_plugin_ids: list[int] = []
+            for plugin in entrypoint.get("artifact_plugins", []):
+                try:
+                    plugin_ids.append(plugins[plugin].resource_id)
```
keithmanville: bug: should be appending to artifact_plugin_ids.

src/dioptra/restapi/v1/workflows/service.py
```diff
@@ -794,13 +794,32 @@ def _register_entrypoints(
...
+                except KeyError:
+                    errors += [plugin]
```
keithmanville: use `.append`

src/dioptra/restapi/v1/workflows/service.py
```diff
@@ -794,13 +794,32 @@ def _register_entrypoints(
...
-                for plugin in entrypoint.get("artifact_plugins", [])
-            ]
+
+            errors = []
+            plugin_ids: list[int] = []
+            for plugin in entrypoint.get("plugins", []):
+                try:
+                    plugin_ids.append(plugins[plugin].resource_id)
+                except KeyError:
+                    errors += [plugin]
```
keithmanville: use `.append`

src/dioptra/restapi/v1/workflows/service.py
```diff
@@ -794,13 +794,32 @@ def _register_entrypoints(
...
+            if len(errors) > 0:
+                raise ImportFailedError(
+                    f"Plugins {errors}",
```
keithmanville: better to have more informative message, since reason is not shown in UI currently.

PR #1126, opened 2026-01-27
fix: update "add to task graph" button for tasks with no params

### LINKED ISSUE #1002: [BUG] "add to task graph" button adds invalid yaml for tasks with no parameters
On the Entrypoint create/edit page, the "add to task graph" button adds invalid yaml for tasks with no parameters.

tasks with no parameters must use the mixed style invocation documented here: https://pages.nist.gov/dioptra/reference/task-engine-reference.html#id26
```
graph:
    step1:
        task: plugin1
        args: [posarg1, posarg2]
        kwargs:
            keyword1: arg1
            keyword2: arg2
```

Update the UI to use the mixed style invocation for tasks with no parameters.

henrychoy:
Closes #1002 

In the entrypoint form, for function tasks, the "add to task graph" button will now use [Mixed Style Invocation](https://pages.nist.gov/dioptra/reference/task-engine-reference.html#mixed-style-invocation) to accomidate tasks with and without parameters.

```
<step-name>:
  task: hello_world
  kwargs:
    greeting: <input-value>
    name: <input-value>
<step-name>:
  task: taskWithNoParams
```

### DIFF
--- a/src/frontend/src/views/CreateEntryPoint.vue
+++ b/src/frontend/src/views/CreateEntryPoint.vue
@@ -906,7 +906,7 @@
   }
 
   function addToArtifactGraph(task) {
-    let string = `<output-name>:\n  contents: <contents>\n  task:\n    args:\n    name: ${task.name}`
+    let string = `<output-name>:\n  contents: <contents>\n  task:\n    name: ${task.name}`
     if(entryPoint.value.artifactGraph.trim().length === 0) {
       entryPoint.value.artifactGraph = string
     } else {


henrychoy:
> I think from a clarity point of view, defaulting to the mixed invocation method would be clearer for novices, i.e. :

Ok just updated it to use mixed invocation for all function tasks to accommodate all tasks with or without params.  Didnt include the args line because params will always be named in the UI, let me know if that sounds right.

> It would be great to detect when artifact tasks have no input parameters and automatically remove the `args` line as well for the artifact tasks.

I don't understand the artifact graph in general yet, but the artifact tasks themselves only list output params?  Also in the example I was given, args seem to reference the input params of the python function?  So not really clear exactly when to remove args?

```
my_artifact:
  contents: $message
  task:
    name: StringArtifactTask
    args:
      type: txt
```

PR #1127, opened 2026-01-27
docs: update notebooks tutorials page

### LINKED ISSUE #1031: OPTIC - ML Attacks

keithmanville:
closes #1031

### DIFF
--- a/docs/source/how-to/import-content/add-a-dataset.rst
+++ b/docs/source/how-to/import-content/add-a-dataset.rst
@@ -21,7 +21,7 @@ Add a Dataset
 =============
 
 To make datasets or other resources available when running experiments, they must be placed in a directory accessible to your workers.
-This directory can be stored anywhere on your host machine's filesystem and then (:ref:`mounted into the worker containers <running-dioptra-mounting-folders-in-the-worker-containers>`).
+This directory can be stored anywhere on your host machine's filesystem and then :ref:`mounted into the worker containers <how-to-data-mounts>`.
 
 Once the dataset directory is configured as part of the Dioptra deployment, adding data is as simple as placing the data into that directory.
 Data added to the datasets directory is immediately accessible to plugin tasks by referencing the ``/dioptra/data`` location.
@@ -44,4 +44,8 @@ Then, to download and add a dataset directly to the ``/dioptra/data`` directory,
 
    uv run ./examples/scripts/download_data.py download --data-dir /dioptra/data DATASET_NAME
 
-For the full list of options, run ``uv run ./examples/scripts/download_data.py -h`` to display the script's help message.
+For the full list of options, run the following to display the script's help message:
+
+.. code-block:: sh
+
+    uv run ./examples/scripts/download_data.py -h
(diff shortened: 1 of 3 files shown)


colton-lapp:
In general, this looks good to me - I like the table. Three thoughts you can feel free to reject:

1. Including the Jupyter Notebook as a sphinx page using an extension like nbsphinx 
image_720
image_720

2. Including a how-to / setup / requirements for running these notebooks a little more explicitly. Something like:

> In order to recreate these tutorials, you'll need to do the following:
> 1. Install Dioptra, including cloning the repository locally
> 2. Locate the examples folder in your local installation, which has Jupyter notebooks
>3. Set up your virtual environment (insert UV instructions, unless this is done in installation)
>4. Follow along in the Jupyter notebook. 

I guess we already have something like this in the examples readme (https://github.com/usnistgov/dioptra/blob/db5f405754632808b60297315493560b8baa18f1/examples/README.md) but I'm not sure if that is current. I think a beginning user might be a little intimidated just by clicking the link to the notebook in github, but maybe that's okay because it's for advanced users 

3. The index title "Notebook Tutorials" is a little bit uninformative to a user. We could maybe change it to "Advanced Tutorials"? Also, unless we are immediately planning on adding more notebook tutorials, we could make the index page specifically for the OPTIC example - that would match the pattern of the first two tutorials a little better.

docs/source/tutorials/advanced/index.rst
```diff
@@ -0,0 +1,74 @@
...
+  :header-rows: 1
+
+  * - Name
+    - Description
+  * - `Hello World with the Dioptra Python Client <https://github.com/usnistgov/dioptra/blob/main/examples/hello_world.ipynb>`__
+    - This tutorial demonstrates how interact with Diopta via the Python Client. It walks through registering a queue,
+      importing plugins and entrypoints, creating an experiment, and finally executing jobs.
+  * - `Adversarial ML with OPTIC <https://github.com/usnistgov/dioptra/blob/dev/examples/mnist_demo.ipynb>`__
+    - This tutorial demonstrates how the OPTIC (Open Perturbation Testing for Image Classifers) plugin can be used to
+      evaluate adversarial attacks and defenses on a model trained to recognize handwritten digits.
```
colton-lapp, line 42: One of these links to main (broken at the moment, but should be fixed after this is merged) and one links to dev. Was that intended?

docs/source/how-to/import-content/add-a-dataset.rst
```diff
@@ -17,8 +17,8 @@
 
 .. how-to-add-a-dataset:
 
-How-To: Add a Dataset
-=====================
+Add a Dataset
+=============
 
 To make datasets or other resources available when running experiments, they must be placed in a directory accessible to your workers.
 This directory can be stored anywhere on your host machine's filesystem and then (:ref:`mounted into the worker containers <running-dioptra-mounting-folders-in-the-worker-containers>`).
```
colton-lapp: Broken link - did you mean? :ref:`how-to-data-mounts` ? Also the parentheses breaks up the sentence a little awkwardly at least at the moment

examples/README.md
colton-lapp: "to use to manage" is kind of awkward

docs/source/tutorials/advanced/index.rst
keithmanville: good catch. I meant to link to main (which will work once we release to main).

PR #1128, opened 2026-01-27
docs: update python client how-to

keithmanville:


colton-lapp:
@keithmanville I went ahead and progressed this page a little more. To answer questions you raised in the standup:

- How are we bringing in Python code? 
    - I think we settled on defining code in directories and bringing it in using `literalincludes` as a general pattern, but with code snippets that aren't repeated across pages and are really short (as these code snippets are), I'm not sure that makes sense. I just used `code-block` elements to define code directly in the `.rst` document. Something else we _could_ do is make a singular python file with all of these commands, and then use comments to demarcate sections of that python file which can be brought in individually (I think you can use line numbers as well). But again, these are mostly single lines of code so that seems a bit excessive. 

Another question I have is around documenting client methods. In other how-tos, we've been defaulting to using `automethod` to pull the docstrings from the client methods. While this generally is good, it is not as easy to follow as showing the _exact_ lines to run in Python. The part that is confusing is the client methods live in resource specific files (i.e. `dioptra.client.users.UsersCollectionClient`), but they are accessed via something like `client.users`. When people look at the `automethod` for the docstring, they only see `UsersCollectionClient`. As a result, it might no be obvious how to call these methods exactly. In this setup guide, I show the exact Python syntax (i.e. `client.users.get()`) and then I show the `automethod`. Let me know what you think of this. A potential solution to this would be to put an example invocation method in the docstrings for these methods, i.e. in the method (`dioptra.client.users.UsersCollectionClient.get` method, put example usage of `client.users.get`).

Finally, I added some common client methods at the end of the setup guide that are likely to be used by downstream tasks. Feel free to remove these if you want.

docs/source/how-to/essential-workflows/setup-python-client.rst
```diff
@@ -36,21 +36,238 @@ Client Setup Workflow
...
+        .. admonition:: Register User
+            :class: code-panel python
+
+            .. code-block:: python
+
+                username = "user"
+                email = "user@localhost"
+                password = "pass"
+                response = client.users.create(username, email=email, password=password)
+                print(response.json
```
colton-lapp: Missing closing parenthetical

docs/source/how-to/essential-workflows/setup-python-client.rst
```diff
@@ -36,21 +36,238 @@ Client Setup Workflow
...
+
+                client.queues.get()
+
+        *Full Client Method Documentation:*
+
+        .. automethod:: dioptra.client.queues.QueuesCollectionClient.get
+
+    .. tab:: Get Parameter Type IDs
+
+        .. admonition:: Get Queue IDs
```
colton-lapp: Wrong header - sorry this was my fault!

docs/source/how-to/essential-workflows/setup-python-client.rst
```diff
@@ -36,21 +36,238 @@ Client Setup Workflow
...
-- Navigate to deployment folder 
-- Run ``docker compose up``
+Set the environment variable for the Dioptra REST API address:
+
+.. admonition:: Initialize Client
+    :class: code-panel console
+
+    .. code-block:: bash
+
+        export DIOPTRA_API="<host>:<port>"
```
colton-lapp: While I think most users will understand what this means, a tiny bit more information about what "setting an environment variable" might be useful for less technical users. Also a note about what the default host/port is would probably be good, potentially linking out to the install guide

docs/source/how-to/essential-workflows/setup-python-client.rst
```diff
@@ -36,21 +36,238 @@ Client Setup Workflow
...
+   .. group-tab:: JSON Client
+
+        .. admonition:: Initialize Client
+            :class: code-panel python
+
+            .. code-block:: python
+
+                from dioptra.client import connect_json_dioptra_client
+
+                client = connect_json_dioptra_client()
```
colton-lapp, line 74: can the Dioptra address be passed as an arg to these methods? If it can, and we don't want to discourage this pattern, we could maybe note that? I could maybe see there being use cases when quickly changing the address would be useful and it could be tedious to set it via environment variables...

colton-lapp (approved):

PR #1199, opened 2026-02-25
refactor: add entrypoint repository

keithmanville:


src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -378,22 +278,19 @@ def get(
...
+                entrypoint_snapshot_id, DeletionPolicy.NOT_DELETED
             )
-        )
-        entrypoint = db.session.scalars(stmt).first()
+
+        if entrypoint is None:
+            raise EntityDoesNotExistError(RESOURCE_TYPE, entrypoint_id=entrypoint_id)
 
         if entrypoint is None:
             raise EntityDoesNotExistError(
```
jkglasbrenner: These guards are repetitive and besides, are they actually needed? Isn't the purpose of `get_one` to raise an error if something can't be found?

Also, maybe this is a something I missed, but seeing `error_if_not_found` on the `get_one` method I thought isn't something we wanted to keep doing? If I recall, the split between `get` and `get_one` was so that we didn't need an error_if_not_found argument anymore. If that pattern still exists elsewhere on the other repos, we should verify if it's actually needed anymore.

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -402,24 +299,24 @@ def get(
...
+                if resource.resource_type == "queue" and not resource.is_deleted
+            }
+        )
+        queues = self._uow.queue_repo.get(queue_ids, DeletionPolicy.NOT_DELETED)
+
+        if len(queues) != len(queue_ids):
+            queue_ids_missing = set(queue_ids) - {queue.resource_id for queue in queues}
+            log.debug("Queue not found", queue_ids=list(queue_ids_missing))
+            raise EntityDoesNotExistError("queue", resource_ids=queue_ids_missing)
+
```
jkglasbrenner: This is something we need to extend in the repo pattern so that we don't have to do this kind of iteration inline. This is probably a pattern we'll see elsewhere, so it's something about working with resource children. There's two versions of this, what happens here that is just based on resource_id, and children that are linked by snapshot ids, if I recall correctly.

That way we embed all of that validation logic too, or at least have a single place to deal with deleted policy checks. At this point, I also don't remember why these lengths would ever differ, something to think about as we migrate this type of logic into the repo pattern.

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -463,29 +360,34 @@ def modify(
...
-            raise QueryParameterNotUniqueError(RESOURCE_TYPE, name=duplicates)
+
+        entrypoint = self._uow.entrypoint_repo.get_one(
+            entrypoint_id, DeletionPolicy.ANY
+        )
+
+        if not entrypoint:
+            raise EntityDoesNotExistError("entrypoint", resource_id=entrypoint_id)
+        elif entrypoint.resource.is_deleted:
+            raise EntityDoesNotExistError("entrypoint", resource_id=entrypoint_id)
```
jkglasbrenner: Why not just let the `get_one` raise an error?

As a general pattern, we should try and move existence checks like this out of the service layer and into the repository pattern. The only time I feel checks like this should remain in the service layer are either:

1. Some of the services haven't been migrated into the repository pattern yet, so it's for backwards compatibility.
2. There's something about the coordination across two or more resources that requires special checks that cannot be expressed in our data retrieval logic. My guess it would be tied to some sort of request-level state that the database layer should not be aware of. At the moment, I can't think of what that might be, but it's possible we'll find one.

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -463,29 +360,34 @@ def modify(
...
+        if len(entrypoint_param_duplicates) > 0:
+            raise QueryParameterNotUniqueError(
+                RESOURCE_TYPE, name=entrypoint_param_duplicates
+            )
+
         artifact_parameter_duplicates = find_non_unique("name", artifact_parameters)
         if len(artifact_parameter_duplicates) > 0:
             raise QueryParameterNotUniqueError(
                 RESOURCE_TYPE, name=artifact_parameter_duplicates
             )
```
jkglasbrenner: These aren't query parameters if they're not embedded in the URL making the request.

Could we potentially move this logic into the Marshmallow schema as a pre- or post-hook? This is about validating user input, and it isn't tied to anything in the database.

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -463,29 +360,34 @@ def modify(
...
-            if duplicate is not None:
-                raise EntityExistsError(
-                    RESOURCE_TYPE, duplicate.resource_id, name=name, group_id=group_id
-                )
-
-        queues = self._queue_ids_service.get(queue_ids, error_if_not_found=True)
+        if len(queues) != len(queue_ids):
+            queue_ids_missing = set(queue_ids) - {queue.resource_id for queue in queues}
+            log.debug("Queue not found", queue_ids=list(queue_ids_missing))
+            raise EntityDoesNotExistError("queue", resource_ids=queue_ids_missing)
```
jkglasbrenner: Same comment as earlier. As a heads up, I'm not going to mark every instance of this going forward to try and cut down on the comment noise, but anywhere this pattern appears, the same comment applies.

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -499,30 +401,35 @@ def modify(
...
-        all_plugin_resources = _deduplicate_plugin_resources(
-            plugin_resources, artifact_plugin_resources
-        )
 
-        new_entrypoint.children = all_plugin_resources + queue_resources
+        all_plugins = _deduplicate_plugins(plugins, artifact_plugins)
+        new_entrypoint.children.clear()
+        new_entrypoint.children.extend(plugin.resource for plugin in all_plugins)
+        new_entrypoint.children.extend(queue.resource for queue in queues)
+
```
jkglasbrenner: These might be something we can move into the repository pattern when we get plugin resources migrated. Obviously for now that's not something we can do anything about, but good to note for later.

At the end of this, there's also resource children manipulation again, so this is another part of a pattern that probably needs to be encoded more generally somehow.

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -499,30 +401,35 @@ def modify(
...
         if commit:
-            db.session.commit()
-            log.debug(
-                "Entrypoint modification successful",
-                entrypoint_id=entrypoint_id,
-                name=name,
-                description=description,
-            )
+            self._uow.commit()
+
```
jkglasbrenner: This looks off and potentially is issuing commit twice?

If this migration to the repo pattern gets rid of EntryPointIdService being injected into other services, then the `if commit` pattern is something we should aim to remove.

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -773,7 +667,7 @@ def append(
         db.session.add(new_entrypoint)
```
jkglasbrenner: This isn't using the repository pattern, but should.

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -787,25 +681,25 @@ def append(
...
-
-        all_plugin_resources = _deduplicate_plugin_resources(
-            plugin_resources, artifact_plugin_resources
-        )
-
-        new_entrypoint.children = all_plugin_resources + queue_resources
+        new_entrypoint.children.extend(queue_resources)
 
         if commit:
             db.session.commit()
```
jkglasbrenner: Is this service class being injected anywhere? If not, then we should make the service always issue commits.

This comment applies to anywhere else in this file that uses the `if commit:` pattern. The rule should be, always commit, only retain the if guard if the class is being injected elsewhere.

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -926,30 +803,31 @@ def delete(
         )
         db.session.add(new_entrypoint)
```
jkglasbrenner: Creating a new entrypoint snapshot, so this should be handled via the repository pattern.

This comment applies to anywhere else that has this pattern.

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -1283,49 +1156,29 @@ def get(
...
             entrypoint_ids_missing = set(entrypoint_ids) - {
                 entrypoint.resource_id for entrypoint in entrypoints
             }
+            log.debug(
+                "Entrypoint not found", entrypoint_ids=list(entrypoint_ids_missing)
+            )
             raise EntityDoesNotExistError(
-                RESOURCE_TYPE, entrypoint_ids=list(entrypoint_ids_missing)
+                "entrypoint", resource_ids=entrypoint_ids_missing
             )
```
jkglasbrenner: This seems like logic that should be more embedded within the repository pattern.

src/dioptra/restapi/db/repository/utils/resources.py
```diff
@@ -238,6 +239,80 @@ def get_one_latest_snapshot(
     return latest
 
 
+def get_one_snapshot(
```
jkglasbrenner, line 329: I'm supposing this is the first time that retrieving snapshots that are not the latest appears in the service layer we're migrating?

src/dioptra/restapi/db/repository/queues.py
jkglasbrenner, line 1: I'm not sure I understand why `error_if_not_found` is suddenly being added here when it didn't exist before. Could be related to my other comments, but in general the `get_one` methods were supposed to remove the ambiguity of returning `Resource | None`.

jkglasbrenner (changes requested):
Gave this a review, focusing specifically on the service layer. I think this is a good initial start, but I think there's more we can do with the repository pattern to generalize and shorten more of the code here. Also, there were a couple of instances of behavior that looks like a regression (restoring the `error_if_not_found` argument for example).

src/dioptra/restapi/v1/entrypoints/service.py
keithmanville: You are correct. I missed that `get_one` already includes this check, but I wanted to avoid a bunch of checks in the service layer so I added back this flag. I realized this later when going through jobs/artifacts, but hadn't gone back to entrypoints to fix it yet.

src/dioptra/restapi/v1/entrypoints/service.py
keithmanville: I don't think its issuing the commit twice. create_snapshot doesn't commit. This is the pattern used by the repos Andy implemented.

I'll see if I can get rid of `if commit`

src/dioptra/restapi/v1/entrypoints/service.py
keithmanville: I was able to move this validation to the schema.

`QueryParameterNotUniqueError` was a subclass of `QueryParameterValidationError`. The latter is actually being raised for query param validation errors. I separated out the NonUnique error into its own (non-child) class and renamed it `InputParameterNotUniqueError`. 

I removed the checks in the service layer since it is now handled by the schema. The Plugin service also has similar checks. I left those as is for now (only refactored the name). That can be handled as part of the Plugin repository refactor.

The uses of `QueryParameterValidationError` are used in a few places in both controller and service layers. These can likely be moved to a schema validation as well, but I'm leaving that as a separate task.

src/dioptra/restapi/v1/entrypoints/service.py
keithmanville: I copied this pattern from the experiments endpoint. I believe the intent was to differentiate between does not exist and deleted, and potentially raise different errors in the future.

I changed it to `get_one` with `NOT_DELETED` as the policy and let it raise the error.

src/dioptra/restapi/v1/entrypoints/service.py
keithmanville: The `DraftCommitService` (used in draft commit workflow) uses `create` and `modify` of all the Resource services with `commit=False`.

The import workflow also uses the Entrypoint service `create` with `commit=False`

src/dioptra/restapi/db/repository/utils/resources.py
keithmanville: That's correct, from what I could determine, this wasn't needed yet for the other repositories.

src/dioptra/restapi/v1/entrypoints/service.py
keithmanville: We want to distinguish between `EntityDoesNotExist` and `EntityDeleted` errors, but both of these should be raised within the repository.

src/frontend/src/views/CreateEntryPoint.vue
henrychoy: this part cant simply be entryPoint.value = ep.  This is causing the banner to disappear when viewing snapshots of a deleted entrypoint because the deleted property wasnt being copied over.

src/dioptra/restapi/db/repository/utils/resources.py
```diff
@@ -174,6 +175,97 @@ def get_latest_snapshots(
...
+    """
+    snaps = get_latest_snapshots(session, snap_class, resources, deletion_policy)
+
+    if len(snaps) != len(list(resources)):
+        missing_resource_ids = set(resources) - {
+            get_resource_id(resource) for resource in resources
+        }
+        raise e.EntityDoesNotExistError(
+            EntityType.NONE, resource_ids=missing_resource_ids
+        )
```
jkglasbrenner: I can see a couple of issues with this. The first is that `set(resources)` is ambiguous because `resources` can either be an integer, a Resource, or a ResourceSnapshot. The second problem is that, even if `set(resources)` resolves to the integer id, `missing_resource_ids` will always evaluate to an empty set (it's basically `A - A`). What we want is the set of requested ids minus the list of found ids.

Also, another possible issue is that `resources` in the code gets consumed up to three times. So far it's fine because we always pass a list when this function gets used. However, if we take the type annotation seriously, then we should try and avoid iterating over resources more than once in the function body.

Here's a suggested fix that should address these problems.

```suggestion
    resource_ids = [get_resource_id(r) for r in resources]
    snaps = get_latest_snapshots(session, snap_class, resource_ids, deletion_policy)
    if len(snaps) != len(resource_ids):
        missing = set(resource_ids) - {snap.resource_id for snap in snaps}
        raise e.EntityDoesNotExistError(EntityType.NONE, resource_ids=missing)
```

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -172,45 +126,44 @@ def create(
...
-                "Entrypoint registration successful",
-                entrypoint_id=new_entrypoint.resource_id,
-                name=new_entrypoint.name,
+        with self._uow(commit):
+            self._uow.entrypoint_repo.create(new_entrypoint)
+            queues = self._uow.entrypoint_repo.create_queues(
+                new_entrypoint, queues=queue_ids
             )
+            self._uow.entrypoint_repo.create_plugins(new_entrypoint, plugins=plugins)
+
```
jkglasbrenner, line 154: So, this might be okay, but I want to articulate a tension I'm noticing. If we were to just follow the pattern, I would expect to see a `self._uow.entrypoint_repo.create_artifact_plugins()` call in the `with self._uow(commit)` block to compliment the `self._uow.entrypoint_repo.create_plugins()` call. But we don't. The main issue is making sure the right list of plugin ids end up in the `resource_dependencies` table. I _think_ this is okay, because the situation can be stated as, "all artifact plugins are plugins, but not all plugins are artifact plugins."

The reason I think this is hard to keep straight is because, in terms of the ORM classes, there's only a Plugin class. We use `EntryPointArtifactPlugin` and `EntryPointPlugin` association tables as the means for distinguishing the difference.

This leads to needing to have logic like this around line 1005 later in this same file:

```python
            # if the removed plugin is also an artifact plugin do not remove the
            # resource dependency relationship
            if artifact_plugin_id not in plugins:
                self._uow.entrypoint_repo.unlink_plugin(
                    new_entrypoint, artifact_plugin_id
                )
```

I know this isn't really a repo pattern thing, but I am finding the bookkeeping difficult to reason about. And based on the above, you could end up with an error if you never attach something as a regular plugin but only as an artifact plugin (I don't know if in practice this is possible).

My suggestion is we need to think up a unit test that could catch this if it is actually a problem.

src/dioptra/restapi/db/repository/utils/resources.py
```diff
@@ -543,24 +739,75 @@ def unlink_child(
...
-    assert_resource_exists(session, child, DeletionPolicy.ANY)
+    parent = get_one_resource(session, parent, DeletionPolicy.NOT_DELETED)
+    child = get_one_resource(session, child, DeletionPolicy.ANY)
+
+    if child.resource_type != child_resource_type.db_table_name:
+        raise e.EntityDoesNotExistError(
+            child_resource_type, resource_id=child.resource_id
+        )
+
+    parent.children.remove(child)
```
jkglasbrenner: The behavior for this was that this was a no-op if there isn't a parent/child relationship. Switching this to `.remove` changes this behavior, and this function will raise a `ValueError` if the child isn't in the parent's children list. Simple example of this:

```python
a = [1, 2, 3]
# Works, no error
a.remove(3)
# Will raise a ValueError
a.remove(4)
```

Practically speaking, what will happen now is you'll get an uninformative error when you run an operation like `DELETE /entrypoints/<id>/queues/<unrelated_queueId>`.

We should add a membership check or turn this into an operation that won't raise an error. An alternative that could work without having to restore the original index-based loop approach:

```python
child_id = get_resource_id(child)
parent.children = [x for x in parent.children if get_resource_id(x) != child_id]
```

Either that, or we can do this, which works but is perhaps harder to understand at a glance:

```python
try:
    parent.children.remove(child)

except ValueError:
    return None
```

src/dioptra/restapi/v1/experiments/service.py
```diff
@@ -476,13 +351,12 @@ def append(
             self._uow.commit()
```
jkglasbrenner: Any reason we can't turn this into the following?

```python
with self._uow(commit=True):
    all_children = self._uow.experiment_repo.add_entrypoints(
        experiment_id,
        entrypoint_ids,
    )
```

That would let us get rid of the commit argument. If this is still being injected somewhere else, we could at least do:

```python
with self._uow(commit):
    all_children = self._uow.experiment_repo.add_entrypoints(
        experiment_id,
        entrypoint_ids,
    )
```

Right?

This also applies to the modify method below.

src/dioptra/restapi/v1/experiments/service.py
```diff
@@ -541,7 +411,7 @@ def delete(self, experiment_id: int, **kwargs) -> dict[str, Any]:
 
         entrypoint_ids = [child.resource_id for child in experiment.children]
 
-        with self._uow:
+        with self._uow(commit=True):
             experiment.children = []
```
jkglasbrenner: This is a bare ORM operation, is there a clear children method that we can use?

src/dioptra/restapi/v1/experiments/controller.py
jkglasbrenner, line 1: `error_if_not_found=True` is still being used in places that it's no longer needed. The only reason it's not raising an error is because `**kwargs` is swallowing it silently.

Lines 399, 419, 457, 477, 528, 588, 602, 614

src/dioptra/restapi/db/repository/entrypoints.py
```diff
@@ -0,0 +1,592 @@
...
+            deletion_policy: Whether to look at deleted entrypoints,
+                non-deleted entrypoints, or all entrypoints
+
+        Returns:
+            A list of EntryPoint objects matching the given IDs
+
+        Raises:
+            EntityDoesNotExistError: if any of the entrypoints do not exist in the
+                database (according to deletion policy)
+        """
```
jkglasbrenner: We'll want to remove the comment to the right of `get_exact` on line 151 that contains the different names we considered for this method.

I'd recommend adding some additional clarification below the first line to better signal what this is for. It's purpose is alluded to in the `Raises:` section, but it's kind of buried and easy to miss. Something like:

```python
        """
        Get the exact list of snapshots for the given entrypoint resource IDs.

        Use this when a caller requires that every ID in the resource_ids sequence
        returns an EntryPoint object.

        ...
        """
```

src/dioptra/restapi/db/repository/entrypoints.py
jkglasbrenner, line 1: Found several spelling errors and copy-paste leftovers in the new file:

- Lines 215-217 (`get_one_snapshot`): lists `snapshot_id` as an Args entry twice; the first should be `resource_id`.
- Lines 304, 325: `queus` -> `queues`.
- Lines 440, 461: `plugins points as children`.
- Lines 378-379, 396-397, 520-521: Args sections name `experiment:` and `child:` instead of `entrypoint:` and `queue:` / `plugin:`. (These are the `unlink_queue`, `unlink_queues`, and `unlink_plugin` docstrings.)
- Lines 486-498 (`set_plugins`): docstring talks about queues instead of plugins.

tests/unit/restapi/lib/db/fakedata.py
jkglasbrenner, line 1: In the fake data `entry_point`, the `plugin` argument on line 303 can be removed as it's no longer being used.

tests/unit/restapi/test_entrypoint_repository.py
```diff
@@ -0,0 +1,1176 @@
...
+        ep.children.append(plugin.resource)
+        queues.append(queue)
+        plugins.append(plugin)
+
+    return ep, queues, plugins
+
+
+class TestEntrypointCreate:
+    """Tests for EntrypointRepository.create()"""
+
```
jkglasbrenner: Style comment: We haven't used classes in any of our pytest functions throughout the test suite, so it seems stylistically off to start now. If you want to signal  different sections in the file, then something like this works:

```python
# ============================================================================
# region EntrypointRepository.create() tests
# ============================================================================

def test_entrypoint_create_success(
    db_session: DBSession, account, entrypoint_repo
):
    ...  # Rest of tests under the current class

# endregion
```

The above region pattern is recognized by VS Code and lets you collapse and expand these segments of code to make it easier to show and hide code, which is probably the main utility of having classes here.

tests/unit/restapi/test_entrypoint_repository.py
jkglasbrenner, line 1: This test file is noticeably longer than the others. Maybe that's just because the behavior here is more varied than anything that's come before, but I wonder if there's a way we could consolidate some of the testing patterns. Not necessarily blocking, but a cleanup thing we may want to revisit as we work through the rest of the repository pattern.

src/dioptra/restapi/errors.py
```diff
@@ -385,11 +385,21 @@ def __init__(self, type: str, constraint: str, **kwargs):
         self.parameters = kwargs
 
 
-class QueryParameterNotUniqueError(QueryParameterValidationError):
-    """Query Parameters failed unique validation check."""
+class InputParameterNotUniqueError(DioptraError):
+    """Input Parameters failed unique validatation check."""
```
jkglasbrenner: Spelling, should be "validation"

src/dioptra/restapi/db/repository/utils/resources.py
```diff
@@ -543,24 +739,75 @@ def unlink_child(
...
+        parent.children.clear()
+    else:
+        children = [
+            child
+            for child in parent.children
+            if child.resource_type != resource_type.db_table_name
+        ]
+        resource_ids = resource_ids - {child.resource_id for child in children}
+        parent.children.clear()
+        parent.children.extend(children)
```
jkglasbrenner: Why not just this?

```suggestion
        parent.children = children
```

That should replace the list with the new list I think.

src/dioptra/restapi/db/repository/utils/resources.py
```diff
@@ -543,24 +739,75 @@ def unlink_child(
     Raises:
         EntityDoesNotExistError: if parent or child do not exist
     """
-    assert_resource_exists(session, parent, DeletionPolicy.ANY)
-    assert_resource_exists(session, child, DeletionPolicy.ANY)
+    parent = get_one_resource(session, parent, DeletionPolicy.NOT_DELETED)
```
jkglasbrenner: This switched from `DeletionPolicy.ANY` to `DeletionPolicy.NOT_DELETED` compared with the original code, but the child below did not. Is this consistent with the old behavior, or will we start raising a new set of errors because of the change?

jkglasbrenner (changes requested):
@keithmanville Did a review and found a few things. Some of these are blocking, some are more minor, and a couple are more in the comment/question territory. Let me know if you have questions about what I marked.

jkglasbrenner:
Finally, there's a subtle bug we missed in #1240 that might show up with the new logic introduced here.

In `src/dioptra/restapi/v1/entity_types.py`, line 67, there's this:

```python
raise DioptraError("Invalid EntityType: {db_table_name}") from e
```

The `f` prefix is missing from the string, so `{db_table_name}` is emitted as-is without any substitution. From what I can see, line 256 in `src/dioptra/restapi/db/repository/utils/resources.py` could cause this particular path to fire, so you would get a weird error message as a result. We can either fix it here or separately.

src/dioptra/restapi/db/repository/utils/resources.py
keithmanville: ah yeah, one side of that should have been snaps, not resources on both sides. I'll implement as suggested.

src/dioptra/restapi/db/repository/utils/resources.py
keithmanville: I think Andy left a comment elsewhere that the direct list assignment caused an issue. I'll see if I can find that.

tests/unit/restapi/test_entrypoint_repository.py
keithmanville: Yeah, I liked the classes because it made the file easier to navigate. I can switch to this comment style.

src/dioptra/restapi/v1/experiments/service.py
keithmanville: yes, will fix this.

src/dioptra/restapi/v1/experiments/service.py
keithmanville: Correct, I think we can get rid of the commit argument here and in modify as well.

src/dioptra/restapi/db/repository/utils/resources.py
keithmanville: good catch. I will fix this. Do you a preference between those two options?

src/dioptra/restapi/v1/entrypoints/service.py
keithmanville: Might have to discuss this one, as I'm not sure I'm following all of the points you are making.

I think you found a bug that there should be a create_plugins call for artifact_plugins since you should be able to register a plugin as an artifact plugin without it being a task plugin.

the plugin -> plugin_file resource dependency relationship can actually be removed since we now have a tighter connection between the two at the snapshot level. I have a separate commit that is part of my plugin repo refactor that does this.

src/dioptra/restapi/db/repository/utils/resources.py
keithmanville: You're right, this is a behavior change that I don't think we want.

src/dioptra/restapi/v1/experiments/service.py
```diff
@@ -341,16 +245,12 @@ def modify(
...
-        try:
+        with self._uow(commit=True):
             self._uow.experiment_repo.create_snapshot(new_experiment)
-        except Exception:
-            self._uow.rollback()
-            raise
-
-        if commit:
-            self._uow.commit()
+            entrypoints = self._uow.experiment_repo.create_entrypoints(
```
jkglasbrenner: Is `create_entrypoints()` what we want to use here? From what I remember, using `PUT` on the `/experiments/{id}` endpoint should allow you to replace the experiment entrypoint list, meaning `entrypoints=[]` should clear out all existing associations. Using `create_entrypoints()` instead means you can only append entrypoint ids but never remove or clear them.

If we do not have a unit test that validates what happens when you modify an experiment and try to clear the list to `[]` or replace the list of entrypoint ids with something else, we should add one. That should tell us if this is doing what I think it might be doing.

jkglasbrenner (changes requested):

src/dioptra/restapi/v1/experiments/service.py
keithmanville: you are correct. I added two tests to check for this behavior, saw that they failed, then fixed the issue by calling `set_entrypoints` instead.

PR #1221, opened 2026-03-11
fix: unlink queues from parent entrypoints on deletion

keithmanville:
This commit fixes a bug in which the REST API could raise a 500 error when a queue attached to an entrypoint was deleted and then the entrypoint was modified.

The fix is to unlink all entrypoints from a queue when it is deleted.

This commit includes a new test.

### DIFF
--- a/src/dioptra/restapi/db/repository/utils/resources.py
+++ b/src/dioptra/restapi/db/repository/utils/resources.py
@@ -563,6 +563,83 @@ def unlink_child(
             break
 
 
+def unlink_parents(
+    session: CompatibleSession[S],
+    child: m.Resource | m.ResourceSnapshot | int,
+):
+    """
+    "Unlink" the all parents from the given child.  This only severs the
+    relationship; it does not delete either resource.  If there is no
+    parent/child relationship, this is a no-op.
+
+    Args:
+        session: An SQLAlchemy session
+        child: A resource, snapshot, or resource_id integer primary key
+            value
+
+    Raises:
+        EntityDoesNotExistError: if the child do not exist
+    """
+
+    child = get_one_resource(session, child, DeletionPolicy.ANY)
+    child.parents.clear()
+
+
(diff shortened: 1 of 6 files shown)


src/dioptra/restapi/db/repository/utils/resources.py
```diff
@@ -559,6 +559,35 @@ def unlink_child(
             break
 
 
+def unlink_parents(
```
keithmanville, line 553: Following existing pattern for unlinking resources. Would it not be better to:
```
child = get_one(child, DeletionPolicy.ANY) # this has it's own existence checks.
child.parents.clear()
```

jkglasbrenner (commented):
This PR includes the `unlink_child()` refactor that changed the helper’s documented contract where it says that unlinking an absent parent/child relationship is a no-op, but the implementation uses `parent.children.remove(child)`, which raises `ValueError` when both resources exist but are not related. This was pointed out in my review of PR #1199, which fixed this problem. In retrospect, I should have put that comment here rather than #1199, but I wasn't keeping straight where the issue originated, just that the change was leading to an issue we had to address.

If we merge this to dev by itself, then that means we're accepting a bug into dev that could cause issues for anyone working off of dev, which doesn't feel right to do. Given that you've made more progress on #1199, I think we're best off just bundling that all together rather than trying to separate it at this point. You can keep the commits separate in the history for easier tracking, but I feel that all of them should land together in dev at the same time rather than across two separate PRs. That's my suggestion based on where things are at.

The other option, if you feel strongly about keeping these changes separate, is to try and do some surgery on the commit history in #1199 and pull out just the changes that are relevant to what you were intending with this PR in the first place, then do the rebasing needed to make it all consistent again. It's doable, but I'm not sure it's worth the effort if #1199 is close to merging at this point.

keithmanville:
superseded by #1199

PR #1254, opened 2026-04-06
ResourcePicker component

henrychoy:
Closes #1228 

- Any dropdown where you select a resource, will now use the new component `ResourcePicker`
- It should display each resource badge with the correct icon/color
- It should support single or multiselect
- It should display the outdated badge and the sync button when necessary
- Clicking on the badge in the dropdown will be disabled, but you can still right-click or wheel click to open the resource
- It should be read-only when viewing history or deleted

### DIFF
--- a/src/frontend/src/assets/main.css
+++ b/src/frontend/src/assets/main.css
@@ -178,5 +178,5 @@ legend {
 }
 
 .bg-red-light {
-  background-color: rgba(244, 67, 54, 0.02);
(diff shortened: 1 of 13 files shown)


henrychoy:
FYI also converted the remaining badges and dropdowns on the entrypoint and artifact edit pages.  Only dropdown that remains the same is the artifact dropdown in the Job creation page.  That one looked better without the Picker because artifacts dont have names and the long description is displayed instead.

PR #1260, opened 2026-04-15
feat(restapi): validation which loops over swaps

### LINKED ISSUE #821: Create a new validation function for Entrypoint graph YAML with specified swaps

jtsextonMITRE:
Closes #821


src/dioptra/sdk/api/swappable_validation.py
```diff
@@ -29,3 +31,15 @@ def get_json_schema(default: bool = False) -> dict:
...
+
+
+def get_swappable_experiment_schema() -> dict:
+    base_schema = dioptra.task_engine.validation.get_json_schema()
+    swappable_schema = get_json_schema()
+
+    if "$defs" in base_schema and "$defs" in swappable_schema:
+        # since the definitions have the same keys, we can just replace the definitions
+        # in the original schema with the ones from the swappable schema.
+        base_schema["$defs"].update(swappable_schema["$defs"])
```
jtsextonMITRE, line 43: Show this to James and Harold

src/dioptra/sdk/api/swap_graph.json
```diff
@@ -0,0 +1,160 @@
...
+                },
+                {
+                    "$ref": "#/$defs/swap_mixed_invocation"
+                }
+            ]
+        },
+        "positional_invocation": {
+            "$ref": "experiment_schema.json#/$defs/positional_invocation"
+        },
+        "keyword_invocation": {
```
jtsextonMITRE, line 64: The positional and keyword invocation styles needed a change to be compatible with swaps.

A step declaration with the format:

```
step1:
   ?xyz:
       ambiguous: 
          thing: [value]
```

can be either:
- a swap called `xyz` where the first option is a choice called `ambiguous` which uses a task called `thing`
- a task using the keyword invocation style called `?xyz` where the first argument is called `ambiguous` that takes an object.

To prevent this, we disallow `?` to be used at the start of keyword invocation.

Additionally the swap rendering function assumes any top level entry under the step name starting with `?` is a swap, so disallowing it here and for positional invocations will allow that to be caught before it causes render issues. 


As for whether `?` should be disallowed in `experiment_schema.json`:
Pros:
- allows making the `swap_graph.json` slightly simpler by just referencing `experiment_schema.json`'s definitions
- assuming we implement this as a blanket restriction on starting with `?` in task names, it may reduce user confusion about where they can and can't use `?` in the task names.

Cons:
- taken on its own, the experiment_schema.json has a seemingly nonsensical restriction on task names given that it has no awareness of swaps

jtsextonMITRE: Resolved instead by restricting task names to [a-zA-Z0-9_] to mimic python functions

src/dioptra/task_engine/experiment_schema.json
```diff
@@ -328,18 +328,21 @@
...
-                "$comment": "The task plugin short name + arg value(s)",
-                "type": [
-                    "string",
-                    "number",
-                    "boolean",
-                    "null",
-                    "array"
-                ]
+            "patternProperties": {
+                    "^(?!dependencies$)[a-zA-Z0-9_]+$": {
```
jtsextonMITRE, line 332: `^[a-zA-Z0-9_]+$` will match `dependencies` as well, so we have to additionally specify that a task cannot be named `dependencies`.

src/dioptra/task_engine/error_message.py
```diff
@@ -170,17 +178,30 @@ def _extract_schema_by_schema_path(
             # lookup in a list.  As a static type checker, mypy seems to want
             # one meaning, and I couldn't figure out how to make that pass mypy
             # checks.
-            subschema = schema[next_path_component]  # type: ignore
+            if isinstance(next_path_component, str) and next_path_component.endswith(
+                "#"
+            ):
+                filename = next_path_component[:-1]  # remove the '#'
+                # Attempt to resolve via provided registry first
+                print("Registry", list(registry), flush=True)
```
keithmanville: remove debug print.

examples/graph.txt
```diff
@@ -0,0 +1,22 @@
+greeting_step:
```
keithmanville: remember to remove before merging

examples/test_validation.ipynb
```diff
@@ -0,0 +1,137 @@
+{
```
keithmanville: remember to remove before merging

src/dioptra/client/entrypoints.py
```diff
@@ -694,6 +695,14 @@ def snapshots(self) -> EntrypointsSnapshotCollectionClient[T]:
                 entrypoint_snapshot_id=2,
                 file_type=FileTypes.TAR_GZ,
             )
+
+            # POST /api/v1/entrypoints/1/snapshots/2/validateSwaps
+            client.entrypoints.snapshots.validate_swaps(
```
keithmanville: is this just out of date documentation? Should be client.entrypoints.validate now.

src/dioptra/restapi/v1/entrypoints/controller.py
```diff
@@ -755,6 +758,51 @@ def get(self, id: int, snapshotId: int):
...
+        **kwargs,
+    ) -> None:
+        """Initialize the validate swaps resource.
+
+        All arguments are provided via dependency injection.
+
+        Args:
+            entrypoint_swaps_validation_service: An EntrypointSwapsValidationService object.
+        """
+        self._swaps_validation_service = swaps_validation_service
```
keithmanville: this should also be checked in `modify`

src/dioptra/restapi/v1/workflows/service.py
keithmanville: add more descriptive docstring that lists exactly what is being validated. e.g. swap outputs, task engine yaml schema, swaps yaml schema, etc.

src/dioptra/client/entrypoints.py
```diff
@@ -952,3 +961,46 @@ def delete_by_id(self, entrypoint_id: str | int) -> T:
             The response from the Dioptra API.
         """
         return self._session.delete(self.url, str(entrypoint_id))
+
+    def validate(
+        self,
+        group_id: int,
+        swaps_graph: str,
```
keithmanville: this will validate any entrypoint, correct? swap or no. I think from the user perspective we should make this clearer by naming the variables as more generic "graph" (which could have swaps or not) so it is clearer that this can always be used to validate.

src/dioptra/restapi/v1/entrypoints/controller.py
```diff
@@ -755,6 +758,51 @@ def get(self, id: int, snapshotId: int):
         )
 
 
+@api.route("/validate")
```
keithmanville: now that we have the heavier weight validation under entrypoints, should the lightweight validation under workflow move here too? Or is there a good reason for keeping these in two very different parts of the api?

I'd like it to be clear to the user what the distinction is and when each should be used. I think some of that comes from where in the API it is and what the endpoints are names, and the rest from good docstrings and documentation.

src/dioptra/restapi/v1/entrypoints/schema.py
```diff
@@ -390,3 +391,114 @@ class DynamicGlobalParametersResponseSchema(Schema):
...
     )
+
+
+class ValidateSwapsRequestSchema(Schema):
+    """The schema for validating swaps on an entrypoint."""
+
+    groupId = fields.Integer(
+        attribute="group_id",
+        data_key="groupId",
+        metadata={"description": "ID of the group for the entrypoint."},
```
keithmanville: groupId needs to be `required=True` because it it used to query plugin param types by group, which doesn't handle a null groupId

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -1683,6 +1689,381 @@ def get_params(
...
+            entrypoint_artifacts: A list of entrypoint artifacts parameters.
+            plugin_snapshot_ids: A list of plugin snapshot IDs needed for the entrypoint.
+
+        Returns:
+            A ValidateSwapsResponseSchema object.
+        """
+
+        log: BoundLogger = kwargs.get("log", LOGGER.new())
+
+        print(swaps_graph, flush=True)
```
keithmanville: remove debug prints

keithmanville (changes requested):

src/dioptra/restapi/v1/entrypoints/controller.py
jtsextonMITRE: I added some documentation which should explain the difference better (and when to use each). 

As for which belongs where, I think the heavier does belong under entrypoints since its main usage is intended to be saving before create/modify, though I'm unsure about the lighter one. I think having it in workflows could make sense from the perspective of it being intended for code-editing, and not being directly related to an entrypoint operation. It does use mostly the same inputs as the heavier one, which is an argument for having it also be in entrypoints, though.

PR #1271, opened 2026-04-29
Import tasks dialog styling

henrychoy:
This fixes some styling issues on the import plugin tasks dialog in the edit plugin file page.

- When a row is expanded with the warning that existing params will be overwritten, the chips should line up vertically
- When a row is expanded with any warning, there should be a top and bottom border to help separate it from other rows
- When a row is expanded, clicking on the expanded part should select/deselect, and it should look like one big row.


keithmanville (approved):
LGTM

PR #1301, opened 2026-06-04
fix (db): update db-migration to create swaps i823

JustKuzya:
- add guide for testing the database upgrade with containerized Postgres DBMS
- create migration script for swaps
- verify that the script works for SQLite and containerized Postgres


keithmanville (commented):
I was expecting to see the JobSwaps ORM object in `db/models/jobs.py` and changes to the response schema in `v1/jobs/schema.py`. Did you forget to commit? It looks like you had this done if you generated the alembic script from the ORM.

src/dioptra/restapi/db/models/jobs.py
```diff
@@ -302,3 +303,34 @@ def __declare_last__(cls) -> None:
...
+    )  # FK to jobs
+    swap_name: Mapped[text_] = mapped_column(primary_key=True)
+    task_alias: Mapped[text_]
+    plugin_file_resource_snapshot_id: Mapped["PluginFile"] = mapped_column(
+        ForeignKey("plugin_files.resource_snapshot_id"),
+        init=False,
+    )  # FK to plugin_files
+
+    # PK constraints settings
+    _table_args__ = (
```
keithmanville: typo: should be `__table_args__`

src/dioptra/restapi/db/models/jobs.py
```diff
@@ -302,3 +303,34 @@ def __declare_last__(cls) -> None:
...
+
+class JobSwap(db.Model):  # type: ignore[name-defined]
+    """
+    Maps the jobs-swaps-plugin_files
+    """
+
+    __tablename__ = "job_swaps"
+
+    # Table fields
+    job_id: Mapped["Job"] = mapped_column(
```
keithmanville: Let's use job_resource_id instead of job_id to match the other ORM models like EntrypointJob and ExperimentJob.

The type should be an integer since it is an ID not the job ORM class.

src/dioptra/restapi/db/models/jobs.py
```diff
@@ -302,3 +303,34 @@ def __declare_last__(cls) -> None:
...
+    task_alias: Mapped[text_]
+    plugin_file_resource_snapshot_id: Mapped["PluginFile"] = mapped_column(
+        ForeignKey("plugin_files.resource_snapshot_id"),
+        init=False,
+    )  # FK to plugin_files
+
+    # PK constraints settings
+    _table_args__ = (
+        PrimaryKeyConstraint("job_id", "swap_name"),
+        ForeignKeyConstraint(
```
keithmanville: job_id should be with jobs.resource_id and plugin_file_resource_snapshot_id with plugin_files.resource_snapshot_id.

But in the end I don't think we need the FK Constraints here, as long as they are defined in the mapped_columns

src/dioptra/restapi/db/models/jobs.py
```diff
@@ -302,3 +303,34 @@ def __declare_last__(cls) -> None:
...
+    job_id: Mapped["Job"] = mapped_column(
+        ForeignKey("jobs.resource_id"), init=False, primary_key=True
+    )  # FK to jobs
+    swap_name: Mapped[text_] = mapped_column(primary_key=True)
+    task_alias: Mapped[text_]
+    plugin_file_resource_snapshot_id: Mapped["PluginFile"] = mapped_column(
+        ForeignKey("plugin_files.resource_snapshot_id"),
+        init=False,
+    )  # FK to plugin_files
+
```
keithmanville: need to add relationship with Jobs. Also requires the entry in Jobs ORM object:

```
job_swaps: Mapped[list["JobSwap"]] = relationship(
    init=False,
    back_populates="job",
)
```

src/dioptra/restapi/db/alembic/versions/30e74c5564b3_add_jobswaps_orm.py
```diff
@@ -0,0 +1,47 @@
...
+Revision ID: 30e74c5564b3
+Revises: ad4f89b2288d
+Create Date: 2026-06-16 10:17:21.101904
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = '30e74c5564b3'
```
keithmanville: seeing single quotes. need to run ruff on this file.

keithmanville (changes requested):

keithmanville:
Left a review, but since changes were all minor, I went ahead and made and committed them.

Also change PR target to `entrypoint-swaps` branch.

src/dioptra/restapi/db/alembic/versions/30e74c5564b3_add_jobswaps_orm.py
JustKuzya: I see in the latest version all of the changes you have suggested seem to be applied.

PR #1307, opened 2026-06-11
Frontend lint error

keithmanville:


keithmanville:
closing: verified that frontend checks fail when linting error is introduced.

PR #1309, opened 2026-06-15
Allow punctuation in unquoted search queries

### LINKED ISSUE #1209: [BUG] Including a period (".") in search bar for Job logs throws error
**Describe the bug**
If you include a period (".") in the search bar for job logs, you get a "failed to fetch logs" notification. 

**To Reproduce**
1. Run a job with logs 
2. Open jobs dashboard
3. Type in any search string with a period 

**Expected behavior**
Punctuation shouldn't cause the search to fail.

arpitjain099:
Fixes #1209.

**Problem:** the query grammar restricted unquoted search terms to `pp.Word(pp.alphanums + "_")`, so any unquoted query containing other punctuation failed `parse_all`. The most common case is a period: searching job logs for something like `file.txt` raised `SearchParseError`, which the UI surfaced as "failed to fetch logs". Quoting the term worked, but users have no reason to expect that.

**Fix:** unquoted searches now accept any printable character that is *not* grammar-significant. The grammar-significant set is just the escape character, the wildcards `*`/`?`, the quote characters, the field separator `:` and the list delimiter `,`; those still have to be quoted or escaped. This mirrors what quoted searches already allow.

**Verified** against the real grammar:
- now parse: `file.txt`, `v1.2.3`, `name:dioptra.restapi`, `some/path-to_file.log`, `50%off`
- still parse unchanged: `search_*`, `tag:classification`, `description:*mnist*,tag:cv`, `name:trial_??`
- still rejected: `field:search all for this`, `bad-name:classification`, `description:bad_delim|tag:cv`, an unclosed quote, `\q`

Added `test_parse_search_text_allows_punctuation` covering the period/slash/dash cases.

One behavior change worth calling out: `bad=assignment` previously errored and now parses as a literal search for the text `bad=assignment` (since `=` has no role in the grammar). I removed it from the "invalid" examples in `test_grammar` to match. If you would rather keep `=` as an explicit error, I can add it back to the excluded set; happy to adjust.


keithmanville:
Thank you for your contribution. I will review and let you know if we have any questions.

In the meantime, can you please follow the [instructions here](https://github.com/usnistgov/dioptra/blob/main/CONTRIBUTING.md#criteria-for-contributions-and-feedback) to add your github username to CONTRIBUTORS.md, thereby accepting the license terms? Thank you!

arpitjain099:
Thanks @keithmanville! Added my username to CONTRIBUTORS.md (accepting the CC BY 4.0 terms). Ready for review whenever you get to it.

keithmanville:
replace with PR #1329

This branch was based on main instead of dev. Sqaushed commits, cherry-picked onto dev, and opened new PR.

PR #1312, opened 2026-06-17
feat(restapi): add swap retrieval endpoint

jtsextonMITRE:
Closes #1215

### DIFF
--- a/src/dioptra/restapi/v1/entrypoints/service.py
+++ b/src/dioptra/restapi/v1/entrypoints/service.py
@@ -1906,9 +1906,9 @@ class SwapsRetrievalService(object):
                             task_name = list(task_def.keys())[0]
 
                         if task_name in task_lookup_dict:
-                            plugin_file_resource_snapshot_id = task_lookup_dict[task_name][
-                                "plugin_file_snapshot_id"
-                            ]
+                            plugin_file_resource_snapshot_id = task_lookup_dict[
+                                task_name
+                            ]["plugin_file_snapshot_id"]
                             swap_info = {
                                 "swap_name": swap_name,
                                 "task_alias": alias,
--- a/src/dioptra/restapi/v1/entrypoints/service.py
+++ b/src/dioptra/restapi/v1/entrypoints/service.py
@@ -1839,7 +1839,7 @@ class SwapsRetrievalService(object):
             entrypoint_snapshot_id: The entrypoint snapshot ID.
         Returns:
             An object of the form:
-            {"swaps": 
+            {"swaps":
                 "swap_name": [{
                     "task_name": ...,
                     "entrypoint_keyword_args": [...],
@@ -1848,7 +1848,7 @@ class SwapsRetrievalService(object):
             }
         """
         log = logger or LOGGER.new()
-        
+
         # Retrieve the snapshot entrypoint to get the graph.
         entry_point = self._entrypoint_snapshot_id_service.get(
             entrypoint_id=entrypoint_id,
@@ -1860,7 +1860,7 @@ class SwapsRetrievalService(object):
         try:
             graph = yaml.safe_load(entry_point.task_graph) or {}
         except Exception as e:
-            raise InvalidYamlError(f"Failed to load YAML: {e}")
+            raise InvalidYamlError("Failed to load YAML") from e
 
         # this looks like {'swap_name1' : ['alias1', 'alias2'], 'swap_name2': ...}
         swaps = self._swaps_validation_service.extract_swaps(graph)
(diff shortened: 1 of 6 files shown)


tests/unit/restapi/v1/test_entrypoint.py
```diff
@@ -1787,6 +1802,8 @@ def _test_dynamic_globals_endpoint(
         )
 
         assert set(expected_globals) == set(evaluated["entrypointParams"])
+
+        print(evaluated, flush=True)
```
keithmanville: remove leftover debug print.

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -1797,6 +1806,136 @@ def raise_validation_errors(
...
+                        task_def = task_defs[alias]
+
+                        keyword_args = (
+                            self._dynamic_global_parameters_service.get_keywords_for_one_task(
+                                task_def
+                            )
+                            - step_names
+                        )
+
+                        if "task" in task_def:
```
keithmanville, line 1901: is this checking the different ways of listing tasks with args (positional / keyword / mixed)?

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -1797,6 +1806,136 @@ def raise_validation_errors(
...
+        ]
+
+        task_lookup_dict = self._swaps_validation_service.build_task_lookup_dict(
+            plugin_plugin_files
+        )
+        not_found_swaps = set()
+        not_found_tasks = set()
+
+        for swap_name in swaps:
+            swap_found = False
```
keithmanville: is it possible to have `swap_found = False` for a validated graph?

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -1797,6 +1806,136 @@ def raise_validation_errors(
...
+                            )
+                            - step_names
+                        )
+
+                        if "task" in task_def:
+                            task_name = task_def["task"]
+                        else:
+                            task_name = list(task_def.keys())[0]
+
+                        if task_name in task_lookup_dict:
```
keithmanville, line 1908: is this possible for a validated task graph

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -1797,6 +1806,136 @@ def raise_validation_errors(
...
+
+        task_lookup_dict = self._swaps_validation_service.build_task_lookup_dict(
+            plugin_plugin_files
+        )
+        not_found_swaps = set()
+        not_found_tasks = set()
+
+        for swap_name in swaps:
+            swap_found = False
+            for step in step_names:
```
keithmanville, line 1888: I think the logic of this loop is a bit confusing because we are doing a mix of traversing through the graph dict along with lookups from what we extracted via other services. Is it possible to simply trust the graph instead of checking against swaps from the extract swaps service?

tests/unit/restapi/v1/test_entrypoint.py
jtsextonMITRE: I think this one is resolved

src/dioptra/restapi/v1/entrypoints/service.py
jtsextonMITRE: Yes, if the key "task" is present it means it's a "long form" definition i.e. keyword, and otherwise the key represents a task name

src/dioptra/restapi/v1/entrypoints/service.py
jtsextonMITRE: This was more of a sanity check but as the swaps dictionary comes from SwapsValidationService and that one directly produces the list of swap_names based on the graph I think it should actually be impossible for swaps_found to be False ever.. removing

src/dioptra/restapi/v1/entrypoints/service.py
jtsextonMITRE: This one I'd be more hesitant about removing, as I was thinking of a scenario where the entrypoint is saved, and then someone made changes to plugins to remove a task? I'm not sure that has to go through validation. Alternatively, another scenario is where someone has upgraded from Dioptra 1.x to 2.x and their old entrypoints have not necessarily been validated and things starts breaking when trying to display them.

On that note, should this be using snapshots somehow? Though I don't see the snapshots in the ORM for that class unless they happen automatically
```
        plugin_plugin_files = [
            plugin_plugin_file
            for entry_point_plugin in entry_point.entry_point_plugins
            for plugin_plugin_file in entry_point_plugin.plugin.plugin_plugin_files
        ]
```

src/dioptra/restapi/v1/entrypoints/service.py
jtsextonMITRE: agreed

PR #1327, opened 2026-06-30
feat(restapi): add swaps to job create

### LINKED ISSUE #824: Update Job create REST API schema to specify slots
The following field should be added to the Job Create Schema:
```
"swaps": [
  { "swap_name": str, "task_alias": str},
  ...
]
```
It should be optional with a default of an empty list.

Update the controller to parse this field and pass it to the service layer.
The service layer should be updated to accept this field. For now if anything other than an empty list is provided, a `DioptraError` error should be raised indicating the feature is not yet implemented.

The client needs to be updated to pass the additional swaps information.

## Blocked by:
- #823
- #755

jtsextonMITRE:
closes #824


src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -303,6 +313,61 @@ def create(
             job_resource=job_resource,
             queue=queue,
         )
+
+        job_swaps = []
```
keithmanville, line 345: the core jobs swaps logic is a fairly lengthy addition to an already complexy create method. Can this be broken out to a helper function that returns a list of job_swaps and raises errors if there are any.

from this method it would look something like `new_job.job_swaps = self._build_job_swaps(...)`

keithmanville (changes requested):

PR #1329, opened 2026-07-07
Allow punctuation in unquoted search queries

### LINKED ISSUE #1209: [BUG] Including a period (".") in search bar for Job logs throws error

keithmanville:
The search grammar limited unquoted search terms to alphanumerics and underscores, so any unquoted query containing other punctuation (most commonly a period) failed to parse and the job-logs search surfaced a "failed to fetch logs" error.

Unquoted searches now accept any printable character that is not grammar-significant (the escape character, the wildcards, the quote characters, the field separator and the list delimiter); those still need to be quoted or escaped. Quoted searches already allowed these characters, so this just brings unquoted searches in line.

Fixes #1209

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -19,3 +19,4 @@ cminiter
 pscemama-mitre
 alexb1200
 jsoref
+arpitjain099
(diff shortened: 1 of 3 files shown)


keithmanville:
@arpitjain099 please see this new PR now based on dev.

PR #1334, opened 2026-07-29
feat(restapi): partial swap rendering 

### LINKED ISSUE #1286: Add new endpoint to create partially rendered swap graphs
add a new endpoint that can partially render a swappable task graph. If there are no swaps, it just returns the task_graph. This will be used to show the user what the graph looks like throughout job creation as they choose swaps.

`POST /entrypoints/{id}/snapshots/{snapshotId}/task_graph`

body:
```
{ "swap_name": "task_alias", ...}
```

This may require updating the render function.

jtsextonMITRE:
Closes #1286

### DIFF
--- a/src/dioptra/restapi/v1/entrypoints/schema.py
+++ b/src/dioptra/restapi/v1/entrypoints/schema.py
@@ -440,6 +440,16 @@ class EntrypointConfigRequestSchema(SwapChoiceRequestSchema):
         },
     )  #  type: ignore
 
+    @validates("sections")
+    def validate_sections(self, sections: list[str]) -> None:
+        invalid_sections = (
+            set(sections) - EntrypointConfigResponseSchema().fields.keys()
+        )
+        if invalid_sections:
+            raise ValidationError(
+                f"Invalid config sections: {sorted(invalid_sections)}."
+            )
+
 
 class DynamicGlobalParametersResponseSchema(Schema):
     globalParameters = fields.List(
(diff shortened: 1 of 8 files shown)


src/dioptra/restapi/v1/entrypoints/schema.py
```diff
@@ -392,6 +392,27 @@ def _deserialize(self, value, attr, data, **kwargs) -> dict[str, str]:
...
+        **additional_metadata,
+    ) -> None:
+        super().__init__(**additional_metadata)
+        self.delimiter = delimiter
+
+    def _deserialize(self, value, attr, data, **kwargs) -> list[str]:
+        try:
+            if value == "":
+                return []
+            return [s.strip() for s in value.split() if s.strip()]
```
keithmanville: value.split should take self.delimiter.

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -1288,13 +1292,13 @@ def get_config(
...
             logger=log,
+            sections=sections,
         )
 
-        # we should be able to run this in all cases because render_swaps_graph will raise
-        # errors if the graph has unspecified swaps or if swaps are specified that aren't
-        # needed
         try:
-            config["graph"] = render_swaps_graph(config["graph"], swap_choices)
+            config["graph"] = render_swaps_graph(
```
keithmanville: graph is technically an optional section, right? I don't know that there is an actual use case for where we don't supply graph, but I get_config would always raise an error on this index even when sections does not contain graph.

PR #1336, opened 2026-08-04
feat: add deployment manager CLI

### LINKED ISSUE #1272: CLI installer for Dioptra deployments

etrapnell-nist:
Adds dioptra-platform CLI commands for managing containerized Dioptra deployments: `install`, `uninstall`, `update`, `upgrade`, `start`, `stop`, `status`, `list`, and `clean`.

Container image tags move from `docker-compose.yml` to a `CONTAINER_TAG` variable in `.env`, separating container builds from the Python package version and allowing for independent container updates.

`install`/`uninstall`/`status`/`list`/`clean` operate on a registry + manifest resource tracking system. Both default and custom install locations are supported. Deployment names are only required when multiple deployments are present; otherwise, if not specified, a default name is assumed for all operations.

`update`/`upgrade` check `latest_build.json` on GitHub (included in this PR) and compare against the deployment's current tag. upgrade edits `.env`, pulls, and recreates containers while preserving volumes and run state; failures roll back `.env`. `install --force` wipes volumes for a fresh start; whereas `upgrade` preserves them. Versioning assumes the following structure: `major.minor.patch[devN][-buildN]`. `-buildN` is an optional container-only build suffix.

Includes unit tests for version comparison, install selection, .env handling, name/slug resolution, uninstall planning, transient-error detection, and upgrade rollback.

Command overview:
  `clean`:      Clean up Dioptra resources that are no longer in use.
  `install`:    Install a new Dioptra deployment.
  `list`:       List all registered Dioptra deployments.
  `start`:      Start a deployment.
  `status`:     Show deployment status.
  `stop`:       Stop a deployment.
  `uninstall`:  Remove a Dioptra deployment.
  `update`:     Check for available updates without applying anything.
  `upgrade`:    Apply available container updates to a deployment.

Closes #1272


jkglasbrenner (commented):
@etrapnell-nist Thanks for the work you've put into this. The behavior I saw in your demo last week matches the user experience we want to provide, so my review is now focused on deployment state tracking and lifecycle management within the implementation. Seeing the current implementation has helped clarify some deployment lifecycle requirements that I realize I did not say clearly (what is here is reasonable given what I verbally described). So, before we make further changes, I want to pause work on this branch until I write a more detailed plan/outline to make sure there is a shared understanding of what the architecture of this should be.

In terms of the direction I think this will need to go:

- The installed `dioptra-platform` package owns the deployment template. Each versioned release of `dioptra-platform` owns and provides a specific template snapshot. The template is allowed to evolve over time, but all changes are managed as updates to the `dioptra-platform` package. This is different from the status quo, where deployment template updates are managed independently of versioned releases.
- The user's deployments should be managed by the current `dioptra-platform` installation and its dependencies. If we do that, we should not need to create separate installer or per-deployment virtual environments, and we can switch to using Cookiecutter programmatically for template rendering instead of invoking it using the `subprocess` module.
- Updating a deployment should regenerate the infrastructure-as-code deployment files from the template bundled with the installed package. This will ensure that users have the latest intended files available for the version of Dioptra they want to use. This would replace the current image-tag-only update mechanism. Regenerating the template also means that we do not need to merge generated files in place, as Cruft does. Container versions would return to being embedded in the rendered deployment files, as they were in the existing deployment-template workflow, and would be controlled through Cookiecutter variables rather than stored as user configuration in `.env`.
- Updates must preserve explicitly identified persistent state, including data volumes, credentials, certificates, and user overrides. Generated files and rebuildable configuration volumes should be recreated from the updated template.
- Remote build metadata (the `latest_build.json` file) may still be needed to advertise compatible container rebuilds that do not require a new package release, but its role and publication process need to be fully defined so that the boundary is clear between information bundled with `dioptra-platform` and information retrieved remotely.

As I mentioned, I'll take on the responsibility of getting the plan/outline together for how this will work and what needs to be updated. One of my goals is to minimize changes to the CLI interface as much as possible.

PR #1349, opened 2026-08-12
fix: catch invalid swaps graph yaml errors

### LINKED ISSUE #1347: [BUG] invalid yaml can cause 500 error
invalid yaml like the following (num2 has no `:`) causes a 500 error. This exception should be caught and raised as a proper Dioptra error.

```yaml
aaa:
  task: multiply_numbers
  kwargs:
    num1: $num1
    num2
```
The likely fix is to add a try except w. re-reraise around entrypoints/service.py", line 1720, in validate
```python
    swaps_yaml = yaml.safe_load(swaps_graph)
```

keithmanville:


jtsextonMITRE (approved):
LGTM

PR #1350, opened 2026-08-12
fix: validate task names in entrypoint graph yaml

keithmanville:
This commit checks task references in both in both regular steps and swap options during entrypoint graph validation.

It separates the task reference validation from swap output validation to keep each check focused.


jtsextonMITRE (approved):
LGTM

Issue #1341, opened 2026-08-11
add an explainer for swappable task graphs


keithmanville:
The explainer should put the entrypoint swaps feature into context and explains why you would want this feature. It should answer the question "what problem is entrypoint swaps solving"

PR #1351, opened 2026-08-12
docs: swaps explainers

### LINKED ISSUE #1340: update reference documentation to encompass swappable task

jtsextonMITRE:
Closes #1340 and #1341


docs/source/explanation/components/entrypoints-explanation.rst
```diff
@@ -94,6 +94,18 @@ Executing an entrypoint as a job requires a worker.
 By linking queues to entrypoints, workers can be assigned jobs to process the associated task graph logic.
 
 
+Entrypoint Swaps
```
keithmanville: can we make this a subsection under task graph above? I think it fits better there logically. Then we don't need a second learn more.

docs/source/reference/task-engine-reference.rst
```diff
@@ -355,6 +355,9 @@ For example:
             inputs: "<input_definitions>"
             outputs: "<output_definitions>"
 
+
```
keithmanville: remove added whitespace

docs/source/reference/task-engine-reference.rst
```diff
@@ -589,6 +592,52 @@ positional invocation.  The list has one value in it, which is itself a
 mapping: {prop1: value1, prop2: value2}.  The task plugin will be invoked
 positionally, where its one positional argument will be a Python dict.
 
+Swaps
```
keithmanville: can we move this to the dioptra components / entrypoint reference page? The task engine reference is much lower level and has details the typical user doesn't need to know about (all of the sections of the task engine yaml with the exception of the graph are generated automatically). Also, the task engine itself isn't aware of swaps as it runs on the rendered graph only.

keithmanville: it can be a subsection under "Task Graph Syntax" on the entrypoint ref page.

docs/source/reference/task-engine-reference.rst
```diff
@@ -589,6 +592,52 @@ positional invocation.  The list has one value in it, which is itself a
...
+                args: [posarg1, posarg2]
+                kwargs:
+                    keyword1: arg1
+                    keyword2: arg2
+            task_alias2:
+                plugin2: [arg1, arg2]
+
+A swap is specified in the task graph as a child of the step. In order to be considered a swap,
+the swap name in the task graph must start with the character ``?``. The ``?`` is excluded from the
+swap name when providing selections during job creation.
```
keithmanville: include swap rules:
- one swap per graph step
- swap names must be unique across the graph
- each swap must receive exactly one selection at job execution time

keithmanville (changes requested):
update job reference and explainers to include specifying swaps.

PR #1353, opened 2026-08-13
fix: reuse task plugin as artifact plugin

### LINKED ISSUE #1348: [BUG] Entrypoint validation fails with a 404 if the same plugin exists as both a function and artifact plugin
To reproduce, add the same plugin to both function and task plugins on a single entrypoint. then attempt to validate.

```
Not Found - Failed to locate Resource with resource_id having value (None) . You have requested this URI [/api/v1/entrypoints/62] but did you mean /api/v1/entrypoints/<int:id>/tags/<int:tagId> or /api/v1/entrypoints/<int:id>/tags or /api/v1/entrypoints/<int:id>/snapshots/<int:snapshotId>/swaps ?
```

jtsextonMITRE:
Closes #1348


keithmanville (approved):
LGTM

PR #1354, opened 2026-08-17
Job page swaps

### LINKED ISSUE #828: Update Jobs create page to allow the user to specify task swaps

henrychoy:
Closes #828, #829 

- Update create job page to allow user to select swaps
- Show partially rendered graph when the user selects/changes entrypoint, and as they select swaps
- Show swaps on job dashboard page

### DIFF
--- a/src/dioptra/restapi/v1/jobs/service.py
+++ b/src/dioptra/restapi/v1/jobs/service.py
@@ -334,6 +334,9 @@ class JobService(object):
             job=new_job,
             artifacts=[],
             has_draft=False,
+            swap_task_names=_build_swap_task_name_lookup(
+                new_job, self._swaps_retrieval_service, log
+            ),
         )
 
     def _build_job_swaps(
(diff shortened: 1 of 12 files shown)


src/dioptra/restapi/v1/jobs/service.py
```diff
@@ -1672,12 +1692,44 @@ def get_logs(
         return records, total_count
 
 
-def _build_job_dict(jobs: list[models.Job]) -> list[utils.JobDict]:
+def _build_swap_task_name_lookup(
```
jtsextonMITRE, line 1695: Should we maybe merge this with `build_task_lookup_dict` in `SwapsValidationService`? Maybe in a shared util somewhere? Seems like they do mostly the same thing.

jtsextonMITRE (changes requested):

src/dioptra/restapi/v1/jobs/service.py
henrychoy: My understanding is that `build_task_lookup_dict` gives us the task_names, but not swapName or taskAlias, so we cant grab the specific task_name for a particular swap that we need.  

I tried adding job and swaps_retrieval_service to `build_task_lookup_dict`, but swaps_retrieval_service.getSwaps also calls `build_task_lookup_dict` and it would recurse forever.  So having trouble combining these two.

jtsextonMITRE: Ah I see, that makes sense, I think I missed that this was doing more than just the name. I think other than that this looked good so I'll just mark as resolved

jtsextonMITRE (approved):

PR #1363, opened 2026-09-01
refactor: refactor entrypoint swaps

### LINKED ISSUE #1325: refactor swaps-related services in entrypoints to follow repository pattern
The entrypoints service layer was re-written to use a repository for data access. However the new swaps functionality does not follow this new pattern. It introduces new services that are injected into other services, which should instead be integrated into the entrypoint repository instead.

Refactor the new swaps related services to follow the established repository data access pattern. This should not change any entrypoints functionality. Tests should remain unchanged.

keithmanville:
closes #1325


src/dioptra/restapi/v1/entrypoints/task_engine_yaml.py
```diff
@@ -0,0 +1,290 @@
...
+        raise EntrypointParameterTypeMismatchError(
+            values=param_values,
+            parameter_names=param_names,
+            correct_types=correct_types,
+        )
+
+    return params
+
+
+_EXPLICIT_GLOBAL_TYPES: Final[set[str]] = {
```
jtsextonMITRE: I don't think this is a regression as this was how it was before, but I wanted to double check that `list` and `mapping` don't also belong in here?

jtsextonMITRE (commented):
The MNIST notebook with swaps still works and the UI looks good too.

src/dioptra/restapi/v1/entrypoints/task_engine_yaml.py
jtsextonMITRE: Alternatively maybe `_convert_parameter_type_to_task_engine_type` doesn't need to have `list` and `mapping` in the conversion map

keithmanville: only the "basic types" should be in the _EXPLICIT_GLOBAL_TYPES. including them in `_convert_...` is unreachable and is a mistake. Committed a fix for this.

PR #1372, opened 2026-09-14
rerun job with swaps bug

### LINKED ISSUE #1368: [BUG] Re-running a job does not restore swap choices
The rerun flow loads the original job, but never initializes selectedSwaps from oldJob.swaps. The entrypoint watcher restores only selections already present in the current form, which is initially empty.

Re-running a swappable job therefore opens with unresolved choices and hidden parameters. Submitting without manually reconstructing the original choices fails, rather than reproducing the original job.

Suggested fix: Initialize selections from the original job’s swapName/taskAlias pairs, validating them against the effective entrypoint snapshot.

Locations: CreateJob.vue:744–763 (src/frontend/src/views/CreateJob.vue#L744), CreateJob.vue:1158–1185 (src/frontend/src/views/CreateJob.vue#L1158)

henrychoy:
Closes #1368 

- When re-runing a job, swap selections are preserved if the original job had them
- When hitting "Update Values" the swap choices will be preserved as long as the latest entrypoint had the same swap options.  
- When hitting "Revert Values", the original params and swap options for the original job should be restored.
- Essentially, this should work as expected whether you update param values, or change the whole task graph, and then re-run an old job.


jtsextonMITRE (approved):
Works on containers for me

PR #1373, opened 2026-09-14
fix: only require parameters for selected swaps

### LINKED ISSUE #1367: [BUG] Job creation fails when unchosen swaps have params without defaults
The UI only displays and submits parameters returned by dynamicGlobalParameters for the selected swaps. However, the backend still constructs values for every declared entrypoint parameter and rejects any without a supplied value or default.

For example, suppose one swap choice requires epsilon, another requires iterations, and neither parameter has a default. Selecting the first choice hides iterations, but submission fails with JobParameterMissingError for it. The user cannot supply the missing value through the form.

Suggested fix: Make parameter requirements consistent across the selected graph, job creation, and worker configuration. Add an integration test with mutually exclusive swap parameters that have no defaults.

Locations: CreateJob.vue:702–733 (src/frontend/src/views/CreateJob.vue#L702), jobs/service.py:267–284 (src/dioptra/restapi/v1/jobs/service.py#L267)

jtsextonMITRE:
Closes #1367


henrychoy (approved):
Works for me.  I can now run jobs with swaps with params that have no default, as long as I choose a swap that doesn't reference the param with no default.

PR #1374, opened 2026-09-14
fix: mixed invocation & unordered set test

### LINKED ISSUE #1369: [BUG] activePlugins is incorrect for mixed-style invocations
_get_required_globals() treats each step key as a task name. For:
```yaml
step:
  task: consume
  kwargs:
    value: $input
```

it collects {"task", "kwargs"} as the used tasks instead of {"consume"}. DynamicGlobalParametersService then fails to identify the plugin containing consume.

This affects the mixed-style YAML generated by the frontend itself.

Suggested fix: Extract the task name using util.step_get_plugin_short_name(step) and collect references separately from the whole invocation.

Location: entrypoints/service.py:2137–2153 (src/dioptra/restapi/v1/entrypoints/service.py#L2137)

jtsextonMITRE:
Closes #1369


henrychoy:
Works for me.  Previously the dynamicGlobalParameters endpoint was returning an empty array for activePlugins, for my example, but now it's populated.  I'm not currently using this property in the frontend, but let me know if I should be displaying it somewhere.

PR #1375, opened 2026-09-16
feat: improve plugin task dialog

henrychoy:
Previously adding params in the plugin task dialog wasn't intuitive because users would forget to hit the add button.  This PR improves the dialog by switching it to a table format where the user can add/delete rows for the params.


jtsextonMITRE (approved):
LGTM

PR #1378, opened 2026-09-18
entrypoint swaps fixes

keithmanville:
issue (blocking): Validate imported entry points before committing the batch - https://github.com/usnistgov/dioptra/commit/8852ea6cac6b17106bada325e89307e616f2b9f2
issue (blocking): Finish removing the retired validation interface - https://github.com/usnistgov/dioptra/commit/b0c4cc6632f6b7c55da28a4592e6e6f9a167e8c6
issue (blocking): Keep step dependencies when rendering a swap - https://github.com/usnistgov/dioptra/commit/6d0b859ac2404d2bbd035d024a5da5c38a474d55
issue (blocking): Restore the existing task-name syntax - https://github.com/usnistgov/dioptra/commit/52f9563db2d9b74f58a2d23156a8ea371e53168d
issue (blocking): Use an unambiguous representation for swap choices - https://github.com/usnistgov/dioptra/commit/796cf1b7da33d1ec17b2e02ab06c4da593e26d7d
issue (blocking): Apply the same validation rule to pending artifact-plugin changes - https://github.com/usnistgov/dioptra/commit/2e74a65e4f82886118423aa6cba7e71fd4d83f96
issue (blocking): Make validate-only requests behave like real dry runs - https://github.com/usnistgov/dioptra/commit/a5974b16b25e3b226ee40b66bb4089a0333a225d
issue (blocking): Validate task-plugin changes before creating an entry-point snapshot - https://github.com/usnistgov/dioptra/commit/55981df85e42129e1a29399d173a6a6b35e121d1
issue (blocking): Prevent any swap selection from forming a dependency cycle - https://github.com/usnistgov/dioptra/commit/7360bf865b056721758687b2daa115dbbc61199e


src/dioptra/restapi/v1/workflows/service.py
```diff
@@ -840,6 +841,7 @@ def _register_entrypoints(  # noqa: C901
                     queue_ids=[],
                     group_id=group_id,
                     commit=False,
+                    on_save=True,
```
jtsextonMITRE, line 844: re: `issue (blocking): Validate imported entry points before committing the batch`

for `OVERWRITE` we delete the entrypoint and then validate the replacement, so if validation fails then overwrite might just delete the old one and give up. maybe this is a separate issue?

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -564,11 +564,11 @@
       <q-btn
         label="Validate"
         color="primary"
-        :disable="taskPluginsChanged"
+        :disable="route.params.id !== 'new' && (taskPluginsChanged || artifactPluginChangesPending)"
```
jtsextonMITRE, line 567: should this use `taskPluginChangesPending`?

src/frontend/src/views/CreateEntryPoint.vue
```diff
@@ -564,11 +564,11 @@
...
         @click="validateEntrypoint()"
       />
-      <q-tooltip v-if="taskPluginsChanged">
-        The task plugin selection has changed. Save your changes before validating.
+      <q-tooltip v-if="route.params.id !== 'new' && (taskPluginsChanged || artifactPluginChangesPending)">
```
jtsextonMITRE, line 570: should this use `taskPluginChangesPending`?

src/dioptra/restapi/v1/entrypoints/service.py
```diff
@@ -955,6 +1038,7 @@ def delete(
         )
 
         with self._uow():
+            self._swaps_validation_service.validate_snapshot(new_entrypoint, log)
             # if the removed plugin is also an artifact plugin do not remove the
             # resource dependency relationship
             if plugin_id not in artifact_plugins:
```
jtsextonMITRE: I think unrelated to this PR, but I think `artifact_plugins` is a list of plugins not ids?

jtsextonMITRE: (fixed in 20f7aac)

PR #1380, opened 2026-09-18
fix: prevent out of order responses on job form

henrychoy:
Closes https://github.com/usnistgov/dioptra/pull/1063#discussion_r4042119067

On the job create page, there are watchers that watch for any update on the entrypoint dropdown or swap choices.  When triggered, they call endpoints that grab the partialGraph or params.  The issue is that there is no guarantee that these calls will finish in chronological order, so if an earlier call takes longer, the form can be updated with old data.

To fix this, each form update will be assigned an id, and a check will be made before each call to make sure it's on the latest id.  If not, the calls are not made.  Vue actually has a built in watcher cleanup, but it cannot coordinate between two watchers, so we're using this id approach.

To test this fix, you need to manually delay the responses from the api.  Insert this line in `CreateJob.vue` on line 838.  This delays getting the partialGraph by 5 seconds.  To repo the issue, select a swap choice, then quickly select a non-swap entrypoint, after 5 seconds, verify that your task graph is not overwritten with the older request for the swap graph.

You can also do the same test by adding this line on line 855 instead.  This delays getting the list of params.  Perform the same test and verify your form isn't overwritten by old data.

```
await new Promise((resolve) => setTimeout(resolve, 5000));
```

You can also verify delaying the entrypoint watcher as well.  Insert the code below on line 781.  Select an entrypoint, notice that the partialGraph is delayed, switch to another entrypoint, verify that the graph isn't overwritten with your first request after 5 seconds.

```
if (latestFormRefreshId === 1) {
  await new Promise((resolve) => setTimeout(resolve, 5000));
}
```


jtsextonMITRE (commented):
It does fix the issue of dynamic global params arriving after switching entrypoints.

Two other things which I'm not sure if they fall under this issue:

1. Selecting a swap before the graph request arrives from `/config` results in the graph not rendering with that swap choice. Reproduce by holding the `/config?sections=graph` result and then select an entrypoint with swaps, choose a swap, then release that result to get an unrendered graph.

2. Clicking `Update Values` on the rerun jobs form while an earlier param request is pending can populate values from the old snapshot. Reproduce by running a job with swaps, changing the swap parameter names, hold the `/dynamicGlobalParameters` response. Then, hold the next `/entrypoints/` response, and click `Update Values`. Release the `dynamicGlobalParameters` response, and now the form will have old parameters until the update values response is released. Though, the backend will prevent actual submission.

